Issue #3, opened 2018-07-30
Future use of k8s plugin to directly create the run image


clemensutschig:
This would save time compared to creating a tarball, uploading it to Nexus, starting the build and downloading the tarball again.

PR #8, opened 2018-08-02
Offer alternative way to build in OpenShift

michaelsauter:
See https://github.com/opendevstack/ods-jenkins-shared-library/issues/3.

Instead of moving the artifacts to Docker via Nexus, we upload the
artifact directly. This should be a bit faster (as downloading is not
necessary), it should be less brittle (as paths in Jenkins and
Dockerfile do not need to match), requires less params (e.g. Nexus
credentials) and less code (as archiving is done automatically by OC).

Also, we now follow the build instead of simply waiting for it, removing
the need to go into OpenShift to see what happened with the build.

Addresses #3.

In a next step, all quickstarters should make use of the new stage.

### DIFF
--- a/vars/stageStartOpenshiftBuild.groovy
+++ b/vars/stageStartOpenshiftBuild.groovy
@@ -1,0 +1,31 @@
+def call(def context) {
+  stage('Build Openshift Image') {
+    if (!context.environment) {
+      println("Skipping for empty environment ...")
+      return
+    }
+
+    timeout(context.openshiftBuildTimeout) {
+      patchBuildConfig(context)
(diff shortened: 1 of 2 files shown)


michaelsauter:
To make use of this in an existent repository created with an older version of the quickstarter, have a look at the following PRs. Before doing so, ensure that your Jenkinsfile does point to `ods-library@0.1-latest`.

* The original PR that adjusts the quickstarters to the changes here: https://github.com/opendevstack/ods-project-quickstarters/pull/9.
* If you are using Python, additionally look at https://github.com/opendevstack/ods-project-quickstarters/pull/24.
* If you are using node, additionally look at https://github.com/opendevstack/ods-project-quickstarters/pull/26.

PR #32, opened 2018-10-31
Set project version

### LINKED ISSUE #29: Set SonarQube project version
We should set the version of the project when scanning.

I suggest the following semantics: if `sonar-project.properties` contains the string `sonar.projectVersion=`, then we leave that as is, assuming the user wants to have their own versioning scheme. Otherwise, we set the version dynamically to the Git revision by passing it to `sonar-scanner` like this: `sonar-scanner -Dsonar.projectVersion=sha`.

michaelsauter:
Closes #29.


michaelsauter:
@clemensutschig @stitakis ping

PR #42, opened 2018-11-19
Making constructBitbucketUrl more robust

rattermeyer:
The Bitbucket URL contained new lines. Don't know exactly how this could happen.
But anyhow we should remove newlines and trim the string

Following is the debug output.

+ echo ''\''http://cd_user
:MjAxODY4NzcxOTIxOrhA+JMfHyfNBf8zkpBTZANVG9Jr
@192.168.56.31:7990/scm/te233/te233-be-spring-boot.git'\'''


rattermeyer:
Before merging: Found the issue. Still think would be a good think to make this more robust.
I encoded the password and also the cd_user access token using: `echo <user> | base64` instead of `echo -n <user> | base64`. So echo printed out a newline. We should document this.

michaelsauter:
@rattermeyer Well, this is the classic base64 encoding mistake :) We could give an example how to encode properly, agreed. I still think it's a good idea to merge this.

PR #43, opened 2018-11-20
Move OCP env cloning to master node

### LINKED ISSUE #36: Move environment cloning to master node
Currently, automatic environment cloning is done via two provided stages, `stageCreateOpenshiftEnvironment` and `stageTriggerAllBuilds`. These are called from each Jenkinsfile on the slave node. However, those stages do not need to be on the slave node, nor do they depend on e.g. build output. Therefore, it would be better to move them into the master node, and just execute them if automatic environment cloning is configured.

Benefits:
* Each Jenkinsfile does not need to call those stages
* Jenkins overview does not show the stages if they are not use

michaelsauter:
Also, use the new clone-project.sh script. That way, the actual cloning
can be tested directly via the script. Also, this script removes the
need to trigger all builds - it simply imports the images from the
cloned project.

Closes #36.


michaelsauter:
Tested it in our project - it took just 39s to clone the environment and trigger all the deployments :)

PR #52, opened 2018-12-13
add sq report pull into shared lib scanSQ task

### LINKED ISSUE #46: Add stage to produce / export CNES / SQ report
we should execute the cnes report automatically after the SQ review - to have a state (= doc) in the build - and store it on a pvc... this way - each build also solidly contains the exported SQ results.

This should be a quick thing to do - add cnes jar to jenkins master .. and on master node execute cnes export after SQ run - that way we have a PVC from master :)

clemensutschig:
fixes #46


clemensutschig:
@michaelsauter - my fix is against v0.1 - can you do the merge then as well?

michaelsauter:
@clemensutschig I will rebase against master and merge.

PR #71, opened 2019-03-29
Check that all required data is present before setting the build status

### LINKED ISSUE #68: NullPointerException if error occurs during prepare stage
If an error occurs during the prepare stage, then the call to set the build status in BitBucket might throw a NPE. We should figure out if we are able to set the build status, and if not, we should not even attempt to set the status. This affects 1.0.x and master.

michaelsauter:
Closes #68. 

Backport to master following ...


clemensutschig:
@michaelsauter  nice - that means we are done with 1.0.2 .. let me know if you want me to cut the release :)

PR #72, opened 2019-04-05
Feature/bix 531 git lfs jenkins configuration

gerardcl:
closes https://github.com/opendevstack/ods-core/issues/76 
it is related to same PR for ods-core repo https://github.com/opendevstack/ods-core/pull/96

### DIFF
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -57,24 +57,7 @@ class OdsPipeline implements Serializable {
         try {
           setBitbucketBuildStatus('INPROGRESS')
           script.wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'XTerm']) {
-            checkout([  $class: 'GitSCM', 
-              branches: [[name: 'refs/heads/'+context.config.gitBranch]],
-              doGenerateSubmoduleConfigurations: false,
-              extensions: [
-                  [$class: 'GitLFSPull'],
-                  [$class: 'CheckoutOption', timeout: 20],
-                  [$class: 'CloneOption',
-                          depth: 0,
-                          noTags: false,
-                          shallow: false,
-                          timeout: 120]
-              ],
-              submoduleCfg: [],
-              userRemoteConfigs: [
-                  [credentialsId: context.config.credentialsId,
-                  url: context.config.gitUrl]
-              ]
-            ])
+            checkoutWithGitLFSpull()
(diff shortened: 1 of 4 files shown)


src/org/ods/OdsPipeline.groovy
```diff
@@ -85,6 +86,21 @@ class OdsPipeline implements Serializable {
...
+                    extensions: [
+                        [$class: 'GitLFSPull']
+                    ],
+                    submoduleCfg: [],
+                    userRemoteConfigs: [
+                        [credentialsId: context.credentialsId,
+                        url: context.gitUrl]
+                    ]
+                ])
+  }
```
michaelsauter: Does this have any performance impact in cause I don't use Git LFS? Also, if I understand correctly, this requires the slave changes - so a less breaking change would be to do a normal checkout if users don't need Git LFS ..

gerardcl: Hi @michaelsauter , I indeed first implemented a method that was checking if the repo was using git lfs and also was dealing with a non git-lfs enabled slave, but since we are using slaves based on the slave base image which has git lfs I considered not requiring it. And, moreover, I detected that the gitLFSPull extension is indeed also dealing with this, so to check if the repo has git lfs enabled or not. 

Therefore, it does a normal checkout or it adds a git lfs pull stage if the repo is git lfs enabled. No performance impact.

There would be a performance impact if we would like to check it twice or to check if the base image is git lfs enabled, which I already tested. And, as said, we should discard checking if the slave is git lfs enabled since it must be based on the jenkins slave.

michaelsauter: I'm not sure I fully understand. My main question would be: does this break on an old slave? If so, doing a check might be worth avoiding the "breaking" change ...

gerardcl: ok! I got you! your proposal is to assure backward compatibility with legacy or non-going-to-be-rebuild jenkins slaves. 
I'm going to come back with this pre-check 👍

gerardcl: hi @michaelsauter , what do you think? are we now ready to promote this change? or shall we also add another check like if the repo is also enabled so to not load the extension all the time? IMHO loading the extension all the time is the right way if the slave node is ready, so less checks. But I would also fine to add such check so to only load the extension when there is really LFS files to be managed...

let me know your thoughts!

Issue #67, opened 2019-03-05
Tagging images is deprecated


michaelsauter:
See:
```
[Pipeline] stage
[Pipeline] { (Deploy to Openshift)
[Pipeline] openshiftTag
NOTE: steps like this one from the OpenShift Pipeline Plugin will not be supported against OpenShift API Servers later than v3.11

Starting "Tag OpenShift Image" with the source [image stream:tag] "foo:45-8af086bd" from the project "bar" and destination stream(s) "foo" with tag(s) "latest" from the project "bar".

Exiting "Tag OpenShift Image" successfully.
[Pipeline] }
[Pipeline] // stage
```

michaelsauter:
We could actually do this VERY differently going forward.

Since we want to define the OCP config via templates, and apply via Tailor during the Jenkins build, maybe we should switch the image via Tailor, too. Otherwise you might run into the situation that you deploy twice - once e.g. because you changed env variables in a deployment config, and then once because you change the image. Doing both thinks in one JSON patch would avoid that.

FYI @torstenhe 

@rattermeyer @tjaeschke @clemensutschig BTW - is the whole OCP config as part of quickstart generation / Jenkins deployment / document generation thing something that is already on your list as a pre-requisite of the release manager?

michaelsauter:
Solved in #144.

PR #85, opened 2019-05-10
WIP: (Optionally) deploy via Tailor

### LINKED ISSUE #96: (Optionally) Deploy via Tailor
In order to manage all OpenShift resources via code, the Jenkins pipeline must be able to read configuration from the repository, and apply any changes necessary to bring the current cluster into the desired state.

Here's how it works in detail:

When no `openshift` folder is present, a deployment is just an update of the
container image.

However, when an `openshift` folder is present, this stage can fully control the
OpenShift resources for this service via code, e.g. it can deploy changes to
environment variable, or changing constraints etc. of the DeploymentConfig. It
does this by using Tailor (https://github.com/opendevstack/tailor) to sync the
"desired state" expressed in templates and parameters with the "current state"
found in OpenShift. Tailor automatically calculates the required changes and
applies them as one atomic JSON patchset.

The "tailor" binary is pre-installed in the slave-base. The behaviour of Tailor
can be customized by adding a "Tailorfile" (or a more specific, per-project
version of it such as "Tailorfile.foo-dev" and "Tailorfile.foo-test").

The actual rollout of the deployment is triggered explicitly via "oc rollout",
which also allows us to monitor the progress and fail the build if the rollout
does not work. Note that any automatic deployment triggers are disabled to
achieve this. Also, the image being rolled out is not tagged with "latest" - the
stage simply deploys the image tag that was built in the previous stage. This
assumes that the template processed by Tailor consumes the "TAGVERSION" param
and that the template uses it correctly to set the image.

This addresses #59 and #67.

michaelsauter:
Closes #96 and #67.


vars/stageDeployToOpenshift.txt
```diff
@@ -0,0 +1,19 @@
+stageDeployToOpenshift deploys the service.
+
+A deployment is more than just updating an image, it might also involve adding
+an environment variable, or changing constraints etc. Because of this, a
+deployment should be one atomic JSON patchset.
```
henrjk: This sounds a bit like that we would want to have someone author a JSON patchset which is then applied.

vars/stageDeployToOpenshift.txt
```diff
@@ -0,0 +1,19 @@
...
+
+When an "openshift" folder is present, we use the "tailor" binary in the
+slave-base to apply its templates. The behaviour is customizable by
+adding a "Tailorfile" (or a more specific, per-project version of it).
+
```
henrjk: On the other hand tailor computes JSON patchset comparing the local definition with the current infrastructure.   So I got a bit confused what was meant with the initial JSON patchset mentioned.

Furthermore I assumed that one creates a JSON patchset so that openshift sees all changes at once (atomic). However the next sentence mentions the triggers which confused me. 

I believe what is envisioned however is that the openshift folder contains the entire openshift resources for a project and one can now make a change by either applying a new TAGVERSION or making some config change in unrelated areas such as a secret or configuration maps which are referenced by a deployment. 
Is this what is envisioned here?

vars/stageDeployToOpenshift.txt
michaelsauter: No, should be automatic - need to reword.

vars/stageDeployToOpenshift.txt
michaelsauter: The `openshift` folder would only contain templates (desired state) of the component, not the whole project. This is a potential downside of the approach taken here.

A "change" is simply a Git commit - and this might be a change in some e.g. Java code that then gets deployed via the image, or it might be a change in the OCP template, which then gets deployed via the JSON patchsets that Tailor generates from it.

The vision is that all changes to the component are done in one repo, and deployed in one Jenkins pipeline - no manual work via different means.

henrjk:
@michaelsauter Regarding your question about [Replace upload-templates.sh with tailor · Issue #38 · opendevstack/ods-project-quickstarters](https://github.com/opendevstack/ods-project-quickstarters/issues/38) work. The current work in progress does not copy the templates into the created projects. But it makes sense to me that tailor templates should end up in the final projects (and perhaps be executed from there?).

Currently the refactored scripts like create/delete-component no longer depend on uploaded templates, and instead apply them with tailor. Ideally the scripts should continue to work if used by other scripts. However I am not sure how much value the reuse of these scripts has. 

I believe newer quickstarters where not actually using these scripts/and templates but rather bringing their own (perhaps adapting copies of the existing scripts/templates).

Perhaps it would make more sense to have some validation code so that labeling rules for components could be ensured to ensure that tailor updates of one component would not damage other components.

Hopefully these comments help.

vars/stageDeployToOpenshift.groovy
```diff
@@ -5,13 +5,34 @@ def call(def context) {
...
+      // Output version to make it easier to diagnose issues.
+      sh "tailor version"
+      // Ensure that no automatic triggers exist as we rollout manually later on.
+      sh "oc --namespace ${context.targetProject} set triggers dc/${context.componentId} --manual"
+      // Detect any <project>.env, Tailorfile.<project> or Tailorfile files.
+      def paramFile = ''
+      if (fileExists("openshift/${context.targetProject}.env")) {
+        paramFile = "--param-file ${context.targetProject}.env"
+      }
+      def tailorFlagsAndArgs = "--non-interactive --namespace ${context.targetProject} --selector app=${context.projectId}-${context.componentId} update --ignore-path bc:/spec/output/imageLabels --ignore-path bc:/spec/strategy/dockerStrategy/buildArgs ${paramFile} --param TAGVERSION=${context.tagversion}"
```
henrjk: This code expects that component to have an app label of the form <projectId>-<componentId>. Typical labels I have seen in quickstarters looked more like this:

```yaml
   labels:
      app: ${PROJECT}
      component: ${COMPONENT}
      env: ${ENV}
  ```

Is this intended here?

michaelsauter: You are correct, the labels created by quickstarters look like you said. I have to adapt it here, or we need to change it when creating components. I used the labels as I have them here because that is the format we use in our initiative.

Thinking a bit more about this, I thought that we might want to put a `Tailorfile` into the `openshift` folder, and specify the label in there. Then it is easy to change for the user.

michaelsauter:
FYI @stitakis

michaelsauter:
From my side, this is now ready to go in. Please provide feedback, ask your questions, etc.

Here's what's left and coming in separate PRs:

* Ensure all newly provisioned components have an `openshift` folder with a proper template. This will be done by @henrjk with the concept we discussed today in the ODS open office hours. See https://github.com/opendevstack/ods-project-quickstarters/issues/38.
* Support encrypted params. Tailor already supports them, but right now this stage does not. For this to work, we need to setup the Jenkins instance with a public/private keypair, and then expose the private key to this stage. Then Tailor will automatically detect and decrypt `*.env.enc` files using the private key, and developers can encrypt those files with the public key which should also be seeded into the repository.
* Prevent accidental damage. Right now this will (without confirmation!) delete resources that have the targeted label but do not exist in the templates. This sounds worse than it is as the resources would have been created initially by those templates, but of course we can do better. Especially around deleting PVCs because of modified disk size or similar. The fix would need to be in Tailor, see https://github.com/opendevstack/tailor/issues/111 for ideas.
* Prevent password leakage. Jenkins already tries hard not to leak password in cleartext. In the Tailor output though, one would see the secrets (even when they are encrypted at rest) in cleartext when they are applied. Tailor should mask them, see https://github.com/opendevstack/tailor/issues/109.

All of those sound like additions to me that would not prevent this from going into master post the 1.1.0 release. Agreed?

stitakis:
@michaelsauter if doable can we add please "support encrypted params" and "prevent password leakage" to 1.2.0

michaelsauter:
@stitakis Yes, I see all of the mentioned follow-ups as part of 1.2.0.

henrjk:
@michaelsauter please take a look at the code in PR #91 at your convenience.

gerardcl:
just saw your last PR @henrjk ( #91 ), so same comments from this PR from my side go there! :)

hugowschneider:
@michaelsauter , maybe is two late, but...

Why should we stop using latest for the DCs? Since latest is just a reference to a "real" tag, which could be names after $TAGVERSION and OC resolves this reference to the sha256 identifier, we could keep latest and benefit from this convention and triggers :). It would keep life simpler. 

It also would bring a problem if my component uses a image for doing it's work. If I have latest I just need to use one reference in its configuration, but if I do not have latest I need to keep changing it...

michaelsauter:
@hugowschneider No it's not too late yet :)

We can also use `latest`, in fact. However, it is necessary to remove the redeploy trigger on changing the latest change. Otherwise, OCP would deploy twice (once because of config change, once because of image change). Since I wanted to avoid that, I thought it is pointless to keep `latest`. Also, I find it better to see immediately which Git Sha the image referes to. Having `latest` makes this harder, because you first need to find the container image sha, and then you see that there is another tag which represents the Git sha ....

For your use case, keep in mind that you can still tag the latest image with `latest` ... we could even do that. But I would still deploy an explicit tag. 

Thoughts?

hugowschneider:
> @hugowschneider No it's not too late yet :)
> 
> We can also use `latest`, in fact. However, it is necessary to remove the redeploy trigger on changing the latest change. Otherwise, OCP would deploy twice (once because of config change, once because of image change). Since I wanted to avoid that, I thought it is pointless to keep `latest`. Also, I find it better to see immediately which Git Sha the image referes to. Having `latest` makes this harder, because you first need to find the container image sha, and then you see that there is another tag which represents the Git sha ....
> 
> For your use case, keep in mind that you can still tag the latest image with `latest` ... we could even do that. But I would still deploy an explicit tag.
> 
> Thoughts?

Disabling the trigger I agree (and reenable it afterwards).

I think the greatest advantage of having **latest** in my point of view, is during migration and image import processes, if I do not use latest, I always need to find which is the tag I am using. And **manually** tagging latest, to reference it, can lead to "human/I forgot to tag" errors...

Looking to it, I would say we could set the DC to the tag name, even if I think the trigger will be helpful in some "manual" case during development, but I would still tag it to latest automatically during the pipeline.

michaelsauter:
@stitakis @henrjk @hugowschneider @gerardcl I've updated it now, and think it's really really cool :)

Most important points:
1. We now rollout and watch it. This is great because it'll fail the build if the container does not boot
2. If there is an `openshift` folder, it uses the templates inside to update OpenShift
3. It deals with any triggers in a smart way, so if they exist it honours them, if not it does the manual work

I would appreciate if you can trial this. The easiest thing to test is without the `openshift` folder - you can simply point your `Jenkinsfile` to this branch, and benefit from the rollout watching.

clemensutschig:
@michaelsauter 1.2?

michaelsauter:
@clemensutschig No, I need to rework this slightly because of the changes you made in regards to the MRO, plus I want to give this a little more testing in not-so-unusual edge cases (no deployment config, multiple deployment configs ...)

clemensutschig:
@michaelsauter  - is this still wip? just asking ...

michaelsauter:
@clemensutschig Yes.

This code is in use by Farmera though, and works well.

Now with some new insights from MRO, I might adjust it slightly. In the original PR, I wanted to be more flexible than what the MRO currently allows and I need to figure out what it means. But I expect to be able to close this off shortly after we finalize the release stuff as it is closely related and will fit nicely together :)

Issue #86, opened 2019-05-13
' in commit message breaks Build Openshift Image step


m-apsolon:
didn't try any other specials chars

m-apsolon:
sorry, didn't see in time, exactly that what you got @ errors;

michaelsauter:
Fix will be `value.replace("'","'\"'\"'")`, thanks @hugowschneider

clemensutschig:
@m-apsolon  - do you contribute the fix? or work with @renedupont  - to make this happen?

m-apsolon:
@clemensutschig no, haven't had time this week, can do next week

renedupont:
@clemensutschig I'm planning on doing it tomorrow if time allows.

michaelsauter:
FYI I played a little with this yesterday, and noticed that besides `'`, also `"` breaks the patching. Need to escape both. And the suggested solution above does not really work ;)

clemensutschig:
@michaelsauter  - I suggest to remove the chars in the first place :) (and also a line feed and do a trim at the end)

PR #90, opened 2019-05-17
Feature/additional labels & commit message sanitizing

### LINKED ISSUE #89: Dump generated BC labels into file (release.json) during jenkins build
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/stageStartOpenshiftBuild.groovy#L20-L28 

generates the BC patch json - with the `ods specific` labels.  at this stage we could also dump this into a file that is baked into the image (release.json) .. 

Great news is that additional entries can be already added :) - but as BC build args only  
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/stageStartOpenshiftBuild.txt#L7 

@michaelsauter  - we should document this a little better :)

https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/stageStartOpenshiftBuild.groovy#L1 - should be extended for imagelabels = [:]

clemensutschig:
fixes #89 an #86

### DIFF
--- a/vars/stageStartOpenshiftBuild.groovy
+++ b/vars/stageStartOpenshiftBuild.groovy
@@ -1,4 +1,4 @@
-def call(def context, def buildArgs = [:]) {
+def call(def context, def buildArgs = [:], def imageLabels = [:]) {
   stage('Build Openshift Image') {
     if (!context.environment) {
       println("Skipping for empty environment ...")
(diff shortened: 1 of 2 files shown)


vars/stageStartOpenshiftBuild.groovy
```diff
@@ -1,34 +1,51 @@
...
+	odsImageLabels.push("{\"name\":\"ods.build.lib.version\",\"value\":\"${context.odsSharedLibVersion}\"}")
+	
+  // add additional ones - prefixed with ext.
+  for (def key : imageLabels.keySet()) {
+    def val = imageLabels[key]
+    odsImageLabels.push("{\"name\": \"ext.${key}\", \"value\": \"${val}\"}")
+  }
+  
+  // write the file - so people can pick it up in the Dockerfile
+  writeFile file: 'docker/release.json', text: odsImageLabels.join(",")
```
michaelsauter: Isn't this missing the wrapping `[]` to make it valid JSON?

PR #91, opened 2019-06-13
Feature/tailor deploy debug

henrjk:
While working on https://github.com/opendevstack/ods-project-quickstarters/issues/38 I ran into some issues with the code in PR #85. 

Here is a version which worked for me.

### DIFF
--- a/vars/stageStartOpenshiftBuild.groovy
+++ b/vars/stageStartOpenshiftBuild.groovy
@@ -5,6 +5,11 @@ def call(def context, def buildArgs = [:]) {
       return
     }
 
+    if (! resourceExists("${context.targetProject}", "dc", "${context.componentId}")) {
+      println("CHK: invoking context.tailorUsingOpenshiftFolder()")
+      context.tailorUsingOpenshiftFolder()
+    }
+
     timeout(context.openshiftBuildTimeout) {
       patchBuildConfig(context, buildArgs)
       sh "oc start-build ${context.componentId} --from-dir docker --follow -n ${context.targetProject}"
@@ -42,6 +47,14 @@ private void patchBuildConfig(def context, def buildArgs) {
   }
 
   sh """oc patch bc ${context.componentId} --type=json --patch '[${patches.join(",")}]' -n ${context.targetProject}"""
+}
+
+private boolean resourceExists(String namespace, String type, String name) {
+  def statusCode = sh(
+    script:"oc --namespace ${namespace} get ${type}/${name} &> /dev/null",
+    returnStatus: true
+  )
+  return statusCode == 0
 }
 
 return this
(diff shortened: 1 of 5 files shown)


vars/stageStartOpenshiftBuild.groovy
```diff
@@ -5,6 +5,11 @@ def call(def context, def buildArgs = [:]) {
       return
     }
 
+    if (! resourceExists("${context.targetProject}", "dc", "${context.componentId}")) {
+      println("CHK: invoking context.tailorUsingOpenshiftFolder()")
```
gerardcl: hi! this should go out or be kept as a debut message, right?

henrjk: Good catch! I would remove it.

michaelsauter:
@henrjk So from your testing, we learned that:

- Initial builds don't work because the deploy step happens afterwards
- Initial deploy does not work because it assumes a `DeploymentConfig` to be present

However, after thinking a bit about this on my bike ride this morning, I think the problem is more general:

- Any changes users want to make to the `BuildConfig` or `ImageStream` should actually happen before the build step
- The `DeploymentConfig` might be changed or even deleted later on, so presence of it should not be required. One could imagine a repository that defines e.g. only some `ConfigMap` resources ... in that case, running Tailor should still work

As a result of that line of thought, maybe we need to split things a little - how about running Tailor just for `BuildConfig` and `ImageStream` resources inside the build step before building the image, and then do the "tailor deploy" step afterwards, but do not assume an existing `DeploymentConfig` in it. What do you think?

henrjk:
One additional thing I learned is:
* Initial builds also don't work because patching of the build config (done to record imageLabels and more) fails (as there is nothing to be patched).

(See https://github.com/opendevstack/ods-jenkins-shared-library/blob/3d84dea2594591cd4f948dc2af8499c0ae612a41/vars/stageStartOpenshiftBuild.groovy#L9)

> * Any changes users want to make to the `BuildConfig` or `ImageStream` should actually happen before the build step
> * The `DeploymentConfig` might be changed or even deleted later on, so presence of it should not be required. One could imagine a repository that defines e.g. only some `ConfigMap` resources ... in that case, running Tailor should still work
> 
Make sense to me. 

> As a result of that line of thought, maybe we need to split things a little - how about running Tailor just for `BuildConfig` and `ImageStream` resources inside the build step before building the image, and then do the "tailor deploy" step afterwards, but do not assume an existing `DeploymentConfig` in it. What do you think?

How would this mesh with the additional learning mentioned on top?

michaelsauter:
@henrjk What I meant by "Initial builds don't work because the deploy step happens afterwards" was actually that the build step fails because patching does not work, and then starting the build does not work because there is no `BuildConfig`. As a result of this, my suggestion: "Any changes users want to make to the BuildConfig or ImageStream should actually happen before the build step" would also address the patching problem ... if you create the `BuildConfig` before patching, it'll work.

michaelsauter:
Closing as #85 has been reworked and covers the feedback from this PR.

PR #101, opened 2019-07-24
Feature/97

ungerts:
Closes #97, #98

PR #108, opened 2019-08-14
Feature/make jenkins ods shared lib ready for MRO

clemensutschig:
replaces #101 (and contains all ungerts/feature/97)

@ungerts  fyi ... if ok - please close your PR

### DIFF
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -150,7 +150,7 @@ class OdsPipeline implements Serializable {
       }
     } else
     {
-      def foundTests = script.sh(script: "ls -la ${testLocation}/.xml | wc -l", returnStdout : true).trim()
+      def foundTests = script.sh(script: "ls -la ${testLocation}/*.xml | wc -l", returnStdout : true).trim()
       script.echo "Found ${foundTests} tests in ${testLocation}"
     }
     
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -95,14 +95,16 @@ class OdsPipeline implements Serializable {
             notifyNotGreen()
           }
           if (!!script.env.MULTI_REPO_BUILD) {
-            logger.debug "MRO build - caught error for ${context.componentId}"
             // this is the case on a parallel node to be interupted
-            if (!(err instanceof org.jenkinsci.plugins.workflow.steps.FlowInterruptedException))
+            if (err instanceof org.jenkinsci.plugins.workflow.steps.FlowInterruptedException)
+            {
+              throw err
+            } else 
             {
               context.addArtifactURI('failedStage', script.env.STAGE_NAME)
               stashTestResults(true)
+              return this
             }
-            return this
           } else {
             throw err
           }
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -112,6 +112,7 @@ class OdsPipeline implements Serializable {
         def debug = script.env.DEBUG
         if (debug != null && context.debug == null) 
         {
+            logger.debug("Setting ${debug} on ${context.projectId}")
             context.debug = debug
         } 
       } else {
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -87,7 +87,7 @@ class OdsPipeline implements Serializable {
           setBitbucketBuildStatus('SUCCESSFUL')
           logger.info "***** Finished ODS Pipeline *****"
         } catch (err) {
-          context.addArtifactURI('failedStage', env.STAGE_NAME)
+          context.addArtifactURI('failedStage', script.env.STAGE_NAME)
           updateBuildStatus('FAILURE')
           setBitbucketBuildStatus('FAILED')
           if (context.notifyNotGreen) {
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -97,15 +97,13 @@ class OdsPipeline implements Serializable {
           if (!!script.env.MULTI_REPO_BUILD) {
             logger.info "MRO build - caught error"
             context.addArtifactURI('failedStage', script.env.STAGE_NAME)
+            stashTestResults(true)
             logger.info "MRO build - returning context: ${context.getBuildArtifactURIs()}"
             return this
           } else {
             throw err
           }
-        } /*finally {
-          // in case called from outside
-          return this
-        } */
+        } 
       }
     }
   }
@@ -142,7 +140,7 @@ class OdsPipeline implements Serializable {
     }
   }
 
-  private void stashTestResults () {
+  private void stashTestResults (def hasFailed = true) {
     def testLocation = "build/test-results/test"
     
     logger.info "stashing testResults : config: ${context.testResults}, defaultlocation: ${testLocation}, same? ${(context.getTestResults() == testLocation)}"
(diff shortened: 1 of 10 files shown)


clemensutschig:
@michaelsauter  / @ungerts  - added new pieces into stageDeploy / Build .. so we have access to the image and also to the deployment status ... (this will increase net time a bit, but help a lot .. also for MRO)

clemensutschig:
@metmajer - fyi ..

vars/stageDeployToOpenshift.groovy
```diff
@@ -12,6 +12,27 @@ def call(def context) {
...
+          break
+      }
+    }
+    
+    if (ocpDeployment[6] != "Complete") {
+      error "Deployment ${context.componentId}${ocpDeployment[2]} failed (status: ${ocpDeployment[6]}), please check the error in the OCP console"
+    }
+    
+    deploymentId = "${context.componentId}-${ocpDeployment[2]}"
+    context.addArtifactURI("OCP Deployment Id", deploymentId)
```
michaelsauter: I think something like this could work:
```
$ oc rollout status -n foo-dev dc/be-main --watch
replication controller "be-main-589" successfully rolled out
```
It returns the ID you are looking for, and the `--watch` part avoids the need to use `sleep`.

You probably also want to use some kind of timeout - otherwise we're waiting for a long time if the container is crashing ...

vars/stageStartOpenshiftBuild.groovy
```diff
@@ -7,8 +7,23 @@ def call(def context, def buildArgs = [:], def imageLabels = [:]) {
...
-      sh "oc start-build ${context.componentId} --from-dir docker --follow -n ${context.targetProject}"
+      sh (script: "oc start-build ${context.componentId} --from-dir docker --follow -n ${context.targetProject}", label : "start openshift build")
     }
+
+    def ocpbuild = sh(returnStdout: true, script:"oc get build --sort-by=.status.startTimestamp --no-headers -n ${context.targetProject} -l buildconfig=${context.componentId} | tail -n1", label : "find last build").trim().split(/\s+/)
+    
+    def ocpbuildId = ocpbuild[0]
+    def ocpbuildStatus = ocpbuild[3]
+    
+    if (ocpbuildStatus.toString().trim().toLowerCase() != "complete") {
```
michaelsauter: I think `oc start-build` returns with exit code 1 if it fails, so not sure if this case can happen?

vars/stageStartOpenshiftBuild.groovy
```diff
@@ -7,8 +7,23 @@ def call(def context, def buildArgs = [:], def imageLabels = [:]) {
...
+      error "OCP Build ${ocpbuildId} was not successfull - status ${ocpbuildStatus}"
+    }
+  
+    def ocpCurrentImage = sh(returnStdout: true, script:"oc get istag ${context.componentId}:${context.getTagversion()} -n ${context.targetProject} --no-headers", label : "find new image").trim().split(/\s+/)
```
michaelsauter: I would like to see this (and the other examples) replaced with JSONpath expressions. Eg.
```
oc get istag -n foo-dev be-main:1-1598e8f1 -o jsonpath='{.image.dockerImageReference}'
```

I believe this is:
1. easier to read
2. more stable as the path is part of the API, whereas the CLI output is (I think) subject to change

clemensutschig: @michaelsauter  - top will do .. (i wish the json engine would be good enough for more complex stuff) - see comment above ..

vars/stageStartOpenshiftBuild.groovy
clemensutschig: I never believe this cli .. better check 🗡

vars/stageDeployToOpenshift.groovy
clemensutschig:
```
utschig@WKS0000627058 MINGW64 /c/code_bix/opendevstack_at_BIX/ods-jenkins-shared-library (feature/97)
$ oc get build --sort-by=.status.startTimestamp -o jsonpath='{.items[.items.length()-1].metadata.name}' -n pltfmdev-dev -l buildconfig=app-be-adeu
error: error parsing jsonpath {.items[.items.length()-1].metadata.name}, invalid array index .items.length()-1

utschig@WKS0000627058 MINGW64 /c/code_bix/opendevstack_at_BIX/ods-jenkins-shared-library (feature/97)
$ oc get build --sort-by=.status.startTimestamp -o jsonpath='{.items[.items.length-1].metadata.name}' -n pltfmdev-dev -l buildconfig=app-be-adeu
error: error parsing jsonpath {.items[.items.length-1].metadata.name}, invalid array index .items.length-1
```

yak .. at least I have to keep this .. but I'll surround it with a timeout

clemensutschig:
@ungerts i have detected a major glitch.. when used standalone an error from build stage does Not bubble anymore (And the build is not set to error).. and funnily cannot be caught ..

clemensutschig:
Ok finally ist complete to work with an as is ods boilerplate pipeline

src/org/ods/OdsPipeline.groovy
```diff
@@ -58,40 +67,125 @@ class OdsPipeline implements Serializable {
...
+    context.addArtifactURI("testResults", foundTests)
+    
+    if (hasFailed && foundTests.toInteger() == 0) 
+    {
+      logger.debug "ODS Build did fail, and no test results,.. returning"
+      return
+    }
+
+    // stash them in the mro pattern
+    script.stash(name: "test-reports-junit-xml-${context.componentId}-${context.buildNumber}", includes: 'build/test-results/test/*.xml', allowEmpty : true)
```
gerardcl: same here, `includes` should pop up all `xml` files if in any subdirectory too:
```
includes: 'build/test-results/test/**/*.xml'
```

clemensutschig:
@gerardcl top Thanks a Lot!

vars/stageDeployToOpenshift.groovy
```diff
@@ -12,6 +12,28 @@ def call(def context) {
       destTag: "latest",
       namespace: context.targetProject
     )
+
+    def ocpLatestImage = sh(returnStdout: true, script:"oc get istag ${context.componentId}:latest -n ${context.targetProject} -o jsonpath='{.image.dockerImageReference}'", label : "find last image").trim()
+      
+    def ocpDeployment
+    timeout(context.openshiftBuildTimeout) {
```
michaelsauter: I think this is "misusing" the build timeout setting. Right now this is meant to stop the container build if it takes too long. Therefore it needs to be configurable because ODS cannot know how long a container image build will take.

Here, the allowed time will be different because the time it takes to boot the service is not equal to the time it takes to build it. 

Therefore I suggest to have a separate timeout parameter here ...

vars/stageDeployToOpenshift.groovy
```diff
@@ -12,6 +12,28 @@ def call(def context) {
...
+      while(true) {
+        ocpDeployment = sh(returnStdout: true, script:"sleep 10 && oc describe dc ${context.componentId} -n ${context.targetProject} | grep -e ${ocpLatestImage} -e Status -e 'Latest Version'", label : "find new deployment").trim().split(/\s+/)
+        
+        echo ("Found last deployment id: ${ocpDeployment[2]} - status ${ocpDeployment[6]}")
+        if (ocpDeployment[6] != "Pending" && ocpDeployment[6] != "Running") {
+            echo "OCP Deployment done - reporting status"
+            break
+        }
+      }
```
michaelsauter: I will try to come up with a different solution for this which does not rely on `sleep 10`. But we can merge this like it is here and then improve (if that is possible at all).

vars/stageDeployToOpenshift.groovy
michaelsauter: I think this should actually work, just different syntax - see https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html.

michaelsauter:
@clemensutschig I've finished my work in #85. That includes a major rework of how deployments are made. You can see the relevant changes in `vars/stageDeployToOpenshift.groovy`. I believe this to improve on the additions you made here. However, it will conflict a bit - I'm fine with having this PR merged, and then I would update my PR to modify your changes.

However, if you think this PR needs a lot more time, then I'd prefer to merge my PR first instead :) Happy to help integrate the changes here then.

Let me know what you think.

PR #110, opened 2019-08-21
arroba to ASCII for clone project script only

### LINKED ISSUE #109: Auto Clone Environment fails during curl download
The statement https://github.com/opendevstack/ods-jenkins-shared-library/blob/443bbda8587070841111b61468b8134f78cbd447/src/org/ods/OdsPipeline.groovy#L165 fails if `@` is used in `userPass`. 
The current substitution https://github.com/opendevstack/ods-jenkins-shared-library/blob/443bbda8587070841111b61468b8134f78cbd447/src/org/ods/OdsPipeline.groovy#L163  
does not successfully escape the character.

gerardcl:
closes #109 
this approval comes with the approval of https://github.com/opendevstack/ods-project-quickstarters/pull/305 too

### DIFF
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -160,13 +160,14 @@ class OdsPipeline implements Serializable {
 
       logger.info 'Environment does not exist yet. Creating now ...'
       script.withCredentials([script.usernameColonPassword(credentialsId: context.credentialsId, variable: 'USERPASS')]) {
-        def userPass = script.USERPASS.replace('@', '%40').replace('$', '\'$\'')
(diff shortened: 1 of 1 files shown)


clemensutschig:
@gerardcl regarding the rights.. the sa has no rights for cloning.. Steve kehoe..

gerardcl:
ok, will check swith Stephen...since it is also blocking other teams...

Michael, as said in the bro-PR of this one, I also agree, anyway shall we tackle better in this related issue the special characters topic or we create a issue that handles it better?

michaelsauter:
@gerardcl First fix, then make it pretty in a separate issue :)

Issue #122, opened 2019-09-06
Extend documentation with instruction about how to add snyk scanning to ods project


stitakis:

stitakis:
#123

rattermeyer:
yes, and we generally should move documentation to antora. maintaining both sets is cumbersome.
PR follows in a few minutes

PR #124, opened 2019-09-10
update antora documentation

rattermeyer:
removed content and added link to the antora documentationand add reference from .txt files

solves #122


michaelsauter:
I like that things become more unified, but I am still lost on the process - what should I do when it comes to documentation, and how do I do that? Would appreciate if you shed more light on this.

PR #128, opened 2019-09-25
add missing dependency

### LINKED ISSUE #127: maven build is broken - missing dependency
Maven build via `mvn install` fails:

> [ERROR] Failed to execute goal org.codehaus.gmavenplus:gmavenplus-plugin:1.5:compile (default) on project jenkins-shared-library: Error occurred while calling a method on a Groovy class from classpath.: InvocationTargetException: startup failed:
[ERROR] D:\projekte\ods\src\ods-jenkins-shared-library\src\org\ods\OdsPipeline.groovy: 99: unable to resolve class org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
[ERROR]  @ line 99, column 32.
[ERROR]                if (err instanceof org.jenkinsci.plugins.workflow.steps.FlowInterruptedException)
[ERROR]                                   ^
[ERROR]
[ERROR] 1 error
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.

stefanlack:
fixes #127

I have added one dependency.

However, now the following tests are failing:

```
Failed tests:   testDetermineEnvironment_githubflow(org.ods.OdsContextTest): expected:<...eview-123> but was:<...od>
  testDetermineEnvironment_autoclone(org.ods.OdsContextTest): expected:<hotfix-123> but was:<rel-1.0.0>
  testDetermineEnvironment_gitflow(org.ods.OdsContextTest): expected:<dev> but was:<prod>
```

Additional work has to be done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,5 @@ install: true
 jdk:
   - openjdk8
 
-script:
-  - ./mvn clean test -B
+
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: java
+install: true
+
+
+jdk:
+  - openjdk8
+
+script:
+  - ./mvn clean test -B
+
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,13 @@
   </repositories>
 
   <dependencies>
+    <!-- https://mvnrepository.com/artifact/org.jenkins-ci.plugins.workflow/workflow-step-api -->
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-step-api</artifactId>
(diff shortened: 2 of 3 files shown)


stefanlack:
@michaelsauter :

 Failing OdsContextTest is correted now. While the assertions all have been valid, setup was not correct. When calling `OdsContext#determineEnvironment`, the environment is only determined, if it not was determined before:

```
  void determineEnvironment() {
    if (config.environment) {
      // environment already set
      return
    }
```

Solution in Testclass: set config.environment to null before each assertion. Move this code in newly created method org.ods.OdsContextTest#determineEnvironment in order to remove duplication.

PR #139, opened 2019-10-11
Extract environment check to utils

### LINKED ISSUE #138: environmentExists check exists twice
We should have only one piece of code doing that check.

michaelsauter:
Closes #138.

### DIFF
--- a/src/org/ods/OpenshiftUtils.groovy
+++ b/src/org/ods/OpenshiftUtils.groovy
@@ -5,7 +5,7 @@ class OpenshiftUtils {
     static boolean environmentExists(def script, String name) {
         def statusCode = script.sh(
             script:"oc project ${name} &> /dev/null",
-            label: "check if OCP environment exists",
+            label: "check if OCP environment ${name} exists",
             returnStatus: true
         )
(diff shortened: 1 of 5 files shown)


michaelsauter:
@henrjk Noticed that there were tests, and that I broke them. Now I fixed the tests and this would be ready from my side.

src/org/ods/OpenshiftUtils.groovy
```diff
@@ -0,0 +1,13 @@
+package org.ods
+
+class OpenshiftUtils {
+
+    static boolean environmentExists(def script, String name) {
+        def statusCode = script.sh(
+            script:"oc project ${name} &> /dev/null",
+            label: "check if OCP environment exists",
```
clemensutschig: michael - can you please add the `env` back into the label?

michaelsauter: Good point, I didn't even realize that the two versions of `enviornmentExists` differed and by chance I took the "wrong one" without the name in the label. Fixed in ef1c2f2b6ddd04fef05123392e4f41067e5f4d2b.

Issue #142, opened 2019-10-14
Seldom error in stageDeployToOpenshift (ArrayIndexOutOfBounds)- when checking for new deployment


clemensutschig:
**Describe the bug**
.....
java.lang.ArrayIndexOutOfBoundsException: 6
	at org.codehaus.groovy.runtime.dgmimpl.arrays.ObjectArrayGetAtMetaMethod.invoke(ObjectArrayGetAtMetaMethod.java:41)
....
	at stageDeployToOpenshift.call(/var/lib/jenkins/jobs/trainods-cd/jobs/trainods-cd-be-golang-master/builds/2/libs/ods-library/vars/stageDeployToOpenshift.groovy:23)
	at ___cps.transform___(Native Method)

**To Reproduce**
hard, only if deployment scheduling takes forever 

**Expected behavior**
no error, deployment search continues

**Affected version (please complete the following information):**
 - OpenDevStack [1.2.x/master]

michaelsauter:
I will address this when I redo that logic using other `oc` commands. I will also add timeouts to all calls.

michaelsauter:
This should be fixed with #147.

PR #147, opened 2019-10-25
Improve deployment rollout

### LINKED ISSUE #144: Improve deployment stage
There are a couple of improvements I'm planning to implement for the deployment stage:

* Right now, the deployment assumes that `ImageStream` and `DeploymentConfig` of a certain name exist, and it does not fail in a friendly way if this is not the case.
* The deployment happens because the `latest` tag is updated. The command used for that is deprecated (see #67), and more generally, it assumes that an image trigger is present which might not be the case.
* There is no deployment timeout - right now the step can take very long and the user cannot control how long it is allowed to take.
* Since 1.2, the result of the deployment is checked (which is awesome as it fails the build when the deployment fails - combined with health checks that allows for a very robust rollout), however the code relies on a lot of `oc` output, which is brittle (see e.g. #142) and does quite a few calls, which can be optimised.
* Finally, it would be nice to expose the "building blocks" of deployment so that quickstarters which need more control over the deployment (e.g. because they have two `DeploymentConfig` resources) can use those building blocks.

I have a prototype of this ready and will give it some test time this week and then open a PR.

michaelsauter:
Closes #144.


michaelsauter:
Fixes #142.

PR #151, opened 2019-11-13
fix: add resource cpu to container template

SimonGolms:
specify cpu and memory for resource requests, otherwise the slave will try to reserve all available cpu
- set request cpu to 10 millicores
- set limit cpu to 300 millicores

### DIFF
--- a/src/org/ods/OdsContext.groovy
+++ b/src/org/ods/OdsContext.groovy
@@ -122,6 +122,8 @@ class OdsContext implements Context {
           name: 'jnlp',
           image: config.image,
           workingDir: '/tmp',
+          resourceRequestCpu: '10m',
+          resourceLimitCpu: '300m',
           resourceRequestMemory: '1Gi',
           resourceLimitMemory: '2Gi',
           alwaysPullImage: config.podAlwaysPullImage,


michaelsauter:
@SimonGolms How did you come up with the specific values?

SimonGolms:
10m; to stay alive
300m: best practice / randomly

If you need a custom jenkins slave, you will probably need other resource requirements as well. the information is therefore only used as fallback to avoid unnecessary stress on the cluster.

How did you calculate the memory specifications? Feel free to adjust the values according to your own preferences.

michaelsauter:
@SimonGolms Thanks for the explanation. The memory values are based on the "worst" quickstarters (e.g. Angular needs for its production build 2Gi). This should be adjusted in the future to have those values appropriate for the quickstarters and easier to modify for users.

michaelsauter:
@rattermeyer Can you have a look at this as well?

michaelsauter:
Moving this to v3.

On some clusters, 50m seems to be the minimum. Further, some slaves might need more than 300m to run at a reasonable speed. 

We probably need sth. that is configurable, and specific for each type of quickstarter.

michaelsauter:
Closing this in favour of #175.

Issue #158, opened 2019-11-22
Special characters in last commit message break build


tjaeschke:
**Describe the bug**
The shared lib version comes with a newline special char, which breaks the build. This only happens, if the shard lib version is the last env variable.

The command `env | grep 'library.ods-library.version' | cut -d= -f2` doesn't trimm linefeeds and has to be extended to `env | grep 'library.ods-library.version' | cut -d= -f2 | tr -d '\n\r'`

**To Reproduce**
Ensure that the lib version is the last env variable and start a build

**Expected behavior**
The shared lib version doen't have any special characters

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack all

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
[prov-cd-ods-provisioning-app-production] Running shell script
+ oc patch bc prov-app --type=json --patch '[{"op": "replace", "path": "/spec/source", "value": {"type":"Binary"}},{"op": "replace", "path": "/spec/output/to/name", "value": "prov-app:5-3431c12f"},{"op": "replace", "path": "/spec/output/imageLabels", "value": [
        {"name":"ods.build.source.repo.url","value":"http://cd_user@192.168.56.31:7990/scm/opendevstack/ods-provisioning-app.git"},{"name":"ods.build.source.repo.commit.sha","value":"3431c12f503c91bb6bf1ddd9cd76d08bbba4b25d"},{"name":"ods.build.source.repo.commit.msg","value":"Merge branch master of https://github.com/opendevstack/ods-provisioning-app into #283  Conflicts: 	src/main/resources/application.properties"},{"name":"ods.build.source.repo.commit.author","value":"Torsten Jaeschke (torsten.jaeschke@opitz-consulting.com)"},{"name":"ods.build.source.repo.commit.timestamp","value":"2019-11-22 13:00:13 +0100"},{"name":"ods.build.source.repo.branch","value":"production"},{"name":"ods.build.jenkins.job.url","value":"https://jenkins-prov-cd.192.168.56.101.nip.io/job/prov-cd/job/prov-cd-ods-provisioning-app-production/5/"},{"name":"ods.build.timestamp","value":"Fri Nov 22 12:48:32 UTC 2019"},{"name":"ods.build.lib.version","value":"production
"}
      ]}]' -n prov-test
[...]

PR #160, opened 2019-11-22
Remove Linefeed in library version

tjaeschke:
relates to #158

### DIFF
--- a/src/org/ods/OdsContext.groovy
+++ b/src/org/ods/OdsContext.groovy
@@ -45,7 +45,7 @@ class OdsContext implements Context {
     config.nexusPassword = script.env.NEXUS_PASSWORD
     config.openshiftHost = script.env.OPENSHIFT_API_URL
     config.bitbucketHost = script.env.BITBUCKET_HOST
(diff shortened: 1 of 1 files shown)


src/org/ods/OdsContext.groovy
```diff
@@ -45,7 +45,7 @@ class OdsContext implements Context {
     config.nexusPassword = script.env.NEXUS_PASSWORD
     config.openshiftHost = script.env.OPENSHIFT_API_URL
     config.bitbucketHost = script.env.BITBUCKET_HOST
-    config.odsSharedLibVersion = script.sh(script: "env | grep 'library.ods-library.version' | cut -d= -f2", returnStdout: true, label : 'getting ODS shared lib version')
+    config.odsSharedLibVersion = script.sh(script: "env | grep 'library.ods-library.version' | cut -d= -f2 | tr -d '\\n\\r'", returnStdout: true, label : 'getting ODS shared lib version')
```
michaelsauter: wouldn't it be easier to use `.trim()` at the end?

tjaeschke: Yes, this would be easier, even with my brain whispering, "but there are piped shell commands..." :-D

Issue #159, opened 2019-11-22
step find last build in stageStartOpenshiftBuild fails


stefanlack:
**Describe the bug**
_stageStartOpenfhistBuild_ fails in step _find last build', in case it may be the first build.

The following code sets the variable _buildId_  to the value _build_ :

https://github.com/opendevstack/ods-jenkins-shared-library/blob/4dc12830e6aeec8689c41f24e2400644dadc3dd5/vars/stageStartOpenshiftBuild.groovy#L18-L29

**Log Output**
```
[Pipeline] sh (find last build)
[prov-cd-ods-provisioning-app-production] Running shell script
+ oc -n prov-test get build build -o 'jsonpath={.status.phase}'
Error from server (NotFound): builds.build.openshift.io "build" not found
```

**Affected version:**
 - OpenShift: 3.11
 - Master branches

**Additional context**

stefanlack:
Last commit on file found https://github.com/opendevstack/ods-jenkins-shared-library/commit/139c79c242f69bf86501734b56a84b126fc823a5 

@michaelsauter  for your information.

michaelsauter:
Ouch. This works in our cluster. Could you copy the stage into your Jenkinsfile and output `startBuildInfo` - would be interesting to know what your result is.

stefanlack:
The `startBuildInfo` is already printed to the log:

**Start of relevant log:**
```
[Pipeline] sh (Test existance of path .spec.output.imageLabels)
[prov-cd-ods-provisioning-app-production] Running shell script
+ oc -n prov-test get bc/prov-app -o 'jsonpath={.spec.output.imageLabels}'
[Pipeline] sh
[prov-cd-ods-provisioning-app-production] Running shell script
+ oc patch bc prov-app --type=json --patch '[{"op": "replace", "path": "/spec/source", "value": {"type":"Binary"}},{"op": "replace", "path": "/spec/output/to/name", "value": "prov-app:19-906716ce"},{"op": "replace", "path": "/spec/output/imageLabels", "value": [
[...]

PR #161, opened 2019-11-22
Added pattern for different oc outputs

tjaeschke:
relates to #159

### DIFF
--- a/vars/stageStartOpenshiftBuild.groovy
+++ b/vars/stageStartOpenshiftBuild.groovy
@@ -15,7 +15,12 @@ def call(def context, def buildArgs = [:], def imageLabels = [:]) {
         returnStdout: true
       ).trim()
       echo startBuildInfo
-      buildId = startBuildInfo.split(' ').first().split('/').last()
+      def match = (startBuildInfo =~ /\"[a-z_A-Z0-9\s-\/]*\"/)
+      if(match.find()) {
+        buildId = match.group(0).replace('"','')
+      } else {
+        buildId = startBuildInfo.split(' ').first().split('/').last()
+      }	    
     }
 
     if (buildId == null) {
--- a/vars/stageStartOpenshiftBuild.groovy
+++ b/vars/stageStartOpenshiftBuild.groovy
@@ -15,12 +15,13 @@ def call(def context, def buildArgs = [:], def imageLabels = [:]) {
         returnStdout: true
       ).trim()
       echo startBuildInfo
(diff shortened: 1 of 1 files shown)


stefanlack:
LGTM

michaelsauter:
Does not work for us. Please do not merge. I'll update.

vars/stageStartOpenshiftBuild.groovy
```diff
@@ -15,7 +15,12 @@ def call(def context, def buildArgs = [:], def imageLabels = [:]) {
         returnStdout: true
       ).trim()
       echo startBuildInfo
-      buildId = startBuildInfo.split(' ').first().split('/').last()
+      def match = (startBuildInfo =~ /\"[a-z_A-Z0-9\s-\/]*\"/)
+      if(match.find()) {
+        buildId = match.group(0).replace('"','')
+      } else {
+        buildId = startBuildInfo.split(' ').first().split('/').last()
+      }
```
michaelsauter:
```suggestion
def startBuildInfoHeader = startBuildInfo.readLines().first()
def match = (startBuildInfoHeader =~ /${context.componentId}-([0-9]+)/)
if (match.find()) {
    buildId = res[0][1]
} else {
  error "Did not find build ID in build start result '${startBuildInfo}'"
}
```

stefanlack:
@michaelsauter can you please give an example of the content of _startBuildInfo_ in your enviroment, so we can test both versions?

This is the contend on our site:
`String startBuildInfo = 'build "prov-app-1" started'`

michaelsauter:
One sec

michaelsauter:
@stefanlack It is `build.build.openshift.io/be-main-667 started`. I noticed though that `readLines` results in `Scripts not permitted to use staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods readLines java.lang.String`, I'm testing an alternative and will update.

michaelsauter:
@stefanlack @tjaeschke Please review my latest suggestion.

vars/stageStartOpenshiftBuild.groovy
tjaeschke: We will test it. Looks good to me.

tjaeschke:
@michaelsauter The regex was a quickshot and I agree, we should revisit it after 2.0.

michaelsauter:
@tjaeschke Can you please apply my suggestion then and merge?

tjaeschke:
closes #159

PR #163, opened 2019-11-26
removed unnecessary cd_user

tjaeschke:


src/org/ods/OdsContext.groovy
```diff
@@ -385,12 +385,7 @@ class OdsContext implements Context {
       returnStdout: true, script: 'git config --get remote.origin.url',
       label : 'getting GIT url'
     ).trim()
-    // check if the url already contains a user
-    if (!gitUrl.contains('@')) {
-      return gitUrl.replace('://', '://cd_user@')
-    } else {
-      return gitUrl
-    }
+    return gitUrl
```
michaelsauter: Are you 100% sure? I am worried that we did this replacement for a reason 😆

tjaeschke: I would suggest, we keep all project specifix cd user PRs open until end of week or thursday evening.

michaelsauter: OK. I tried to trace usages of this variable and I would agree that theoretically it should be fine. The image label will change (in `stageStartOpenshiftBuild`), but that should be OK.

tjaeschke:
relates to https://github.com/opendevstack/ods-project-quickstarters/issues/289

PR #164, opened 2019-11-27
Adjust location of clone script

michaelsauter:
@henrjk Not sure, but you experimented quite a lot with that feature if I remember correctly. Do you still have an existing clone, and could potentially try this adjustment to see if nothing changed?


henrjk:
At the moment `clone-project.sh` and the associated export/import scripts are not in ods-quickstarter as gat as I can tell.

To me it feels more like an ods-core concern. Do you see reasons not to move it to ods-core?

michaelsauter:
@henrjk Oops. I meant to reference `ods-core` actually, as the scripts have been moved there with https://github.com/opendevstack/ods-core/pull/256. I've update the PR here now to point to `ods-core`.

henrjk:
I just tried out cloning on a newly provisioned bePythonFlask component based on ods-quickstarters and the latest provisioning app. 

I did not notice any issues and the cloned environment appears to work as desired.

PR #176, opened 2019-12-20
WIP: Analyse pull requests with SonarQube

### LINKED ISSUE #174: Integrate SonarQube with pull requests
Pull requests should include the assessment of SonarQube so that issues can be addressed before the merge.

We need to figure out how ODS can support this without forcing it for users that use the free version of SonarQube which does not provide that feature.

michaelsauter:
Very rough draft of pull request integration. Needs further tweaking, error handling, etc, and companion PRs in `ods-core` to add the required SCM plugin to SonarQube and setup of the Bitbucket token.

But general approach works, and PRs in Bitbucket get decorated with comments from SonarQube. 

Closes #174.

### DIFF
--- a/vars/stageScanForSonarqube.groovy
+++ b/vars/stageScanForSonarqube.groovy
@@ -1,5 +1,14 @@
-def call(def context, def requireQualityGatePass = false) {
-  if (context.sonarQubeBranch == '*' || context.sonarQubeBranch == context.gitBranch) {
+def call(def context, def requireQualityGatePass = false, def longLivedBranches = []) {
+  def prAnalysisEnabled = context.analysePullRequestsWithSonarQube
+  // If the Jenkinsfile author did not specify explicitly, we'll default to
+  // configuration in /etc/sonarqube/config.json.
+  if (prAnalysisEnabled == null) {
+    prAnalysisEnabled = isPullRequestAnalysisAvailable(context.odsConfig)
+  }
+
+  def scanAllBranches = prAnalysisEnabled || context.sonarQubeBranch == '*'
+
+  if (scanAllBranches || context.sonarQubeBranch == context.gitBranch) {
     stage('SonarQube Analysis') {
       withSonarQubeEnv('SonarServerConfig') {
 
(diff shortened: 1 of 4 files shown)


michaelsauter:
@clemensutschig we are using the personal access token of the cd user. So far, the token is required because the SonarQube integration with Bitbucket seems to be based on tokens, not on user/pwd. That choice makes sense for me though - you don't really want to pass the pwd to SonarQube ...

In general, we should consider moving to tokens everywhere in our code as well, as this is the recommended (and in my eyes better) way to interact with the Bitbucket API.

michaelsauter:
Open Todo:

- [ ] As far as I can tell, the API does not allow to detect which edition is used. For PR analysis, the commercial "Developer Edition" is required. How can users tell ODS that they want to use this feature? At the moment, I see a few options: 1. add a new setting to the `Jenkinsfile` which is off by default, and users need to turn it on if they want PR analysis; 2. write a file into the base slave which says if the feature should be on/off. The value is taken from the global config, 3. query a config map value in the `cd` namespace. I don't like (3), I don't really like (1) either. Maybe (2) and have a way to opt-out in the Jenkinsfile?

- [ ] The default branch for which SonarQube analysis is performed is `master` at the moment. This should probably change to `*` being the default, otherwise PR branches will never be scanned.

- [ ] For the integration to work, a personal access token is needed. We should create this for new projects for the (project) cd user, and we would need to find a way to upgrade existing projects. Also, we still need to figure out how we should share the token with the pipeline - probably a Jenkins secret would be better than an environment variable?

@clemensutschig @hugowschneider @renedupont @henrjk Any input on the above open questions?

michaelsauter:
As there hasn't been any suggestion / comment, I will do the following:

1. Add a file into the base image which describes the SonarQube capabilities. The Jenkins stage will read this file and determine if PR analysis should be performed or not based on that.
2. Expose a configuration option which allows users to explicitly turn the feature on/off for whatever reason they might have to override the slave file (or if the file cannot be found).
3. We can actually control on which branches the stage runs based on the SonarQube capabilities. If PR analysis is turned off, we run SonarQube scans only on the configured branch, otherwise the setting does not apply.

I have not made up my mind yet regarding the personal access token (last todo). Feedback still welcome :)

vars/stageScanForSonarqube.groovy
```diff
@@ -1,5 +1,19 @@
-def call(def context, def requireQualityGatePass = false) {
-  if (context.sonarQubeBranch == '*' || context.sonarQubeBranch == context.gitBranch) {
+def call(def context, def requireQualityGatePass = false, def longLivedBranches = []) {
+  def prAnalysisEnabled = context.analysePullRequestsWithSonarQube
+  // If the Jenkinsfile author did not specify explicitly, we'll default to
+  // configuration in /etc/sonarqube/config.json.
+  if (prAnalysisEnabled == null) {
+    try {
```
henrjk: Should there be an info in the log somewhere to tip users off about the existence of `analysePullRequestsWithSonarQube`?

michaelsauter: I actually hope that no-one will need/want to use it. It is only there to be able to control it explicitly even if the config is not present.

michaelsauter:
@henrjk Thanks for  reviewing!

To your question if we can determine the edition: I browsed the API to see if we can find out which edition is used. There was none - and I doubt this would be a good thing to have anyway.

Therefore we now have the manual configuration in the base slave, see https://github.com/opendevstack/ods-core/pull/374.

Do you have hints regarding the question how to deal with the token?

henrjk:
Is the Bitbucket token we are talking about the personal token as described at https://confluence.atlassian.com/bitbucketserver066/personal-access-tokens-978197652.html ?

michaelsauter:
@henrjk yes it is.

By now I am pretty set on using Jenkins credentials to sync the bitbucket token. Therefore I'll revise the pull request.

Also, I think we should probably stick the configuration file in the Jenkins master, not the slave image. Then we can pick it up in the ODS context. That allows things to work even when people do not use the base image, and we can use the configuration in other places too.

michaelsauter:
Revised based on https://github.com/opendevstack/ods-core/pull/410. Now the only thing that is left is to deal with the BB token properly.

michaelsauter:
I need to review how this compares to https://marketplace.atlassian.com/apps/1220660/sonarqube-for-bitbucket-server?hosting=server&tab=overview.

michaelsauter:
Superseded by #243.

Issue #74, opened 2019-04-11
Hard coded protocol in OdsPipeline.groovy breaks Jenkins Build


tjaeschke:
In line 103 of `/src/org/ods/OdsPipeline.groovy`
```
https://${context.bitbucketHost}/rest/build-status/1.0/commits/${context.gitCommit}
```
the harcoded protocol breaks the Jenkins Build, if Bitbucket is not running on https, for example in a local development environment.

michaelsauter:
Hmm. Should we add an optional context param then, named e.g. `bitbucketScheme` to deal with this or do you have a better idea? I don't really like what I am suggesting ...

clemensutschig:
What prevents us from adding the Scheme to the bitbucketHost variable in
the first Place ? That would fix this issue, no?

Michael Sauter <notifications@github.com> schrieb am Di., 7. Mai 2019,
23:40:

michaelsauter:
@clemensutschig We could, but:

* All `*Host` variables should consistently include the scheme or not. Personally, I would say that a URL is made up of 4 parts: scheme, host, path and query string. So by the name of it I would guess it does NOT include the scheme...
* Ideally the change is backwards compatible, so if we add the scheme we need to do some checks in the code whether the scheme is set and default to `https://`

gerardcl:
I vote +1 on last comment since we also had issues with Nexus Host management/usage and had to implement a fast solution to get ony the host.
So, we could indeed do that for any url...

michaelsauter:
The proposed long-term solution is to pass URLs instead of hosts. Assigning this to ODS 3.

michaelsauter:
See also https://github.com/opendevstack/ods-core/issues/259.

michaelsauter:
Closing this as that specific bug is already fixed by #182.

PR #182, opened 2020-01-14
Expose bitbucketUrl, taken from BITBUCKET_URL if present

michaelsauter:
Following this, we should update ods-core to set BITBUCKET_URL instead of
BITBUCKET_HOST. Older projects still using BITBUCKET_HOST will continue
to work as HTTPS is already assumed at the moment.

Therefore, I consider this a non-breaking change. It should not be
needed to run any migration script to convert old projects as HTTPS
should be used in real environments anyway, the HTTP variant only exists
for (temporary) testing/local environments.

Addresses one part of #74.

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -276,8 +276,11 @@ When you write stages, you have access to both global variables (defined without
 | odsSharedLibVersion
 | ODS Jenkins shared library version, taken from reference in Jenkinsfile.
 
+| bitbucketUrl
+| BitBucket URL - value taken from BITBUCKET_URL. If BITBUCKET_URL is not present, it will default to https://<BITBUCKET_HOST>.
+
 | bitbucketHost
-| BitBucket host - value taken from BITBUCKET_HOST.
+| BitBucket host - value taken from bitbucketUrl.
 
 | environmentCreated
 | Whether an environment has been created during the build.
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -280,7 +280,7 @@ When you write stages, you have access to both global variables (defined without
 | BitBucket URL - value taken from BITBUCKET_URL. If BITBUCKET_URL is not present, it will default to https://<BITBUCKET_HOST>.
 
 | bitbucketHost
-| BitBucket host - value taken from bitbucketUrl.
+| BitBucket host - value derived from bitbucketUrl.
 
 | environmentCreated
 | Whether an environment has been created during the build.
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/index.adoc
```diff
@@ -276,8 +276,11 @@ When you write stages, you have access to both global variables (defined without
 | odsSharedLibVersion
 | ODS Jenkins shared library version, taken from reference in Jenkinsfile.
 
+| bitbucketUrl
+| BitBucket URL - value taken from BITBUCKET_URL. If BITBUCKET_URL is not present, it will default to https://<BITBUCKET_HOST>.
+
 | bitbucketHost
-| BitBucket host - value taken from BITBUCKET_HOST.
+| BitBucket host - value taken from bitbucketUrl.
```
hugowschneider: Again `BitBucket host - value taken from bitbucketUrl.` I guess there is a mix of Host and URL in the sentence

src/org/ods/Context.groovy
```diff
@@ -148,9 +148,12 @@ interface Context {
     // ODS Jenkins shared library version, taken from reference in Jenkinsfile.
     String getOdsSharedLibVersion()
 
-    // BitBucket host - value taken from BITBUCKET_HOST.
+    // BitBucket host - value taken from BITBUCKET_URL.
```
hugowschneider: I guess the old comment `// BitBucket host - value taken from BITBUCKET_HOST` was c

src/org/ods/OdsPipeline.groovy
```diff
@@ -205,7 +205,7 @@ class OdsPipeline implements Serializable {
             --request POST \\
             --header \"Content-Type: application/json\" \\
             --data '{\"state\":\"${state}\",\"key\":\"${buildName}\",\"name\":\"${buildName}\",\"url\":\"${context.buildUrl}\"}' \\
-            https://${context.bitbucketHost}/rest/build-status/1.0/commits/${context.gitCommit}
+            ${context.bitbucketUrl}/rest/build-status/1.0/commits/${context.gitCommit}
```
hugowschneider: Can we have a variable to set a --insecure flag? That could help if for any reason the repo is behind a self-signed certificate

docs/modules/ROOT/pages/index.adoc
michaelsauter: No, this value is really derived from `bitbucketUrl`. I can make the wording clearer.

src/org/ods/Context.groovy
michaelsauter: Actually it is correct, but I can make the wording clearer.

src/org/ods/OdsPipeline.groovy
michaelsauter: Different issue, different PR :)

src/org/ods/Context.groovy
michaelsauter: 1df927d

PR #183, opened 2020-01-14
Add nexusUrl in favour of nexusHost

michaelsauter:
nexusHost contained the value of NEXUS_HOST previously, which, despite
its name, contained not just the host but also the scheme part.

For consistency, we now prefer NEXUS_URL, and expose this as nexusUrl.
nexusHost will now NOT include the scheme anymore. This is a breaking
change. Users of ODS will be affected and need to update their
Jenkinsfiles accordingly. There is no way to make this non-breaking, we
could only delay the impact by setting NEXUS_HOST/nexusHost to the same
value as nexusUrl, but that would be weird end end up being a breaking
change when we adapt it to be host only in the future.

Addresses part of #74.

There will need to be follow-up PRs in ods-quickstarters and ods-core.

I do not see a way to make this non-breaking (different to #182, which does not have that problem). If you have ideas, please mention them. FYI @tjaeschke @clemensutschig @metmajer @hugowschneider

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -201,8 +201,11 @@ When you write stages, you have access to both global variables (defined without
 | notifyNotGreen
 | Whether to send notifications if the build is not successful.
 
+| nexusUrl
+| Nexus URL (with scheme).
+
 | nexusHost
-| Nexus host (with scheme).
+| Nexus host (without scheme).
 
 | nexusUsername
 | Nexus username.
@@ -210,8 +213,8 @@ When you write stages, you have access to both global variables (defined without
 | nexusPassword
 | Nexus password.
 
-| nexusHostWithBasicAuth
-| Nexus host (with scheme), including username and password as BasicAuth.
+| nexusUrlWithBasicAuth
+| Nexus URL (with scheme), including username and password as BasicAuth.
 
 | branchToEnvironmentMapping
 | Define which branches are deployed to which environments.
--- a/src/org/ods/Context.groovy
+++ b/src/org/ods/Context.groovy
@@ -64,7 +64,10 @@ interface Context {
     // Enable/disable notifications
     void setNotifyNotGreen(boolean notifyNotGreen)
 
-    // Nexus host (with scheme).
+    // Nexus URL (with scheme).
+    String getNexusUrl()
+
+    // Nexus host (without scheme).
     String getNexusHost()
 
     // Nexus username.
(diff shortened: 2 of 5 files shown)


michaelsauter:
The breaking changes are:
 
1. `context.nexusHost` will still be accessible, but it does not contain the scheme anymore. This may cause problems everywhere it is used in a `Jenkinsfile`. Example: by default, `be-java-springboot` has a `stageBuild` defined which exposes an env var `"NEXUS_HOST=${context.nexusHost}"`. This will not continue to work, but the build environment currently expects to have the scheme present in `NEXUS_HOST` which will not be the case any longer
2. `NEXUS_HOST` is not set anymore in the Snyk stage, but `NEXUS_URL`. The build environment might expect `NEXUS_HOST`. Example: by default, `be-java-springboot` depends on the `NEXUS_HOST` variable to configure Nexus properly
3. `Jenkinsfile`s using `nexusHostWithBasicAuth` will break as the variable is not available anymore - authors will need to change to `nexusUrlWithBasicAuth`. Example: `be-python-flask` uses the variable as an argument to `stageOpenshiftBuild`.

michaelsauter:
ping

metmajer:
@michaelsauter will you create issues for the affected quickstarters?

michaelsauter:
@metmajer No worries. Thanks for reviewing.

Yes I will need to create follow-up issues to change the quickstarters, and we need to docs for converting existing projects to the new style.

I'll create them and then merge this.

michaelsauter:
@metmajer @clemensutschig Bringing this old issue up again. Please see the history above. Given that the change will be breaking, I want to ensure you are still aboard to bring this into 3.0? I can rework this PR so that is it up-to-date and includes docs, and then I'd update the quickstarters. But existing `Jenkinsfile`s would need to be converted by people when they update to 3.0. Let me know what you think.

clemensutschig:
I seriously vouch against this..  this is not needed right now.. and brings more instability

michaelsauter:
I'm fine with it. Moving to v4 then.

michaelsauter:
@clemensutschig I thought a little about it, and I think I found a solution which doesn't harm us right now, but makes the change easier in the future:

1. When users set `NEXUS_HOST` in Jenkins (as they did so far), then we fill both `context.nexusUrl` and `context.nexusHost` with its value (which does contain the scheme). The shared library would print a warning in that case that `context.nexusHost` does contain a scheme.
2. When users set `NEXUS_URL` in Jenkins (which would be the new default in 3.0), then we fill `context.nexusUrl` with scheme and `context.nexusHost` without scheme. The quickstarters would be adapted.

As a result, all newly provisioned quickstarters profit from consistent naming and correct schemes. All current components (provisioned with the old approach) continue to work as-is (with the only limitation that they cannot get just the Nexus host without scheme easily, as is the situation right now already).

@clemensutschig @metmajer Does that sound good to you? Then I'll implement this week.

michaelsauter:
Closing in favour of #366.

PR #185, opened 2020-01-15
Fixing oc rollout status output parser

### LINKED ISSUE #184: oc rollout output parser does not work always
In some situation the output of `oc rollout` is something like:

```
Deployment config "docker-plain-test" waiting on image update
Waiting for latest deployment config spec to be observed by the controller loop...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 of 1 updated replicas are available...
Waiting for latest deployment config spec to be observed by the controller loop...
replication controller "docker-plain-test-1" successfully rolled out
```

And it is not being parsed properly what breaks the build even having a successful deployment.

hugowschneider:
Changed output parser to use precise string lookup instead of splitting strings

This can be referenced back to openshift code  in:
https://github.com/openshift/openshift-controller-manager/blob/master/pkg/apps/deploymentconfig/deploymentconfig_controller.go

/closes #184

### DIFF
--- a/vars/utilsOpenshift.groovy
+++ b/vars/utilsOpenshift.groovy
@@ -96,16 +96,37 @@ OpenshiftDeployment watchRollout(Context context, String targetProject, String c
       ).trim()
       debug(context, rolloutResult)
     }
+
+    /*
+    Considering that the output looks like
+
+    -----------
+    Deployment config "docker-plain-test" waiting on image update
+    Waiting for latest deployment config spec to be observed by the controller loop...
+    Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
+    Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
+    Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
+    Waiting for rollout to finish: 0 of 1 updated replicas are available...
+    Waiting for latest deployment config spec to be observed by the controller loop...
+    replication controller "docker-plain-test-1" successfully rolled out
+    -----------
+
+    and the part that needs extraction is docker-plain-test-1
+     */
+
     // rolloutResult is e.g.: replication controller "foo-123" successfully rolled out
     // Unfortunately there does not seem a more structured way to retrieve this information.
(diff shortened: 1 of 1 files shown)


vars/utilsOpenshift.groovy
```diff
@@ -96,16 +96,37 @@ OpenshiftDeployment watchRollout(Context context, String targetProject, String c
...
-      error "Got '${rolloutResult}' as rollout status, which cannot be parsed properly ..."
-    }
-    def rolloutId = rolloutInfo[1] // part within the quotes
-    if (!rolloutId.startsWith("${componentId}-")) {
+    // It is not posssible to work with Regex because all regex supporting classes are not
+    // serializable
+    def prefix = "replication controller \""
+    def index = rolloutResult.indexOf(prefix)
+    def line = rolloutResult[index + prefix.size()..-1]
+    index = line.indexOf("\" successfully rolled out")
```
michaelsauter: I am not sure this works well in all cases. What if the rollout is NOT successful? We would still want to extract the ID ...

hugowschneider: True, but I thought unsuccessful roll out would have an exited code != 0. So, this needs fixing, do you know how the unsuccessful message looks like?

michaelsauter: I am not sure - it might be that your code actually works for all cases, but this needs to be evaluated properly. All I'm saying is that things might be tricky. This particular piece of code is quite bad and has been problematic before, so we should proceed with caution.

michaelsauter: As an idea ... you actually made two changes:

1. select the line which has `replication controller` in it
2. do not split on quotes

I think your solution would also work if you would have just done (1), right?

By the way, similar stuff is done in https://github.com/opendevstack/ods-jenkins-shared-library/pull/161, maybe worth checking that, too?

PR #188, opened 2020-01-21
Make docker dir configurable

### LINKED ISSUE #181: Be able to configure the docker directory to use when building
Currently the `--from-dir` for `oc start-build` in [stageStartOpenshiftBuild.groovy](https://github.com/opendevstack/ods-jenkins-shared-library/blob/1bd490154235c405bb0d32354ba14a8d25993d37/vars/stageStartOpenshiftBuild.groovy#L13) is fixed to `docker`.
Some build tools already generate a docker directory including a Dockerfile.
Currently, I need to copy stuff over to the `docker` directory.
Additionally, If I choose to use a PV for my build I would need to take care of cleaning up the docker directory again.
Being able to set the docker directory to use one would have more flexibility and might reduce copy/cleanup logic.

It would be nice if I could configure the directory used for `--from-dir`.

none

**Additional context**
none

oalyman:
Resolves #181

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -290,6 +290,9 @@ When you write stages, you have access to both global variables (defined without
 
 | ciSkip
 | Whether the build should be skipped, based on the Git commit message.
+
+| dockerDir
+| The docker directory to use when building the image in openshift
 |===
 
 == Slave customization
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -292,7 +292,7 @@ When you write stages, you have access to both global variables (defined without
 | Whether the build should be skipped, based on the Git commit message.
 
 | dockerDir
-| The docker directory to use when building the image in openshift
+| The docker directory to use when building the image in openshift. Defaults to `docker`.
 |===
 
 == Slave customization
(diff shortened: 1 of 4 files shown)


src/org/ods/OdsContext.groovy
```diff
@@ -436,6 +440,10 @@ class OdsContext implements Context {
     config.displayNameUpdateEnabled = displayNameUpdateEnabled
   }
 
+  String getDockerDir() {
+    return config.dockerDir
+  }
```
michaelsauter: I think you'll need to add this to the `Context` interface too.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -290,6 +290,9 @@ When you write stages, you have access to both global variables (defined without
 
 | ciSkip
 | Whether the build should be skipped, based on the Git commit message.
+
+| dockerDir
+| The docker directory to use when building the image in openshift
```
michaelsauter: Maybe this could be "The Docker context directory to use when building the image in OpenShift. Defaults to `docker`."?

src/org/ods/OdsContext.groovy
oalyman: done.

docs/modules/ROOT/pages/index.adoc
oalyman: done.

PR #191, opened 2020-02-07
Bugfix Mail content when Jenkins build fails #93

### LINKED ISSUE #93: Mail content when Jenkins build fails
When a Jenkins build fails I get an error mail. But the place holder variables like ${script.DEFAULT_SUBJECT}, ${script.DEFAULT_CONTENT}, ... are not replaced by the actual content. 
So all mails have the content ${script.DEFAULT_CONTENT} without pointing to the real issue.

stefanlack:
fixes #93 

However, value of reply-to is still invalid: $script.DEFAULT_REPLYTO

### DIFF
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -216,18 +216,19 @@ class OdsPipeline implements Serializable {
   }
 
   private void notifyNotGreen() {
-    if (context.notifyNotGreen) {
-      script.emailext(
-        body: '${script.DEFAULT_CONTENT}', mimeType: 'text/html',
-        replyTo: '$script.DEFAULT_REPLYTO', subject: '${script.DEFAULT_SUBJECT}',
-        to: script.emailextrecipients([
-          [$class: 'CulpritsRecipientProvider'],
-          [$class: 'DevelopersRecipientProvider'],
-          [$class: 'RequesterRecipientProvider'],
-          [$class: 'UpstreamComitterRecipientProvider']
-         ])
-      )
-    }
+    String subject =  "Build $context.componentId on project $context.projectId  failed!"
+    String body = "<p>$subject</p> <p>URL : <a href=\"$context.buildUrl\">$context.buildUrl</a></p> "
+
+    script.emailext(
+            body: body, mimeType: 'text/html',
+            replyTo: '$script.DEFAULT_REPLYTO', subject: subject,
+            to: script.emailextrecipients([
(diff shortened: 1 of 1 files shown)


stefanlack:
Old - Email was:
image

New - Email is:
image

src/org/ods/OdsPipeline.groovy
```diff
@@ -216,18 +216,19 @@ class OdsPipeline implements Serializable {
...
-        replyTo: '$script.DEFAULT_REPLYTO', subject: '${script.DEFAULT_SUBJECT}',
-        to: script.emailextrecipients([
-          [$class: 'CulpritsRecipientProvider'],
-          [$class: 'DevelopersRecipientProvider'],
-          [$class: 'RequesterRecipientProvider'],
-          [$class: 'UpstreamComitterRecipientProvider']
-         ])
-      )
-    }
+    String subject =  "Build $context.componentId on project $context.projectId  failed!"
```
clemensutschig: add

``` if (context.notifyNotGreen) {  ```

File: src/org/ods/OdsPipeline.groovy [hunk 2]
```diff
@@ -216,18 +216,19 @@ class OdsPipeline implements Serializable {
...
+    String body = "<p>$subject</p> <p>URL : <a href=\"$context.buildUrl\">$context.buildUrl</a></p> "
+
+    script.emailext(
+            body: body, mimeType: 'text/html',
+            replyTo: '$script.DEFAULT_REPLYTO', subject: subject,
+            to: script.emailextrecipients([
+                    [$class: 'CulpritsRecipientProvider'],
+                    [$class: 'DevelopersRecipientProvider'],
+                    [$class: 'RequesterRecipientProvider'],
+                    [$class: 'UpstreamComitterRecipientProvider']
```
clemensutschig: is this ireally correct ... did you check the receipient list?

src/org/ods/OdsPipeline.groovy
stefanlack: ok, added in catch block in org.ods.OdsPipeline#execute (as this is the version we use in the opendevstack installation at our site.)

src/org/ods/OdsPipeline.groovy
stefanlack: I actualy do not check or modify the receipient list. It is unchanged from the previous version.
Please keep in mind that the reciepients list can also contain email adresses outsite your organisation.
For this reason, I am going to configure the _Allowed Domains_ property to restrict the email adresses. (out of scope for this pr)

PR #195, opened 2020-02-15
move build to gradle, testing to spock and CI to github actions

oalyman:
* Use gradle instead of maven
* Use spock for testing
* Use github action workflow instead of travis CI


build.gradle
```diff
@@ -0,0 +1,101 @@
...
+            credentials {
+                username = "${nexus_user}"
+                password = "${nexus_pw}"
+            }
+        }
+    }
+}
+
+dependencies {
+    implementation "org.codehaus.groovy:groovy-all:2.4.12"
```
metmajer: @oalyman any specific reasons you're not using the same dependencies, we use in the MRO shared library?

oalyman: 2.4.12 is the groovy version jenkins supports.

metmajer: @oalyman good finding! Do you want to provide a quick PR for the MRO?

oalyman: Sure. I'll do it in a second

PR #203, opened 2020-03-19
Add sonarqube markdown reports to artifacts

martsec:
This PR changes the stashed SonarQube reports from docx to markdown so THe MRO pipeline is able to generate the PDF for them and attach it to a document. 

Requisite to be able to merge/deploy https://github.com/opendevstack/ods-jenkins-shared-library/pull/203


metmajer:
@martsec I suggest to keep .docx, as it may be expected by other users and add .md as an additional option.

martsec:
> @martsec I suggest to keep .docx, as it may be expected by other users and add .md as an additional option.

True! Changed it to stash both report outputs

PR #221, opened 2020-04-01
Extract stages into classes

### LINKED ISSUE #215: Extract stages into src
Right now the logic of the stages is defined in `vars/`. It would be better to instantiate a new object there, and have it do the logic. That should be easier to test and allows to organise the code better.,

### LINKED ISSUE #216: Ask for lastVersion of build instead of grepping
The latest build can be found by reading `.status.lastVersion` of the `BuildConfig`. This is safer than grepping for it in the output of `oc start-build`.

### LINKED ISSUE #217: Ask for latestVersion of DeploymentConfig instead of grepping for it
The latest rollout can be found by reading `.status.latestVersion` of the `DeploymentConfig`. This is safer than grepping for it in the output of `oc rollout`.

### LINKED ISSUE #218: Configuring podLabel has no effect
**Describe the bug**
When passing an explicit `podLabel`, it has no effect.

See https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/OdsContext.groovy#L166-L168, which is then overwritten later on by https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/OdsContext.groovy#L204.

**Affected version**
 - OpenDevStack 2+

michaelsauter:
That way, it is easier to organise code.

This change also:

* Moves everything related to building/deploying apps into org.ods.component
* Moves `build_service` to `services`
* Rename interfaces to `IContext` etc. as this is more commonly understood as an interface name
* Improves image label sanitisation and adds tests for it
* Makes `OpenShiftService` independent from component context, so that it can be used from other stages (e.g. quickstarter provisioning)
* Adds more tests
* Fixes #218
* Closes #217
* Closes #216
* Closes #215


michaelsauter:
FYI I'm going to go over this once more tomorrow, and look at the merge conflicts. Also, I'm planning to migrate the other stages as well, just want to do this in a separate PR to have a smaller changeset here.

vars/odsPipeline.groovy
```diff
@@ -1,17 +1,5 @@
...
-  def debug = env.DEBUG
-  if (debug != null && config.debug == null) {
-    config.debug = debug
-  }
-
-  def logger = new OdsLogger(this, debug)
-  OdsPipeline pipeline = new OdsPipeline(this, logger)
-
-  pipeline.execute(config, body)
+  odsAppPipeline(config, body)
```
renedupont, line 2: Did you add this odsAppPipeline instead of doing whats in there here because you want to retire odsPipeline at some point or why this additional vars file?

michaelsauter: Yes correct. I will add an echo statement that mentions this.

michaelsauter:
@renedupont @oalyman @clemensutschig @metmajer Quick poll: Do you like `odsAppPipeline` or `odsComponentPipeline` more?

I was leaning towards "app" because it is shorter and less generic, but now I am considering to go with "component" as that is how other parts are referring to it ("componentId", MRO calls it component, quickstarters use the "component" term), and because in a microservices world, "app" might not really reflect what is being built.

clemensutschig:
odsComponentPipeline = for one component

michaelsauter:
@clemensutschig @renedupont @oalyman This is now complete. Could you approve please if fine with you?

oalyman:
@michaelsauter I would not have changed the name of the Context to IContext.
I don't think Interfaces need a special name to signal they are an Interface.
In Groovy it's obvious from the usage (implements).
You also won't find a single Interface ISomething in the SDK.
For me that's like using hungarian notation nowadays.
But just my 2 cents.

michaelsauter:
@oalyman Yes ... I originally did not use `I` but I've seen it many times in Groovy/Java land - and on the other hand, I have seen people confused about the `OdsContext` vs. `Context`. Maybe it's the naming that was wrong. Do you have a better suggestion? Or would you prefer to go back to the old names?

oalyman:
@michaelsauter I would have left it as it was.
But you don't need to change it again, it's naming in the end.

PR #228, opened 2020-04-03
Convert Snyk stage to new style

michaelsauter:
This commit also adapts the behaviour: previously the stage would only
fail if `failOnSnykScanVulnerabilities` was `true` and `snyk test` failed.

Now the stage fails on every error within the stage, such as `snyk auth`
failing, regardless if `failOnSnykScanVulnerabilities` is `true` or not.

Setting `failOnSnykScanVulnerabilities` to `false` only prevents an
error when `snyk test` fails.

Further, the stage now allows to override the `--project-name` flag
(which is defaulted to `componentId`), and it adds more defaults
(`projectId` for `organisation` and `build.gradle` for `buildFile`).

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -340,10 +340,12 @@ class Context implements IContext {
     config.groupId
   }
 
+  @NonCPS
   String getProjectId() {
     config.projectId
   }
 
+  @NonCPS
   String getComponentId() {
     config.componentId
   }
@@ -376,6 +378,7 @@ class Context implements IContext {
     config.sonarQubeBranch = sonarQubeBranch
   }
 
+  @NonCPS
   String getFailOnSnykScanVulnerabilities() {
     config.failOnSnykScanVulnerabilities
   }
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -1,0 +1,88 @@
+package org.ods.component
+
+import org.ods.services.SnykService
+
+class ScanWithSnykStage extends Stage {
+  protected String STAGE_NAME = 'Snyk Security Scan'
+  protected SnykService snyk
+
+  ScanWithSnykStage(def script, IContext context, Map config, SnykService snyk) {
+    super(script, context, config)
+    if (!config.containsKey('failOnVulnerabilities')) {
+      config.failOnVulnerabilities = context.failOnSnykScanVulnerabilities
+    }
+    if (!config.organisation) {
+      config.organisation = context.projectId
+    }
+    if (!config.projectName) {
+      config.projectName = context.componentId
+    }
+    if (!config.buildFile) {
+      config.buildFile = 'build.gradle'
+    }
+    this.snyk = snyk
+  }
+
+  def run() {
+    if (!config.snykAuthenticationCode) {
+      script.error "Option 'snykAuthenticationCode' is not set!"
+    }
+
+    def message = "Snyk scan mode: build will " + (config.failOnVulnerabilities ? "" : "not ") +
+      "fail if vulnerabilities are found (failOnVulnerabilities=${config.failOnVulnerabilities})!"
+    script.echo message
+
+    if (!snyk.version()) {
+      script.error 'Snyk binary is not in $PATH'
+    }
+
+    if (!snyk.auth(config.snykAuthenticationCode)) {
+      script.error 'Snyk auth failed'
+    }
+
+    boolean testSuccess
+
+    def envVariables = [
+      "NEXUS_HOST=${context.nexusHost}",
+      "NEXUS_USERNAME=${context.nexusUsername}",
+      "NEXUS_PASSWORD=${context.nexusPassword}"
+    ]
+    script.withEnv(envVariables) {
+      if (!snyk.monitor(config.organisation, config.buildFile, config.projectName)) {
+        script.error 'Snyk monitor failed'
+      }
+
+      testSuccess = snyk.test(config.organisation, config.buildFile)
+      if (testSuccess) {
+        script.echo 'No vulnerabilities detected.'
+      } else {
+        script.echo 'Snyk test detected vulnerabilities.'
+      }
+    }
+
+    generateAndArchiveReport()
+
+    if (!testSuccess && config.failOnVulnerabilities) {
+      script.error 'Snyk scan stage failed. See snyk report for details.'
+    }
+  }
+
+  private generateAndArchiveReport() {
+    def targetReport = "SCSR-${context.projectId}-${context.componentId}-${snyk.reportFile}"
+    script.sh(
+      label: 'Create artifacts dir',
+      script: 'mkdir -p artifacts/SCSR'
+    )
+    script.sh(
+      label: 'Rename report to SCSR',
+      script: "mv ${snyk.reportFile} artifacts/${targetReport}"
+    )
+    script.archiveArtifacts(artifacts: 'artifacts/SCSR*')
+    script.stash(
+      name: "scrr-report-${context.componentId}-${context.buildNumber}",
+      includes: 'artifacts/SCSR*',
+      allowEmpty: true
+    )
+    context.addArtifactURI('SCSR', targetReport)
+  }
+}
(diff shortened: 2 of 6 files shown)


vars/odsComponentStageScanWithSnyk.groovy
```diff
@@ -0,0 +1,15 @@
+import org.ods.component.ScanWithSnykStage
+import org.ods.component.IContext
+
+import org.ods.services.SnykService
+import org.ods.services.ServiceRegistry
+
+def call(IContext context, Map config = [:]) {
+    def snykService = ServiceRegistry.instance.get(SnykService)
```
renedupont, line 8: Not sure if I missed it but I couldn't find where you add this to the service registry?

vars/stageScanForSnyk.groovy
```diff
@@ -1,91 +1,13 @@
...
-        }
-
-        if (failed && context.failOnSnykScanVulnerabilities) {
-          error "snyk scan stage failed (see snyk report for details) due: " + errorMessage
-        }
-      }
-    }
-  }
+  echo "'stageScanForSnyk' has been replaced with 'odsComponentScanWithSnyk', please use that instead."
+  odsComponentScanWithSnyk(
```
renedupont, line 3: odsComponentStageScanWithSnyk(

renedupont: same for echo 1 line above

vars/odsComponentStageScanWithSnyk.groovy
```diff
@@ -0,0 +1,15 @@
...
+    if (!snykService) {
+        snykService = new SnykService('snyk-report.txt')
```
renedupont, line 10: there is no constructor for this, you probably mean:
`snykService = new SnykService(this, 'snyk-report.txt')`
?

michaelsauter: Fixed in aac6b07.

vars/stageScanForSnyk.groovy
michaelsauter: Fixed in aac6b07.

vars/odsComponentStageScanWithSnyk.groovy
michaelsauter: You are right, I'm not adding it before. Except in the tests ... test induced damage I guess :( Maybe better to test just the class - I'll think about that ...

michaelsauter:
@renedupont @stitakis @oalyman ping

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
+package org.ods.component
+
+import org.ods.services.SnykService
+
+class ScanWithSnykStage extends Stage {
+  protected String STAGE_NAME = 'Snyk Security Scan'
```
stitakis, line 6: is this a constant... what about to make if `public final`?

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+  protected SnykService snyk
```
stitakis, line 7: is this variable used only internally? If yes `private` would be a better option.

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+  ScanWithSnykStage(def script, IContext context, Map config, SnykService snyk) {
```
stitakis, line 9: I see an issue with this constructor implementation:
Map `config` is modified in the constructor. The caller of the constructor would have a map with different state after calling the constructor. This could be misleading side effect.

I think a better approach could be to not modify the map. The values of `failOnVulnerabilities`, `organisation`, `projectName`, `buildFile` could be stored as class fields with `private` scope.

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+      config.organisation = context.projectId
+    }
+    if (!config.projectName) {
+      config.projectName = context.componentId
+    }
+    if (!config.buildFile) {
+      config.buildFile = 'build.gradle'
+    }
+    this.snyk = snyk
+  }
```
stitakis, line 24: a logging line that includes the stage resolved variables could be helpful

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+  def run() {
+    if (!config.snykAuthenticationCode) {
```
stitakis, line 27: This stage should fail without auth code, so the earliest that fails than better.
For this reason, would be better to move this to the constructor and let it fail if no auth code is available?

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+      testSuccess = snyk.test(config.organisation, config.buildFile)
+      if (testSuccess) {
+        script.echo 'No vulnerabilities detected.'
+      } else {
+        script.echo 'Snyk test detected vulnerabilities.'
+      }
+    }
+
+    generateAndArchiveReport()
```
stitakis, line 63: I think we need to add here a `try / catch` block to capture any error and log the details.

src/org/ods/component/Context.groovy
```diff
@@ -340,10 +340,12 @@ class Context implements IContext {
     config.groupId
   }
 
+  @NonCPS
```
stitakis, line 343: What is this annotation for?

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -0,0 +1,88 @@
...
+import org.ods.services.SnykService
+
+class ScanWithSnykStage extends Stage {
+  protected String STAGE_NAME = 'Snyk Security Scan'
+  protected SnykService snyk
+
+  ScanWithSnykStage(def script, IContext context, Map config, SnykService snyk) {
+    super(script, context, config)
+    if (!config.containsKey('failOnVulnerabilities')) {
+      config.failOnVulnerabilities = context.failOnSnykScanVulnerabilities
```
stitakis, line 12: I like the test very much! Cool stuff! :-)
My coverage report reported that there is a missing test case for `!config.containsKey('failOnVulnerabilities')`

src/org/ods/component/ScanWithSnykStage.groovy
stitakis, line 27: Again... I like the test very much! Excellent! :-)
My coverage report reported that there is a missing test case for `!config.snykAuthenticationCode`

src/org/ods/component/Context.groovy
michaelsauter: To be able to call this method from a constructor (Jenkins CPS magic ...).

src/org/ods/component/ScanWithSnykStage.groovy
michaelsauter: You are right. If we change it, we should clone the map though I think. In reality this isn't really a concern as the way you call this is through e.g. `odsComponentStageScanWithSnyk(context, [foo: bar])`, so you don't care about the modification.

src/org/ods/component/ScanWithSnykStage.groovy
michaelsauter: Added as an issue: #238.

src/org/ods/component/ScanWithSnykStage.groovy
michaelsauter: Added as an issue: #238.

src/org/ods/component/ScanWithSnykStage.groovy
stitakis: It might be that today is not a concern because the stage is called as you described. But the usage could change later. A developer that uses the stage will not expect the map to be changed. 
But actually the map is not needed internally. One could just copy from the map the values needed and store them in internal class variable to avoid any potential mistake coding. This is what I mean: 
```
  private String failOnVulnerabilities
  private String organisation
  private String projectName
  private String buildFile

  ScanWithSnykStage(def script, IContext context, Map config, SnykService snyk) {
    super(script, context, config)
    if (!config.containsKey('failOnVulnerabilities')) {
      failOnVulnerabilities = context.failOnSnykScanVulnerabilities
    }
    if (!config.organisation) {
      organisation = context.projectId
    }
    if (!config.projectName) {
      projectName = context.componentId
    }
    if (!config.buildFile) {
      buildFile = 'build.gradle'
    }
    this.snyk = snyk
  }
```

PR #230, opened 2020-04-03
Add quickstarter pipeline

michaelsauter:
Implements a new pipeline that can be used to provision quickstarters.

It uses the same style as the new "component pipeline", with some simplifications. For now, this pipeline has only been used for the Go quickstarter (https://github.com/opendevstack/ods-quickstarters/pull/222). I'll convert the other ones after this PR is merged .. in the process, I might need to do some minor adjustments to the pipeline.

To visualise what a new `Jenkinsfile` looks like, here's the one from Go (updated):
```
def odsGitRef = env.ODS_GIT_REF ?: 'production'
def odsImageTag = env.ODS_IMAGE_TAG ?: 'latest'
library("ods-jenkins-shared-library@${odsGitRef}")

odsQuickstarterPipeline(
  imageStreamTag: "cd/jenkins-slave-golang:${odsImageTag}",
) { context ->

  odsQuickstarterStageCopyFiles(context)

  stage('Write go.mod') {
    dir(context.targetDir) {
      sh "echo 'module example.com/${context.projectId}/${context.componentId}' > go.mod"
      sh "echo '' >> go.mod"
      sh "echo 'go 1.12' >> go.mod"
    }
  }

  odsQuickstarterStageCreateOpenShiftResources(
    context,
    [directory: 'common/ocp-config/component-environment']
  )

  odsQuickstarterStageRenderJenkinsfile(context)

  odsQuickstarterStageRenderSonarProperties(context)
}
```

Some observations:
* I noticed that the way I implemented it, we basically don't need the `IContext` as `Context` is just data access. Thoughts on this?
* ~There's some boilerplate around the image (`dockerRegistry`/`odsImageTag`). I guess we can get rid of it, but might be a bit of magic~
* ~The `quickstarterId` is required right now - that could also be read from the pipeline itself - it has `jenkinsfilePath: "be-golang-plain/Jenkinsfile"` set. I might default to that, and allow people to override?~
* ~Setting up OpenShift resources will likely be similar across most quickstarters. Maybe we can simplify that too - but unsure what the right level of abstraction is. Right now I'm leaving that for later ...~

### DIFF
--- a/src/org/ods/quickstarter/Context.groovy
+++ b/src/org/ods/quickstarter/Context.groovy
@@ -37,8 +37,8 @@ class Context implements IContext {
   }
 
   @NonCPS
-  String getQuickstarterId() {
-    config.quickstarterId
+  String getSourceDir() {
+    config.sourceDir
   }
 
   @NonCPS
@@ -57,8 +57,8 @@ class Context implements IContext {
   }
 
   @NonCPS
-  String getOutputDir() {
-    config.outputDir
+  String getTargetDir() {
+    config.targetDir
   }
 
   @NonCPS
--- a/src/org/ods/quickstarter/Context.groovy
+++ b/src/org/ods/quickstarter/Context.groovy
@@ -10,18 +10,22 @@ class Context implements IContext {
     this.config = config
   }
 
+  @NonCPS
   String getJobName() {
     config.jobName
   }
 
+  @NonCPS
   String getBuildNumber() {
     config.buildNumber
   }
 
+  @NonCPS
   String getBuildUrl() {
     config.buildUrl
   }
 
+  @NonCPS
   String getBuildTime() {
     config.buildTime
   }
--- a/src/org/ods/quickstarter/Context.groovy
+++ b/src/org/ods/quickstarter/Context.groovy
@@ -1,0 +1,79 @@
+package org.ods.quickstarter
+
+import com.cloudbees.groovy.cps.NonCPS
+
+class Context implements IContext {
+
+  private Map config
+
+  Context(Map config) {
+    this.config = config
+  }
+
+  String getJobName() {
+    config.jobName
+  }
+
+  String getBuildNumber() {
+    config.buildNumber
+  }
+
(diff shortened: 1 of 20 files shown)


michaelsauter:
@oalyman @renedupont I've done a few more changes to make it even easier to author quickstarters. Could you provide feedback to? Thx!

src/org/ods/quickstarter/Context.groovy
```diff
@@ -0,0 +1,79 @@
...
+
+  String getBuildUrl() {
+    config.buildUrl
+  }
+
+  String getBuildTime() {
+    config.buildTime
+  }
+
+  @NonCPS
```
henrjk, line 29: Is there a reason why this is only on some methods?

michaelsauter: I forgot them. Added in 4df3615.

michaelsauter:
@clemensutschig @henrjk @oalyman Good to merge this? I'd love to proceed with converting the quickstarters. If you are OK with the "interface" - so the way that quickstarter `Jenkinsfile`s are written using this PR, then I suggest to merge this, and we can open up cleanup / enhancement issues.

src/org/ods/quickstarter/Context.groovy
```diff
@@ -0,0 +1,83 @@
+package org.ods.quickstarter
+
+import com.cloudbees.groovy.cps.NonCPS
+
+class Context implements IContext {
```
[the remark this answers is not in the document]
michaelsauter: @henrjk With `IContext` I can use the type in the code while exchanging the implementation in the tests.

michaelsauter:
@henrjk Thanks!

Actually I really want to have this as a single library, but I have failed to outline the reasoning

* one repo is easier to manage then multiple (in terms of releases, branches, issues ...)
* there are many things that can be shared between the libraries, such as interaction with OpenShift
* having one repo enforces more similarity (in terms of testing approach, dependencies used etc.) which makes it easier for people to work across libraries

PR #232, opened 2020-04-06
Feature/add image labels and fix broken debug

### LINKED ISSUE #229: Support for dynamic injection of /docs folder into Dockerfile and MRO label support
1. If we want to inject metadata into the image - we only have labels, and we generate already a file that we stick into the docker context (https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/BuildOpenShiftImageStage.groovy#L30) 
In order for it to be visible in the image one needs to add a `COPY` directive into the docker file -> touch every quickstarter. And if we want to add more - touch again.

1. If we want to add extension image labels - there is already a feature called *imageLabels* (https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/BuildOpenShiftImageStage.groovy#L11) but this is not useful if we call the shared lib from `load` in the MRO.. 

1. Inject the copy line into the docker file during build
1. In case environment variable exist, named `ods.build.XXXX` add them into the *imageLabels*

there is non, except for (1) touching all quickstarters.

clemensutschig:
closes #229 

if one specifies now in the env `ods.build.x` - this will make it into the extended image labels

### DIFF
--- a/test/groovy/org/ods/component/ContextSpec.groovy
+++ b/test/groovy/org/ods/component/ContextSpec.groovy
@@ -185,7 +185,7 @@ class ContextSpec extends Specification {
     }
 
     Map<String, String> getCloneProjectScriptUrls(config) {
-        def IContext uut = new Context(script, config, logger)
+        def uut = new Context(script, config, logger)
         uut.assemble()
         return uut.getCloneProjectScriptUrls()
     }
--- a/test/groovy/org/ods/component/ContextSpec.groovy
+++ b/test/groovy/org/ods/component/ContextSpec.groovy
@@ -185,7 +185,7 @@ class ContextSpec extends Specification {
     }
 
     Map<String, String> getCloneProjectScriptUrls(config) {
-        def uut = new Context(script, config, logger)
+        def IContext uut = new Context(script, config, logger)
         uut.assemble()
         return uut.getCloneProjectScriptUrls()
     }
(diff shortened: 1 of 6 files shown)


clemensutschig:
@michaelsauter  - I am getting

org.ods.component.ContextSpec > testCloneProjectScriptUrlsDefault FAILED
    groovy.lang.MissingMethodException at ContextSpec.groovy:189

org.ods.component.ContextSpec > testCloneProjectScriptUrlsAtBranch FAILED
    groovy.lang.MissingMethodException at ContextSpec.groovy:189

super weird,.. any ideas?

test/groovy/org/ods/component/ContextSpec.groovy
```diff
@@ -185,7 +185,7 @@ class ContextSpec extends Specification {
     }
 
     Map<String, String> getCloneProjectScriptUrls(config) {
-        def uut = new Context(script, config, logger)
+        def IContext uut = new Context(script, config, logger)
```
michaelsauter, line 188: `def IContext` should be just `IContext` I guess?

clemensutschig: nope .. it looks like it's reusing the same context from above .. and does not find the method. I have no idea - why this ever worked, for now it doesnt :(

PR #236, opened 2020-04-07
Fix build args patches

michaelsauter:
Includes some indentation fixes, 822832d03ca033826640a0167a52225bc9be2153 addresses the actual problem.

Given that we seed the `BuildConfig` with an empty `dockerStrategy` (https://github.com/opendevstack/ods-quickstarters/blob/master/common/ocp-config/component-environment/component-template.yml#L90), I think this is safe enough.

### DIFF
--- a/src/org/ods/services/OpenShiftService.groovy
+++ b/src/org/ods/services/OpenShiftService.groovy
@@ -133,16 +133,14 @@ class OpenShiftService {
       ]}"""
     ]
 
-    // add build args - TODO: Ensure we are not replacing all buildArgs!!
+    // Add build args
     def buildArgsItems = []
     for (def key : buildArgs.keySet()) {
       def val = buildArgs[key]
       buildArgsItems << """{"name": "${key}", "value": "${val}"}"""
     }
     if (buildArgsItems.size() > 0) {
-      def buildArgsPatch = """{"op": "replace", "path": "/spec/strategy/dockerStrategy", "value": {"buildArgs": [
-      ${buildArgsItems.join(",")}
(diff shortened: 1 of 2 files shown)


src/org/ods/services/OpenShiftService.groovy
```diff
@@ -133,16 +133,14 @@ class OpenShiftService {
...
     def buildArgsItems = []
     for (def key : buildArgs.keySet()) {
       def val = buildArgs[key]
       buildArgsItems << """{"name": "${key}", "value": "${val}"}"""
     }
     if (buildArgsItems.size() > 0) {
-      def buildArgsPatch = """{"op": "replace", "path": "/spec/strategy/dockerStrategy", "value": {"buildArgs": [
-      ${buildArgsItems.join(",")}
-    ]}}"""
+      def buildArgsPatch = """{"op": "replace", "path": "/spec/strategy/dockerStrategy/buildArgs", "value": [${buildArgsItems.join(",")}]}"""
```
henrjk, line 143: Will this work on an empty dockerStrategy as the buildArgs would not exist yet?

Would we want to also support a way to merge additional parameters (perhaps in another PR) so that one could define some args in the build config and then add unrelated args during the build?

michaelsauter: Yes, `replace` is equal to `remove` and `add`, and it also works if it does not exist yet.

Your other question is valid too, but a little more complicated. If you feel this to be important, please open a follow-up issue. I'd probably leave it as-is for now ...

PR #237, opened 2020-04-07
Feature/add image labels - the old shell version

clemensutschig:
now people can add extended image labels into the env to pass it to the shared lib. those have to start with `ods.build.`

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -678,9 +678,13 @@ class Context implements IContext {
         label: 'getting extension environment labels'
       ).trim()
     
+    if (rawEnv.trim().size() == 0 ) {
+      return [:]
+    }
+      
     return rawEnv.normalize().split(System.getProperty("line.separator")).inject([ : ] ) { kvMap, line ->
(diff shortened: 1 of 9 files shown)


src/org/ods/component/Context.groovy
```diff
@@ -672,4 +671,21 @@ class Context implements IContext {
...
+    
+    if (rawEnv.trim().size() == 0 ) {
+      return [:]
+    }
+      
+    return rawEnv.normalize().split(System.getProperty("line.separator")).inject([ : ] ) { kvMap, line ->
+        Iterator kv = line.toString().tokenize("=").iterator()
+        kvMap.put(kv.next(), kv.hasNext() ? kv.next() : "")
+        kvMap
+      }
```
michaelsauter, line 689: Is there any possibility that those lines contain more than one `=` sign?

clemensutschig:
follow up fix for #229

src/org/ods/component/Context.groovy
clemensutschig: as we own those ods.build guys, I cannot think of one .. (and the worst thing that could happen is that we cut half of it off)

Issue #213, opened 2020-03-31
MRO: sonarqube branch should be set to '*' in case of multirepobuild = true


clemensutschig:
**Describe the bug**
as the MRO creates branches dynamically - and relies on a source code review report to generated docs .. the shared lib should set `sonarQubeBranch` to `*` automatically.

https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/OdsPipeline.groovy#L133

@metmajer @martsec - fyi. (current workaround .. set to `sonarQubeBranch` to `*` on master before assemble)

michaelsauter:
Is this intended for 2.x or 3.x?

For 3.x, there is a feature which tells the Jenkins master whether multiple branches are supported by the SonarQube edition in use, and based on that modifies the `sonarQubeBranch`.

Maybe we should still force `*` for MRO regardless, but thought I'll mention it here.

PR #240, opened 2020-04-07
WIP: Improve check whether branch should be checked by SonarQube or not

michaelsauter:
Keeping this in WIP until we resolve if we continue with the config file in the Jenkins Master.

FYI @oalyman

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -102,8 +102,8 @@ class Context implements IContext {
     if (!config.containsKey('cloneProjectScriptBranch')) {
       config.cloneProjectScriptBranch = 'production'
     }
-    if (!config.containsKey('sonarQubeBranch')) {
-      config.sonarQubeBranch = 'master'
+    if (config.containsKey('sonarQubeBranch')) {
+      logger.info "Configuring 'sonarQubeBranch' is deprecated, please use 'branch' option of 'odsComponentStageScanWithSonar'."
     }
     if (!config.containsKey('failOnSnykScanVulnerabilities')) {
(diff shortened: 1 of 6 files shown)


michaelsauter:
This should probably replace the solution implemented for https://github.com/opendevstack/ods-jenkins-shared-library/issues/213.

clemensutschig:
@michaelsauter  - shouldnt this now be closed with all the work you have done on SQ?

michaelsauter:
@clemensutschig No, this is slightly different.

The SQ work that is in master allows to decorate pull requests, but the feature is not automatically enabled yet. To enable it, one has to set `branch: '*'`.

This PR resolved, based on the config baked into the Jenkins Master image, whether all branches should be scanned. As a result, scanning would be seamless = on by default when supported, off by default when not supported.

BUT there has been discussion about pushing the Jenkins Master image to Docker Hub / GitHub (https://github.com/opendevstack/ods-core/issues/491), and that would prevent putting the config file into the image.

I want to wait a little to see what we want to do regarding image pushing for 3.0, and then come back to this if still applicable.

clemensutschig:
@michaelsauter - with the current work, I think we are ok to leave this as is .. what do you think?

michaelsauter:
Given that we will not push the Jenkins images to Docker Hub / GitHub, I think it would be nice to complete this. It gives you auto-configuration which branches should be scanned, based on the SQ installation.

michaelsauter:
To avoid problems in the future, I'll not pursue this any further. Config file will be removed in https://github.com/opendevstack/ods-core/pull/590.

PR #242, opened 2020-04-14
Feature/add mono repo support

### LINKED ISSUE #222: Jenkins shared library should support monorepo - multi build / deploy
Today each repo is bound to a single component - that is set at the root of the shared lib `componentId` - and subsequently used in all stages, thru `context.getComponentId()` . 

This is for the clean microservice architecture very useful, but less so in case there is more components in a repo (e.g. in case of a vendor supplied repo, that contains say an application and a database)

The ability to override a component on a stage, e.g. `stageBuildOpenShift(...., "myComponentId")` so people can use multiple containers in one dc later .. (or multi dc) - see https://docs.openshift.com/container-platform/3.7/architecture/core_concepts/pods_and_services.html

copying shared lib code into the jenkins file (e.g. https://github.com/opendevstack/ods-quickstarters/blob/master/ds-ml-service/Jenkinsfile.template#L103-L135) - which is not good.

@metmajer  @gerardcl  @buegelbeatz

clemensutschig:
closes #222 

current features
1. allows multi ocp build (call build stage again with `componentId`)
1. allows multi ocp deploy / multi container (call deployment stage again with `componentId`, or even better design one deployment with multicontainer)
1. returns (in case the new stages are used, build and deployment information - that can be further used e.g. in the ml quickstarter to connect to the pods)

limitations
1. the artifactsURI interface stayed unchanged for backward compatibility -> still a map, and no object classes

`getBuildArtifacts` now returns a complex structure, see below ... 
```
{
    "builds": {
        "flask": {
            "buildId": "flask-61",
            "image": "docker-registry.default.svc:5000/odsst-dev/flask@sha256:8bb152091b6c707dd9e1f9a2869da811e74a606f670c6d93278d6360757e7548"
        }
    },
    "deployments": {
        "flask": {
            "podName": "flask-59-wnvgc",
            "podNamespace": "odsst-dev",
            "podCreationTimestamp": "2020-04-14T10:28:19Z",
            "podEnvironment": "dev",
            "podNode": ".....",
            "podIp": ".........",
            "podStatus": "Running",
            "containers": {
                "flask": "docker-registry.default.svc:5000/odsst-dev/flask@sha256:8bb152091b6c707dd9e1f9a2869da811e74a606f670c6d93278d6360757e7548"
            },
            "deploymentId": "flask-59"
        }
    },
    "SCRR": "SCRR-odsst-flask.docx",
    "SCRR-MD": "SCRR-odsst-flask.md",
    "OCP Build Id": "flask-61",
    "OCP Docker image": "docker-registry.default.svc:5000/odsst-dev/flask@sha256:8bb152091b6c707dd9e1f9a2869da811e74a606f670c6d93278d6360757e7548",
    "OCP Deployment Id": "flask-59",
    "testResults": "1"
}
```

### DIFF
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -48,11 +48,8 @@ class BuildOpenShiftImageStage extends Stage {
     if (buildStatus != 'complete') {
       script.error "OpenShift Build #${lastVersion} was not successful - status is '${buildStatus}'."
     }
-    context.addArtifactURI("OCP Build Id", buildId)
-
     def imageReference = getImageReference()
     script.echo "Build #${lastVersion} has produced image: ${imageReference}."
-    context.addArtifactURI("OCP Docker image", imageReference)
     
     context.addBuildToArtifactURIs (componentId, [ "buildId" : buildId, "image" : imageReference ])
     
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -65,7 +65,7 @@ class BuildOpenShiftImageStage extends Stage {
 
     private String startAndFollowBuild() {
       def dockerContext = config.dockerfile ?: context.dockerDir
-      openShift.startAndFollowBuild(componentId, context.dockerDir)
+      openShift.startAndFollowBuild(componentId, dockerContext)
     }
 
     private String getLastVersion() {
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -56,7 +56,7 @@ class BuildOpenShiftImageStage extends Stage {
     
     context.addBuildToArtifactURIs (componentId, [ "buildId" : buildId, "image" : imageReference ])
     
-    return ["build" : buildId, "image" : imageReference]
+    return ["buildId" : buildId, "image" : imageReference]
   }
 
     private String getImageReference() {
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -20,6 +20,10 @@ class BuildOpenShiftImageStage extends Stage {
     if (!config.buildTimeoutMinutes) {
       config.buildTimeoutMinutes = context.openshiftBuildTimeout
     }
+    if (!config.dockerDir) {
+      config.dockerDir = context.dockerDir
+    }
+    
     this.openShift = openShift
   }
 
@@ -61,8 +65,7 @@ class BuildOpenShiftImageStage extends Stage {
     }
 
     private String startAndFollowBuild() {
-      def dockerContext = config.dockerfile ?: context.dockerDir
-      openShift.startAndFollowBuild(componentId, dockerContext)
+      openShift.startAndFollowBuild(componentId, config.dockerDir)
     }
 
     private String getLastVersion() {
@@ -115,7 +118,6 @@ class BuildOpenShiftImageStage extends Stage {
     }
 
     // Write docker/release.json file to be reachable from Dockerfile.
-    def dockerContext = config.dockerfile ?: context.dockerDir
-    script.writeFile(file: '${dockerContext}/release.json', text: "[\n" + jsonImageLabels.join(",\n") + "\n]")
+    script.writeFile(file: '${config.dockerDir}/release.json', text: "[\n" + jsonImageLabels.join(",\n") + "\n]")
   }
 }
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -54,6 +54,8 @@ class BuildOpenShiftImageStage extends Stage {
     script.echo "Build #${lastVersion} has produced image: ${imageReference}."
     context.addArtifactURI("OCP Docker image", imageReference)
     
+    context.addBuildToArtifactURIs (componentId, [ "buildId" : buildId, "image" : imageReference ])
+    
     return ["build" : buildId, "image" : imageReference]
   }
 
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -30,7 +30,7 @@ class BuildOpenShiftImageStage extends Stage {
     }
 
     def imageLabels = buildImageLabels()
-    writeReleaseFile(imageLabels)
+    writeReleaseFile(imageLabels, config)
     patchBuildConfig(imageLabels)
 
     // Start and follow build of container image.
--- a/src/org/ods/component/BuildOpenShiftImageStage.groovy
+++ b/src/org/ods/component/BuildOpenShiftImageStage.groovy
@@ -41,9 +41,9 @@ class BuildOpenShiftImageStage extends Stage {
     // Retrieve build status.
     def lastVersion = getLastVersion()
     if (!lastVersion) {
-      script.error "Could not get last version of BuildConfig '${context.componentId}'."
+      script.error "Could not get last version of BuildConfig '${componentId}'."
     }
-    def buildId = "${context.componentId}-${lastVersion}"
+    def buildId = "${componentId}-${lastVersion}"
     def buildStatus = getBuildStatus(buildId)
     if (buildStatus != 'complete') {
       script.error "OpenShift Build #${lastVersion} was not successful - status is '${buildStatus}'."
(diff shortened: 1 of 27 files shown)


clemensutschig:
@martsec  - would this cover your sock shop usecase?

clemensutschig:
> Nice! So for the "soft" implementation of using a mono repo I wrote an article on medium.com how this could be done in a generic way for js/nodejs-frameworks: https://medium.com/@christian_becker/isomorphic-client-server-stack-fe348119323f

@buegelbeatz so the current design would make you run?

gerardcl:
ok, I like it! 

but, the focus here though is only focusing on the use case 1 pod N containers, right? I think it woks only when the use case has really a fixed given deployment like that, but, what happens if one wants to scale up the pod? that is not possible with the approach 1 pod N containers if one of the containers is for example a DB, right? in order to assure for example microservice arch pattern the containers should be itself pods from a k8s point of view.

anyway, this works for the current ds-ml-service implementation though since it could be really this: 1 pod 2 containers.

clemensutschig:
@gerardcl - it also supports multiple deployments..  so the ds usecase works as is..

gerardcl:
then  I will test that as is and come back to you

martsec:
> @martsec - would this cover your sock shop usecase?

@clemensutschig  It is hard to me to imagine how to properly configure the component deployment for our use case and this code though. 

Let's say we have two different docker files in the same component. One being the Database and the other one the application. Each one of them could be in a different pod or the same. 
How would we need to set-up the files to build and deploy both containers?

src/org/ods/component/BuildOpenShiftImageStage.groovy
```diff
@@ -41,30 +41,32 @@ class BuildOpenShiftImageStage extends Stage {
...
   }
 
     private String getImageReference() {
-      openShift.getImageReference(context.componentId, context.tagversion)
+      openShift.getImageReference(componentId, context.tagversion)
     }
 
     private String startAndFollowBuild() {
-      openShift.startAndFollowBuild(context.componentId, context.dockerDir)
+      def dockerContext = config.dockerfile ?: context.dockerDir
```
michaelsauter, line 64: I would move that to the `config` assembly above, and would name the option `dockerDir` instead of `dockerfile`: there are actually two things to customise for the image build: "build context" and "Dockerfile". Right now we only allow to customise the context - the file would need to be customised in the `BuildConfig` itself.

src/org/ods/component/BuildOpenShiftImageStage.groovy
```diff
@@ -106,13 +108,14 @@ class BuildOpenShiftImageStage extends Stage {
...
-  private writeReleaseFile(Map imageLabels) {
+  private writeReleaseFile(Map imageLabels, Map config) {
     def jsonImageLabels = []
     for (def key : imageLabels.keySet()) {
       jsonImageLabels << """{"name": "${key}", "value": "${imageLabels[key]}"}"""
     }
 
     // Write docker/release.json file to be reachable from Dockerfile.
-    script.writeFile(file: 'docker/release.json', text: "[\n" + jsonImageLabels.join(",\n") + "\n]")
+    def dockerContext = config.dockerfile ?: context.dockerDir
```
michaelsauter, line 118: Same as above, resolve in constructor

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -171,4 +173,69 @@ class OpenShiftService {
       label: "Get phase of build ${buildId}"
     ).trim().toLowerCase()
   }
+  
+  public String getContainerForImage (String projectId, String rc, String image) {
+    script.sh(
+      script: """oc -n ${projectId} get rc ${rc} -o jsonpath='{.spec.template.spec.containers[?(contains .image "${image}")].name}'""",
```
michaelsauter, line 179: Missing `label` and `returnStdout`

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -171,4 +173,69 @@ class OpenShiftService {
...
+    )
+  }
+  
+  // Gets pod of deployment
+  Map getPodDataForDeployment(String componentId, String rc) {
+    
+    if (!automaticImageChangeTriggerEnabled(componentId)) {
```
michaelsauter, line 186: At least for the shared lib (not the MRO), we do not need an image trigger for deployments to work. If not image trigger is found, a manual rollout is made.

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -171,4 +173,69 @@ class OpenShiftService {
...
+        returnStdout: true,
+        label: "Getting OpenShift pod data for deployment ${rc}"
+      )
+      if (podStatus && podStatus == "Running") {
+        break
+      } else {
+        script.sleep(5000)
+        index--
+      }
+    }
```
michaelsauter, line 205: This has some overlap with `getRolloutStatus`. I think we should settle on one way to check if a deployment is successful or not. At the moment, I'm not sure which way is better though :)

src/org/ods/component/Stage.groovy
```diff
@@ -4,18 +4,20 @@ class Stage {
...
   protected Map config
+  protected String componentId
 
   public final String STAGE_NAME = 'NOT SET'
 
   Stage(def script, IContext context, Map config) {
     this.script = script
     this.context = context
     this.config = config
+    componentId = config.componentId ?: context.componentId
```
michaelsauter, line 15: It took me a while to get why you are not reading from `config.componentId` everywhere, which I expected. TBH I think the usage would become clearer if all those stages would read directly from `config.componentId`. I would even do the setup in each constructor itself - to advertise that this option is available to the stage. Right now, I can change the `componentId` for every stage, even though some stages don't care and still read from `context.componentId`.

src/org/ods/services/OpenShiftService.groovy
clemensutschig: this is for the tests also :) cool to leave in there?

src/org/ods/services/OpenShiftService.groovy
clemensutschig: @michaelsauter  -we don't need a label - I get the rc by name ... stdOut seems the default

src/org/ods/component/Stage.groovy
clemensutschig: well we use it later in the echo ... :(     `script.withStage(STAGE_NAME + ' (' + componentId + ')', context) `

src/org/ods/services/OpenShiftService.groovy
michaelsauter: Oh I meant `label` for the `script` (for better display in Blue Ocean). And while `returnStdout` seems to be the default, we are consistently setting it so I'd continue with it to avoid confusion.

src/org/ods/services/OpenShiftService.groovy
michaelsauter: Not really I think. If we have it here, then we would need to remove the possibility to rollout without it. BUT I am not sure we want to do that. If it weren't for the MRO, I would actually recommend NOT to have any triggers. I am not a fan of this feature of the `DeploymentConfig` resource - plain Kubernetes `Deployment` resources don't have it ... but that is a bigger topic ...

One way to resolve this would be to expect triggers for `DeploymentConfig` resources, but to also support `Deployment` resources (which OpenShift does, and weirdly even recommends in https://docs.openshift.com/container-platform/4.1/applications/deployments/what-deployments-are.html#deployments-comparing-deploymentconfigs_what-deployments-are: "it is recommended to use Deployments unless you need a specific feature or behavior provided by DeploymentConfigs")

buegelbeatz:
> so the current design would make you run

If frontend and backend ran in the same technology, it would be a bit much overhead, but if different technologies were used (JS + Python f.e.), you could avoid creating an ugly hybrid Dockerfile. In this case, it is definitely the more elegant and maintainable solution.

src/org/ods/services/OpenShiftService.groovy
clemensutschig: solved :)

michaelsauter:
I think one can also use multiple build slaves so that you can run each stage on its own tech in case one uses different languages. That way, you can have different Dockerfiles instead of one "ugly" one.

@buegelbeatz What is the overhead you are having with this approach compared to what you describe? I am not sure I am following yet ...

clemensutschig:
> Let's say we have two different docker files in the same component. One being the Database and the other one the application. Each one of them could be in a different pod or the same.
> How would we need to set-up the files to build and deploy both containers?

So the way this works now is the following
1. you have to have two build configs /and image streams
1. in your jenkins file you call the build stage twice (with `componentId` and `dockerDir` set in the config map)
1. you have one DC with two containers (pointing to the images) - done ..

clemensutschig:
I also fixed two more things - in the non mro case - SC* reports are archived now, in the MRO case not..

MRO:
image

Standalone:
image

src/org/ods/services/OpenShiftService.groovy
clemensutschig: @michaelsauter  - rollout != pod ... so I think both have their place?!

michaelsauter: Sure - I was just thinking if we should unify checking the phase. I think the code above is already true when the pod is started, but checking for the rollout is probably only complete when the healthcheck succeeds. Maybe we actually want the latter?

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -171,4 +173,67 @@ class OpenShiftService {
...
+
+    def podOCData = j.items[0]
+
+    // strip all data not needed out
+    def pod = [ : ]
+      pod.podName = podOCData?.metadata?.name?: "N/A"
+      pod.podNamespace = podOCData?.metadata?.namespace?: "N/A"
+      pod.podMetaDataCreationTimestamp = podOCData?.metadata?.creationTimestamp?: "N/A"
+      pod.deploymentConfigCreationEnvironment = podOCData?.metadata?.labels?.env?: "N/A"
```
michaelsauter, line 225: As mentioned in the other PR, I suggest to remove this.

clemensutschig: @michaelsauter  -. what exactly? ...

clemensutschig: removed .env

clemensutschig:
@michaelsauter  / @gerardcl  - cool?

PR #243, opened 2020-04-16
Decorate pull requests in Bitbucket with SonarQube analysis

### LINKED ISSUE #174: Integrate SonarQube with pull requests

michaelsauter:
When a branch is eligible to be scanned, and it is not one of the
long-lived branches, we ask SonarQube to perform PR analysis and
decorate the associated pull request in Bitbucket.

As a side-effect, we now have a Bitbucket token for the cd_user. This is
required as PR decoration is token based. However, a token is good to
have anyway, and we can move other Bitbucket API calls over to the
BitbucketService in the future, making use of the token instead of the
password.

Closes #174.

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -107,7 +107,9 @@ class Context implements IContext {
     if (!config.containsKey('cloneProjectScriptBranch')) {
       config.cloneProjectScriptBranch = 'production'
     }
-    if (!config.containsKey('sonarQubeBranch')) {
+    if (config.containsKey('sonarQubeBranch')) {
+      script.echo "Setting option 'sonarQubeBranch' of the pipeline is deprecated, please use option 'branch' of the stage."
+    } else {
       config.sonarQubeBranch = 'master'
     }
     if (!config.containsKey('failOnSnykScanVulnerabilities')) {
@@ -248,6 +250,7 @@ class Context implements IContext {
     config.gitBranch
   }
 
+  @NonCPS
   String getCredentialsId() {
     config.credentialsId
   }
@@ -387,6 +390,7 @@ class Context implements IContext {
     config.targetProject
   }
 
+  @NonCPS
   String getSonarQubeBranch() {
     config.sonarQubeBranch
   }
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -324,7 +324,8 @@ class Context implements IContext {
     config.nexusHost.replace("://", "://${config.nexusUsername}:${config.nexusPassword}@")
   }
 
-  String getBranchToEnvironmentMapping() {
+  @NonCPS
+  Map getBranchToEnvironmentMapping() {
     config.branchToEnvironmentMapping
   }
 
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -328,7 +328,7 @@ class Context implements IContext {
   }
 
   @NonCPS
-  Map getBranchToEnvironmentMapping() {
+  Map<String, String> getBranchToEnvironmentMapping() {
     config.branchToEnvironmentMapping
   }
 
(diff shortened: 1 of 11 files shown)


src/org/ods/component/Context.groovy
```diff
@@ -324,7 +327,8 @@ class Context implements IContext {
     config.nexusHost.replace("://", "://${config.nexusUsername}:${config.nexusPassword}@")
   }
 
-  String getBranchToEnvironmentMapping() {
+  @NonCPS
+  Map getBranchToEnvironmentMapping() {
```
henrjk, line 331: Perhaps have Map<String, String> here?

src/org/ods/component/ScanWithSonarStage.groovy
```diff
@@ -1,29 +1,46 @@
...
+  ScanWithSonarStage(def script, IContext context, Map config, BitbucketService bitbucket, SonarQubeService sonarQube) {
     super(script, context, config)
-    if (!config.requireQualityGatePass) {
+    if (config.branch) {
+      config.eligibleBranches = config.branch.split(',')
+    } else {
+      config.eligibleBranches = context.sonarQubeBranch.split(',')
+    }
+    if (!config.containsKey('analyzePullRequests')) {
+      config.analyzePullRequests = true
```
renedupont, line 21: true if key does not exist?

src/org/ods/component/ScanWithSonarStage.groovy
```diff
@@ -1,29 +1,46 @@
...
+    }
+    if (!config.containsKey('requireQualityGatePass')) {
       config.requireQualityGatePass = false
```
renedupont, line 24: Can't you do 
`config.requireQualityGatePass = config.containsKey('requireQualityGatePass')`
here? Or do you think it is less readable?

src/org/ods/component/ScanWithSonarStage.groovy
michaelsauter: Yes, this means "enabled by default". I'm checking for the key, otherwise setting it to `false` explicitly would have no effect.

src/org/ods/component/ScanWithSonarStage.groovy
michaelsauter: No, one can set `requireQualityGatePass` explicitly to `false`.

src/org/ods/component/Context.groovy
michaelsauter: Why not, addressed in d5e030652ad774dc361729a2ca78fc87a200e27e.

michaelsauter:
@henrjk @renedupont Feedback addressed, please review again.

vars/odsComponentStageScanWithSonar.groovy
```diff
@@ -1,15 +1,33 @@
...
 import org.ods.component.IContext
 
+import org.ods.services.BitbucketService
 import org.ods.services.SonarQubeService
 import org.ods.services.ServiceRegistry
 
 def call(IContext context, Map config = [:]) {
+    def bitbucketService = ServiceRegistry.instance.get(BitbucketService)
+    if (!bitbucketService) {
+        bitbucketService = new BitbucketService(
```
henrjk, line 11: Can we assume it is created by the pipeline already (Pipeline.groovy)?

renedupont: AFAIK this is done this way so it can be tested autonomously

henrjk:
Somehow my comments have become my review so that it is now closed. 
I just wanted to mention that these changes look good to me and I am approving as well.
However I did not have the cycles to actually play around with this more deeply.

michaelsauter:
@henrjk I think you can always click on "Files changed" on top, and from there you can review/approve again. That is really weird UI, but it works :)

PR #247, opened 2020-04-17
added severity-threshold option for snyk stage

### LINKED ISSUE #166: Add --severity-threshold behavior into stageScanForSnyk
Right now, we can't catch an error only with HIGH cve issues
The only option right now is to shoot down all of them (LOW-MEDIUM-HIGH)

Behavior to add:
Snyk provide a solution for it as instance: "snyk test --file=your-file --severity-threshold=high"

renedupont:
Fixes #166

### DIFF
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -35,7 +35,10 @@ class ScanWithSnykStage extends Stage {
     }
     if(!config.severityThreshold.matches('\\b^(low|medium|high)$\\b')) {
       script.error "'${config.severityThreshold}' is not a valid value for option 'severityThreshold'! Please use low, medium or high."
-    }
+def allowedSeverityThresholds = ['low', 'medium', 'high']
+if(!allowedSeverityThresholds.contains(config.severityThreshold)) {
+       script.error "'${config.severityThreshold}' is not a valid value for option 'severityThreshold'! Please use one of ${allowedSeverityThresholds}."
+     }
 
     script.echo "Scanning for vulnerabilities with " +
       "organisation=${config.organisation}, " +
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -21,13 +21,10 @@ class ScanWithSnykStage extends Stage {
       config.buildFile = 'build.gradle'
     }
     if (!config.severityThreshold) {
-      // low is the default, it is equal to not providing any option to snyk
+      // low is the default, it is equal to not providing the option to snyk
       config.severityThreshold = 'low'
     } else {
       config.severityThreshold = config.severityThreshold.trim().toLowerCase()
-      if(!config.severityThreshold.matches('\\b^(?i:low|medium|high)$\\b')) {
-        script.error "'${config.severityThreshold}' is not a valid value for option 'severityThreshold'. Please use low, medium or high!"
-      }
     }
     this.snyk = snyk
   }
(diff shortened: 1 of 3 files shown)


src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -20,6 +20,11 @@ class ScanWithSnykStage extends Stage {
     if (!config.buildFile) {
       config.buildFile = 'build.gradle'
     }
+    if (!config.severityThreshold) {
+      // low is the default, other possible values are medium and high.
```
michaelsauter, line 24: It would be nice to convert this comment to an error message that is given to the user when the given value is not in the allowed list. That way we serve both developer (reading the code) and user (calling the stage).

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -54,7 +60,7 @@ class ScanWithSnykStage extends Stage {
         script.error 'Snyk monitor failed'
       }
 
-      testSuccess = snyk.test(config.organisation, config.buildFile)
+      testSuccess = snyk.test(config.organisation, config.buildFile, config.severityThreshold)
```
michaelsauter, line 63: Just a question: Is this setting also applicable for the `monitor` command?

renedupont: I tried to find out but did not find any mentioning of this, I am also not sure if it makes sense considering the following description of it:

> snyk monitor can be used from the CLI to create a project on the Snyk website that will be continuously monitored for new vulnerabilities. After running this command you will see it by logging in to the website and viewing Your projects. You should run this command regularly in order to update the dependencies for your project. Consider running it as part of your build process after running snyk test, that way we will always know what dependencies you have and can notify you when any of them become vulnerable. 

Interestingly they write to do test before monitor, we do this currently the other way around, maybe we can swap that in another ticket.

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -20,19 +20,29 @@ class ScanWithSnykStage extends Stage {
...
     this.snyk = snyk
   }
 
   def run() {
     if (!config.snykAuthenticationCode) {
       script.error "Option 'snykAuthenticationCode' is not set!"
     }
+    if(!config.severityThreshold.matches('\\b^(low|medium|high)$\\b')) {
+      script.error "'${config.severityThreshold}' is not a valid value for option 'severityThreshold'! Please use low, medium or high."
+    }
```
michaelsauter, line 38:
```suggestion
def allowedSeverityThresholds = ['low', 'medium', 'high']
if(!allowedSeverityThresholds.contains(config.severityThreshold)) {
       script.error "'${config.severityThreshold}' is not a valid value for option 'severityThreshold'! Please use one of ${allowedSeverityThresholds}."
     }
```

src/org/ods/component/ScanWithSnykStage.groovy
michaelsauter: Oh good find. Could you create another issue for it?

src/org/ods/component/ScanWithSnykStage.groovy
michaelsauter: In general, not sure why I did the check for `snykAuthenticationCode` here instead of the constructor. Might be better to check all config options there?

renedupont: I first had it in the constructor but then you might not find easily in the logs in which stage this error relates too since it happened before the stage start output.

src/org/ods/component/ScanWithSnykStage.groovy
renedupont: https://github.com/opendevstack/ods-jenkins-shared-library/issues/253

Issue #249, opened 2020-04-19
Refactor Pipeline to use services


clemensutschig:
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy

1. https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L217-L248 should be in `bitbucketService`

1. https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L266-L320 should be in `OpenshiftService`

1. https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L182-L215 should be in a new `JunitService` 

@michaelsauter  . can you pick this up - you already started refactoring, and introduced the `BitbucketService` :)

At least the update of the bitbucket commit we will also need in the mro later :)

michaelsauter:
Yes, adding this to my list. I want to do this bit by bit.

PR #255, opened 2020-04-20
Move setting build status to Bitbucket service

michaelsauter:
PR on top of #254 to avoid merge conflicts.

Addresses part of #249.

### DIFF
--- a/src/org/ods/component/Pipeline.groovy
+++ b/src/org/ods/component/Pipeline.groovy
@@ -11,6 +11,7 @@ class Pipeline implements Serializable {
 
   private GitService gitService
   private OpenShiftService openShiftService
+  private BitbucketService bitbucketService
 
   private def script
   private IContext context
@@ -88,6 +89,7 @@ class Pipeline implements Serializable {
               context.projectId,
               context.credentialsId
             ))
+            bitbucketService = registry.get(BitbucketService)
 
             registry.add(OpenShiftService, new OpenShiftService(script, context.targetProject))
           }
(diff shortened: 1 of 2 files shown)


src/org/ods/component/Pipeline.groovy
```diff
@@ -278,33 +280,14 @@ class Pipeline implements Serializable {
     if (!this.bitbucketNotificationEnabled) {
       return
     }
-    if (!context.jobName || !context.tagversion || !context.credentialsId || !context.buildUrl || !context.bitbucketUrl || !context.gitCommit) {
-      logger.info "Cannot set BitBucket build status to ${state} because required data is missing!"
+    if (!context.jobName || !context.tagversion || !context.buildUrl || !context.gitCommit) {
```
clemensutschig, line 283: @michaelsauter  - couldnt we move this into the bitbucket service (setBuildStatus) .. this is a non reusable check .. best case .. we only have one line in the pipeline - (bitbucketNo..?)

src/org/ods/component/Pipeline.groovy
```diff
@@ -278,33 +280,14 @@ class Pipeline implements Serializable {
...
+      logger.info "Cannot set Bitbucket build status to '${state}' because required data is missing!"
       return
     }
 
-    logger.info "Setting BitBucket build status to ${state} ..."
+    logger.info "Setting Bitbucket build status to '${state}' ..."
```
clemensutschig, line 288: see above (and logger.debug) .. in fact I vouch that all services get the logger injected, to finally harmonize this)

src/org/ods/component/Pipeline.groovy
michaelsauter: I left this here on purpose. The service deals with Bitbucket and has no knowledge of the `context`. Of course we can parse every bit into it, but that does not seem right to me. Also, the service has no knowledge of what a build name should be - it is assembled in the pipeline, and therefore the pipeline needs `jobName` and `tagversion` which the service does not need anymore.

src/org/ods/component/Pipeline.groovy
michaelsauter: I'm torn on this. While I have already violated my own rule a few times, I try to avoid giving logging into the service. The service just does its job (interfacing with Bitbucket), the pipeline is in control how it wants to present this to the user. E.g. "failing the build" is a similar thing which the pipeline should control. The service cannot decide whether the caller wants to stop or continue the build if the service encounters an exception.

michaelsauter: I moved this statement into the service now in 5650666.

michaelsauter:
Ok like this?

PR #259, opened 2020-04-21
Bugfix/swap snyk monitor test

### LINKED ISSUE #253: Call snyk monitor after test
I read here https://support.snyk.io/hc/en-us/articles/360000920818-What-is-the-difference-between-snyk-test-protect-and-monitor- that they advise to call 'monitor' after 'test':

> snyk monitor can be used from the CLI to create a project on the Snyk website that will be continuously monitored for new vulnerabilities. After running this command you will see it by logging in to the website and viewing Your projects. You should run this command regularly in order to update the dependencies for your project. Consider running it as part of your build process after running snyk test, that way we will always know what dependencies you have and can notify you when any of them become vulnerable.

Call monitor after test. Currently we do it the other way around

renedupont:
Fixes #253 

- swapped call of monitor with test
- removed project-name param from monitor call since it is incompatible with all-sub-projects flag and lead to not running the monitor call at all. project-name was used to override the default name that is displayed for the project on snyk.io, see further explanation: https://support.snyk.io/hc/en-us/articles/360000910677-Snyk-CLI-monitored-projects-are-created-with-IDs-in-the-project-name
- renamed result var of test script from 'testSuccess' to 'noVulnerabilitiesFound' to understand better
- moved scanning echo closer to actual scanning (snyk.test)

### DIFF
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -42,7 +42,7 @@ class ScanWithSnykStage extends Stage {
         "buildFile=${config.buildFile}, " +
         "failOnVulnerabilities=${config.failOnVulnerabilities}."
 
-    boolean vulnerabilitiesFound
+    boolean noVulnerabilitiesFound
 
     def envVariables = [
       "NEXUS_HOST=${context.nexusHost}",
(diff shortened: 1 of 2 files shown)


michaelsauter:
@stitakis What test are you envisioning here? I agree with writing tests, but what exactly do you want to prove here? Please get in touch with @renedupont and talk about what you want to test.

stitakis:
@michaelsauter not a big vision... just unit test that verify that the scan stage does what is suppose to do. There is some logic there that it is worth to verify.

renedupont:
@stitakis 
Might not be the unit test that you are envisioning, but so far there is this: 
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/test/groovy/vars/OdsComponentStageScanWithSnykSpec.groovy
Is there anything beyond that you see worth testing?

stitakis:
@renedupont ah yes... the spec contains the unit tests that I was expecting. Excellent! I'll approve this change. Btw: I took at look at the coverage, there 3 cases that are not covered in the unit test.

PR #260, opened 2020-04-21
Feature/refactor junit stash

### LINKED ISSUE #257: Refactor pipeline - MRO changes / cleanups
1. MRO does not support cloning the old way anymore (https://github.com/opendevstack/ods-mro-jenkins-shared-library/blob/master/src/org/ods/util/Project.groovy#L505-L515) - so this is dead code: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L222-L229

1. This code should be cleaned up and moved to a JUnitService (similar to MRO)
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L242-L275

clemensutschig:
Closes #257


clemensutschig:
@michaesauter- verified with mro as well

PR #270, opened 2020-04-27
refactor uploadToNexusStage to new stage architecture

### LINKED ISSUE #266: Why dont we infer Project id and component from the git url of the repo instead of coding them into the jenkinsfile?
I gave our contributor quickstarter some more thought.. and in order to prevent changes to the jenkins file and other pieces (sonar) that one would have to change back before an upstream contribution, we could infer project id and component id directly from the `origin remote url` instead of hard coding them into the jenkins file..

the git url pattern is every time similar (because we create it in the prov app):
http(s)://bitbucketHost/scm/`projectId`/projectId-`componentId`.git

See above 

An unmecessary change commit before upstream push.. Error prone

**Context:**
today: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Context.groovy#L252-L260

### LINKED ISSUE #269: Refactor component/uploadToNexus
Upload to Nexus stage looks for a tar.gz File to upload.. which is only so much helpfull.. 

Configurable stage thru config Settings
1. Component id 
1. Project id
1. File

Build inside each jenkinsfile - errorprone

clemensutschig:
closes #269 - make configurable, and add docs

closes #266  - get from git origin url


docs/modules/jenkins-shared-library/pages/odsComponentStageUploadToNexus
```diff
@@ -0,0 +1,19 @@
...
+
+| distributionFile
+| default is `${projectId}-${componentId}.tar.gz`
+
+| repoType
+| default is `candidates`
+
+| groupId
+| default is the `groupId` on project level, or in case not set at all `org.opendevstack.${projectId}`
+|===
```
michaelsauter, line 19: Could you move this to the `partials` folder? (following a recent update)
Also, there needs to be reference from the `component-pipeline.adoc` in order for this partial to show up.

src/org/ods/component/UploadToNexusStage.groovy
```diff
@@ -0,0 +1,23 @@
...
+  public final String STAGE_NAME = 'Upload to Nexus'
+
+  UploadToNexusStage(def script, IContext context, Map config) {
+    super(script, context, config)
+  }
+
+  def run() {
+    def repoType = config.repoType ?: "candidates"
+    def distFile = config.distributionFile ?: "${componentId}-${context.tagversion}.tar.gz"
+    def groupId  = config.groupId ?: context.groupId
```
michaelsauter, line 13: Minor style comment: I think it is a little clearer to have these in the constructor to see immediately what config options I have in the code. But we can leave here too ...

PR #271, opened 2020-04-27
Merge MRO

michaelsauter:
Merges https://github.com/opendevstack/ods-mro-jenkins-shared-library into this repo, preserving the history of the MRO repo.

Closes https://github.com/opendevstack/ods-mro-jenkins-shared-library/issues/104.

As the diff is impossible to review, here's a summary of what I've done:

- Merge commits (d90641c). There were only a few conflicts. I solved the conflict on the `README.md` by creating a new document under `docs` for the orchestration pipeline.
- Rename `mroPipeline` to `odsOrchestrationPipeline` (7037ad8)
- Move all packages from MRO in package `org.ods` into `org.ods.orchestration` (in 596928b, tests moved in c89e93c)
- Move all "phases" in `vars/` to stages in `org.ods.orchestration` (7e30ebe)
- Make use of the super stage to contain the try/catch logic with reporting back to Jira (dadf8df)
- Move `runOnAgentPod` into super stage (660a27b)
- Fix issues that occurred due to the merge (5eb4238, and a regression on MRO `master` in 
313787d)

Far from being complete, but this allows to use the orchestration pipeline like this:
```groovy
@Library('ods-jenkins-shared-library@master') _
odsOrchestrationPipeline(
  debug: true,
  odsImageTag: 'master'
)
```

I've recorded a few follow-up items: #272, #273, #274, #275 and #276 which should cover most things related to the merge of the two libraries.

I propose to merge this quickly to unblock contributions to the MRO again.

PR #283, opened 2020-04-27
Update orchestration pipeline docs

michaelsauter:
Follow-up from #271.

Now they are easily discoverable next to the component pipeline and the
quickstarter pipeline docs. Further, they have been cleaned from any MRO
references.

FYI @metmajer @clemensutschig


michaelsauter:
Rendered version is now live at https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/orchestration-pipeline.html.

PR #289, opened 2020-04-28
bug / remote origin on initial commit fails

clemensutschig:
1. get origin from openshift bc

1. fix in quickstarter pipeline - `targetDir` if not set should point to `componentId`

1. fix in quickstarter pipeline - add nexus vars


src/org/ods/component/Pipeline.groovy
```diff
@@ -370,10 +370,21 @@ class Pipeline implements Serializable {
...
+          returnStdout: true
+        ).trim()
+      } catch (err) {
+        script.dir ("../workspace@script") {
+          origin = script.sh(
+            script: 'git config --get remote.origin.url',
+            returnStdout: true
+          ).trim()
+        }
+      }
```
michaelsauter, line 386: Makes me wonder if we went down a dead-end and it would be better to configure it explicitly. Or maybe there are better ways to extract it (e.g. from OpenShift resources)?

clemensutschig: hmmmm, dude! ... assuming I have an openshift service at hand, I could get it via OC :) let me play with this

michaelsauter: I just remembered that I commented a while back - in different context - how to auto-configure the project ID: https://github.com/opendevstack/ods-quickstarters/issues/42#issuecomment-558631490

clemensutschig: @michaelsauter  - take a look now :)

PR #290, opened 2020-04-29
Apply and enforce codenarc rules

michaelsauter:
Adjust our codebase to be compliant with a huge part of the ruleset.

* More rules should be enabled in the future
* Some exceptions should be removed in the future
* Test codebase is currently not enforcing compliance

BTW., applying the rules already uncovered a few wrong things like unused variables, methods returning different values than promised, private methods that were called from the outside, etc.

I will do more testing tomorrow, but I am merging this now to unblock contributions again.

FYI @clemensutschig @metmajer @oalyman @renedupont

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -14,7 +14,7 @@ jobs:
               with:
                   java-version: 1.8
             - name: Build with Gradle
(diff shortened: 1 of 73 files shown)


metmajer:
Nice, @michaelsauter!

PR #301, opened 2020-04-30
fix Bitbucket not recognizing last successful build after a failed one

### LINKED ISSUE #300: Bitbucket not recognizing last successful build after a failed one
**Describe the bug**
We observed that Bitbucket does not allow merges even if the last executed build was successful.

**To Reproduce**
1. Create a PR in a Bitbucket repo where you need one successful build to merge that PR
2. Make sure that PR fails (or cancel it) while Jenkins is executing the build
3. Manually start the build again and let it run through successfully
4. See that the Merge is still not possible

**Expected behavior**
If the last build was successful we would expect that the merge is possible

**Screenshots**
ods_bug_successful_build_not_recognized

 - OpenShift: 3.11 (should be; will update if not)
 - OpenDevStack 2.x (will update if I have the answer)

gerardcl:
fixes #300

### DIFF
--- a/src/org/ods/OdsPipeline.groovy
+++ b/src/org/ods/OdsPipeline.groovy
@@ -192,13 +192,13 @@ class OdsPipeline implements Serializable {
     if (!context.getBitbucketNotificationEnabled()) {
       return
     }
-    if (!context.jobName || !context.tagversion || !context.credentialsId || !context.buildUrl || !context.bitbucketHost || !context.gitCommit) {
+    if (!context.credentialsId || !context.buildUrl || !context.bitbucketHost || !context.gitCommit) {
       logger.info "Cannot set BitBucket build status to ${state} because required data is missing!"
       return
     }
 
     logger.info "Setting BitBucket build status to ${state} ..."
-    def buildName = "${context.jobName}-${context.tagversion}"
+    def buildName = "${context.gitCommit.take(8)}"
     def maxAttempts = 3
     def retries = 0
     while (retries++ < maxAttempts) {


gerardcl:
fully tested by the way 🎉

gerardcl:
hi!
I also did test only touching the `buildName` param here:
(https://github.com/opendevstack/ods-jenkins-shared-library/blob/29320ed21a60e0e89fdc4ebc94dc4e6f3be8f7be/src/org/ods/OdsPipeline.groovy#L201)

But the problem is that there we already/still use the `tagVersion` which is the one causing the issue. Is your proposal another one?

Regarding concernds about changing the `tagVersion`: I checked also where is that variable used and it seems no issue since basically we want to track the commit that made that build/label/artifact/image. But I agree on creating a more clear issue for that and reference this one.

michaelsauter:
@gerardcl But that place could just make use of the `gitCommit` as well ... that is also exposed on the `context`.

I would be OK with labelling the image with just the commit, but there might be reasons against it, as multiple runs of the same commit might yield different results (e.g. changed external dependencies, changed shared lib, ...)

gerardcl:
totally agreed! 

then let's fix with the commit SHA (taking first 10 chars)

followup question: isn't that then going to diverge a bit the traceability? in the git PR we will be using build IDs different from the tag version... --> anyway this definitely needs a followup issue to be discussed, will create that later on, first the fix

gerardcl:
@michaelsauter  I have been thinking on your last comment, I think it makes sense to keep the `tagVersion`as it is. If you do not see any other requirement I will NOT create a followup ticket from this one.

Issue #304, opened 2020-05-04
Misleading Documentation In Customizing Slaves


sino92:
**Is your feature request related to a problem? Please describe.**
`branchToEnvironmentMapping` is a required parameter based on the recent changes. The documentation needs an update below `projectId`. 
![Screen Shot 2020-05-04 at 12 22 53](https://user-images.githubusercontent.com/16910110/80956757-1a448b80-8e02-11ea-9210-97917a1883d9.png)

michaelsauter:
@sino92 yea, the dots should show that, but you are right it would be good to have that there.

sino92:
https://github.com/opendevstack/ods-jenkins-shared-library/pull/305 I've left out the three dots implying that people might add more of the optional ones.

PR #305, opened 2020-05-04
Make required params more explicit

sino92:
https://github.com/opendevstack/ods-jenkins-shared-library/issues/304

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
@@ -249,16 +249,8 @@ To completely control the container(s) within the pod, set `podContainers`
 Configuring of a customized slave container in a `Jenkinsfile`:
 ----
 odsComponentPipeline(
-  projectId: projectId,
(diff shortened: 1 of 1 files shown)


docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
```diff
@@ -251,6 +251,14 @@ Configuring of a customized slave container in a `Jenkinsfile`:
...
   projectId: projectId,
   podContainers: [
+    projectId: projectId,
+    componentId: componentId,
+    branchToEnvironmentMapping: [
+      "master": "prod",
+      "develop": "dev",
+      "hotfix/": "hotfix",
+      "*": "review"
+    ],
```
michaelsauter, line 261: This is the wrong level, right? It should be outside `podContainers`. Also note that `projectId` and `componentId` are no longer required. 

To focus on the "customisation" aspect, can we just have an empty `branchToEnvironmentMapping`?

`branchToEnvironmentMapping: [:]` is allowed.

PR #309, opened 2020-05-04
Remove DevelopersRecipientProvider from email recipients

### LINKED ISSUE #308: Remove DevelopersRecipientProvider from email list
`DevelopersRecipientProvider` causes all people who are involved in commits in a change to be emailed. The changeset includes commits of the shared library. Therefore shared library authors might get an email if something breaks in a project. I don't think that makes sense, and propose to remove this.

Trying to collect information about the email recipients, it seems to me that `CulpritsRecipientProvider` would still contain the authors of the failing commits, so I think it is fine but the documentation is terrible.

@gerardcl @renedupont @oalyman  Thoughts?

michaelsauter:
Closes #308.

### DIFF
--- a/src/org/ods/component/Pipeline.groovy
+++ b/src/org/ods/component/Pipeline.groovy
@@ -252,7 +252,6 @@ class Pipeline implements Serializable {
             replyTo: '$script.DEFAULT_REPLYTO', subject: subject,
             to: script.emailextrecipients([
                 [$class: 'CulpritsRecipientProvider'],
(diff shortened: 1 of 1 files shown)


gerardcl:
followup proposal: expose an ODS Pipeline context param that can enable users to overwrite default recipients to whatever they define (e.g.: msteams channel inbox), single or multiple ones?

shall we go for that too?

michaelsauter:
@gerardcl Sounds good, but separate issue please.

PR #318, opened 2020-05-06
Feature - use stash instead of byte[] for overall reports, and beautify DTR xml

### LINKED ISSUE #277: DTR should also work with non jira configuration
today the DTR only (Really) works in case jira is configured .. this should also work in the non jira case.

secondly - the xUnit attachments should be formatted. `groovy.xml.XmlUtil.serialize(xmlString)`

### LINKED ISSUE #310: MRO - rework overall report creation (and single store) logic
Today single repo docs (tir/dtr) are put into the json tree that represent the project.. this has two major drawbacks:

1. We cannot dump the tree into the overall docs where it would belong, as the bytes are still there and 
1. we reinvent jenkins std functionality.. namely 'stash

So rather than putting byte[] into the Project json structure we should put in the names of the docs we stash and use that to generate the overall reports.. 

References:
1. Byte[] into Structure: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy#L1143

1. Overall.. byte [] out of Structure: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/usecase/DocGenUseCase.groovy#L83

1. Finalize..  toString() contains the byte[] : https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/FinalizeStage.groovy#L65 

If we remove the document byte[] before the overall doc creation.. it will fail

clemensutschig:
closes #310 
closes #277 

down to 15xx cat 3 findings :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -113,7 +113,7 @@ codenarc {
     configFile = file('codenarc.groovy')
     maxPriority1Violations = 0
     maxPriority2Violations = 0
-    maxPriority3Violations = 1820
+    maxPriority3Violations = 1626
     reportFormat = 'html'
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -113,7 +113,7 @@ codenarc {
     configFile = file('codenarc.groovy')
     maxPriority1Violations = 0
     maxPriority2Violations = 0
-    maxPriority3Violations = 1626
+    maxPriority3Violations = 1514
     reportFormat = 'html'
 }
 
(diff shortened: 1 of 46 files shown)


build.gradle
```diff
@@ -113,7 +113,7 @@ codenarc {
     configFile = file('codenarc.groovy')
     maxPriority1Violations = 0
     maxPriority2Violations = 0
-    maxPriority3Violations = 1820
+    maxPriority3Violations = 1626
```
michaelsauter: Awesome!

PR #325, opened 2020-05-09
Feature/svc common usage (Jenkins / Nexus / GitService)

### LINKED ISSUE #274: Move NexusService
`org.ods.orchestration.service.NexusService` should be moved to `org.ods.services.NexusService`.

### LINKED ISSUE #275: Move JenkinsService
`org.ods.orchestration.service.JenkinsService` should be moved to `org.ods.services.JenkinsService`.

### LINKED ISSUE #285: Merge GitUtil into GitService
`org.ods.orchestration.util.GitUtil` should be merged into `org.ods.services.GitService`. Also, that service needs to be extended with more functionality - no other code should call `git` directly.

clemensutschig:
1. closes #274 
1. closes #275 
1. closes #285 
1. extracts logger into org.ods.util

This PR also ensures that common services are really shared at runtime (and not blindly overwritten in the 'component' pipeline) .. which uncovered a whole set of more issues (Pipelinesteps, NPE,.. :))

Tested both standalone (with flask and also in the MRO context) - both work now :)


clemensutschig:
@martsec  @georgfedermann  - fyi, the only thing that changed are packages ...

src/org/ods/component/Pipeline.groovy
```diff
@@ -60,7 +63,7 @@ class Pipeline implements Serializable {
         }
 
         prepareAgentPodConfig(config)
-        logger.info "***** Starting ODS Pipeline (${config.componentId})*****"
+        logger.info "***** Starting ODS Build Pipeline (${config.componentId}) *****"
```
michaelsauter, line 66:
```suggestion
        logger.info "***** Starting ODS Component Pipeline (${config.componentId}) *****"
```

src/org/ods/orchestration/InitStage.groovy
```diff
@@ -22,15 +26,15 @@ class InitStage extends Stage {
...
+        def git = new GitService(steps)
         git.configureUser()
 
         // load build params
         def buildParams = Project.loadBuildParams(steps)
         steps.echo("Release Manager Build Parameters: ${buildParams}")
 
         // git checkout
-        def gitReleaseBranch = GitUtil.getReleaseBranch(buildParams.version)
+        def gitReleaseBranch = org.ods.services.GitService.getReleaseBranch(buildParams.version)
```
michaelsauter, line 37:
```suggestion
        def gitReleaseBranch = GitService.getReleaseBranch(buildParams.version)
```

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -296,4 +296,11 @@ class OpenShiftService {
         return images
     }
 
+    String getOriginUrlFromBuildConfig (String project, String bcName) {
```
michaelsauter, line 299: We could still use the instance variable `project`, right? I know that how it is used from the pipeline, here the `project` is e.g. `foo-cd` and for the other methods it is e.g. `foo-dev`, but that shouldn't matter. You anyway instantiate a new `OpenShiftService` just to call this method, so you can pass the project to the constructor there ...

michaelsauter: Anyway, I can also do this cleanup later on - I need to work on the OpenShiftService this week anyway.

clemensutschig: @michaelsauter  - I would leave this to you .. to do this in one big shot ..

PR #328, opened 2020-05-11
Feature/changelog update 

### LINKED ISSUE #315: Checking odsComponentPipeline argument input types
**Describe the bug**
Passing an empty list instead of a map type can cause this error. Need to check the object type of user arguments before they make their way to other methods in the shared library. 
Screen Shot 2020-05-05 at 16 32 15
Bug available in master.

clemensutschig:
closes #315

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+### Added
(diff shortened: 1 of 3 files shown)


clemensutschig:
@sino92 - feel free to add more pieces if you discover more .. there is a really interesting case - which we should be aware of ... in a case like this `groovy.lang.MissingMethodException` - it looks like the services are set null ... (unhandled exception groovy goodness?)

clemensutschig:
@sino92  - any more of those?

PR #334, opened 2020-05-13
Implement component deployment using Tailor / OpenShift templates

### LINKED ISSUE #96: (Optionally) Deploy via Tailor

### LINKED ISSUE #273: Merge OpenShiftService
`org.ods.orchestration.service.OpenShiftRegistry` should be removed in favour of `org.ods.services.OpenShiftRegistry`.

### LINKED ISSUE #276: Unify script/steps approach
MRO used an interface, `IPipelineSteps`, the ODS lib used duck-typing. Both works, but we should use just one approach. Previously I thought that the interface approach would not work for e.g. `withCredentials` as that writes variables to the "script", but, one can also read those variables via `env`, which is anyway cleaner.

FYI @renedupont

michaelsauter:
Closes #96.

Supersedes #85.

This PR also contains a few prerequisites:

- It moves `PipelineSteps`/`IPipelineSteps` from `org.ods.orchestration.util` to `org.ods.util`, making it the go-to approach how to access Jenkins pipeline steps from outside the global context. This closes #276 - however some services do not use `steps` yet so we'll have to convert them one by one.
- It merges both `OpenShiftService` classes into one (mainly using the approach from the component pipeline). This closes #273. The new OpenShift service also contains improvements to the image URL deciphering so that image tags are actually resolved and checked in the orchestration pipeline.

Now, to the actual goal of the PR itself:

When a repository contains an `openshift` folder, then the contents of that are assumed to be OpenShift templates that can be applied using Tailor. This support is added into the `odsComponentBuildOpenShiftImage` stage (to control `BuildConfig` and `ImageStream` resources) and the `odsComponentRolloutOpenShiftDeployment` stage (to control all other resources). Both stages offer some control over the Tailor invocation via `config` options, and apart from that all standard Tailor customisations are available (namely via a `Tailorfile` in the `openshift` folder).

One notable change, which also affects deployment without Tailor, is that previously the orchestration pipeline relied on an image trigger being present. This is no longer the case, and the component pipeline doesn't check for triggers anymore at all. The mechanism whether a rollout is started is now based on `.status.latestVersion`. Its value is checked before deployment and after both `tailor apply` and `oc tag` have been run. If it hasn't increased yet, a manual rollout will be performed. This makes deployment a lot more robust.

Other notes:

- Fixes a bug in src/org/ods/orchestration/FinalizeStage.groovy#L61, which used an incorrect attribute access
- Allows to customise the secret name for the secret key for Tailor (via `TAILOR_PRIVATE_KEY_SECRET` - this should eventually probably be read from the `metadata.yml`?

FYI @metmajer @jorge-romero @martsec I hope I did not touch anything that you are working on at the moment - I just had some time waiting for Jenkins jobs to finish so I did some CodeNarc style cleanup :)


michaelsauter:
@clemensutschig For sure.

clemensutschig:
> @clemensutschig For sure.

merge it ... go :)

PR #336, opened 2020-05-14
Various cleanup

### LINKED ISSUE #225: Remove OWASP support
It is only half-finished, and therefore we should not support it. Plus, the same can be achieved with Snyk now.

michaelsauter:
- Removes `stageOWASPDependencyCheck` (closes #225, FYI @renedupont)
- ~Removes the global `withStage`, as that can easily be done with `echo` calls now~ (see c677cac)
- Marks the `Stage` as abstract to make it clearer for stage authors how to use it
- Fixes the GitHub Actions cache setup (FYI @oalyman - after our chat yesterday I read the docs and the `cache` action must be before using `gradle`)

Apart from that, there is a bigger change based on an observation I made while implementing #334.
The PR changes how the OpenShift related stages can be customised to target "non-default OpenShift resources".

The default is that e.g. `BuildOpenShiftImage` looks for a `BuildConfig` of the same name as the `context.componentId`. In the `master` branch, it was possible to pass another `componentId` to the `config` options, which overwrote `context.componentId`.

I now believe this was not ideal because the concept of a "component" is different / more than just the "name" of an OpenShift resource. All resources of a component (= repository) should be labelled with `app=projectId-componentId` even if e.g. two `BuildConfig` resources belonging to it have names other than `componentId`.

Therefore, the PR introduces another mechanism to control the "name" of the resources to target: by passing `name` to the config option (which still defaults to `context.componentId` of course). This approach is also more similar to other stages such as `ScanWithSnykStage` which allows to pass `projectName` but defaults it to `context.componentId`.

It also reads nicely:
```
odsBuildOpenShiftImage(context, [name: 'foo'])
odsRolloutOpenShiftDeployment(context, [name: 'foo'])
```

Of course I'll need to create a follow-up PR in the quickstarters to align them.

UPDATE: The option has been renamed to `resourceName` in a5a2dce.


clemensutschig:
@michaelsauter  - I am not (too) happy with the changed (again) way of working. But if we do it, then call `name` what it is ... a `resource`

michaelsauter:
@clemensutschig Now is the best time to do it since `master` has not been released.

`componentId` is a fixed thing which corresponds to the repository (in regards to Bitbucket) and the umbrella selector (in regards to OpenShift). I believe both things are static can should not be changed during the pipeline. Stages can either use the value directly - or use it to derive a default for more specific configuration.

I didn't think of `resource` ... however I'd still prefer `name`:
- `name` corresponds to the field `.metadata.name` in the resource
- `resource` sounds a lot like it being "one resource" - whereas we set the name for potentially multiple resources, such as `BuildConfig` and `ImageStream` when we build an image ...

renedupont:
I see as a minor disadvantage of removing the withStage.groovy file that custom stages in the Jenkinsfile (like a stage for building the project) can not use this anymore easily.

Is there actually a good way to use the Stage.groovy as a superclass for custom stages?

clemensutschig:
> `resourceName` maybe?

sold :) assuming you change the quickstarters as well .. @martsec - as you already use the monorepo - headsup, the param will change

michaelsauter:
@renedupont Good point.

I haven't tried using it for a custom stage. One thing that we could do, while still not having to keep the global, is to extract it into a class, that can then be imported and used.

renedupont:
@michaelsauter yes sounds ok. 
Alternatively we could create a subfolder (like 'utils') under /vars for functionalities (like withStage) that custom stages could use. Maybe there is more stuff in the future that we could provide to custom stages. I don't know if subfolders under /vars are still globally accessible though.
Or would you prefer importing such things anyway?

michaelsauter:
@renedupont I've decided to take that part of the change out (see c677cac). That way we can merge this PR and discuss the best approach for the `withStage` global separately.

martsec:
> > `resourceName` maybe?
> 
> sold :) assuming you change the quickstarters as well .. @martsec - as you already use the monorepo - headsup, the param will change

@jorge-romero @jafarre-viewnext  @hrcornejo  We wil need to update or monorepo components.

PR #349, opened 2020-05-22
Remove naming reference to jira unexecuted tests.

martsec:
The mandatory tests could be defined elsewhere other than Jira, so the naming of the methods should not contain any reference.

### DIFF
--- a/src/org/ods/orchestration/FinalizeStage.groovy
+++ b/src/org/ods/orchestration/FinalizeStage.groovy
@@ -75,19 +75,19 @@ class FinalizeStage extends Stage {
         levaDocScheduler.run(phase, MROPipelineUtil.PipelinePhaseLifecycleStage.PRE_END)
 
         // Fail the build in case of failing tests.
-        if (project.hasFailingTests() || project.hasUnexecutedJiraTests()) {
+        if (project.hasFailingTests() || project.hasUnexecutedTests()) {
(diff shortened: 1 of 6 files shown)


metmajer:
@martsec can you explain the rationale behind this? Test cases in Jira are defined to be mandatory by the business.

martsec:
> @martsec can you explain the rationale behind this? Test cases in Jira are defined to be mandatory by the business. Do I understand correctly that your intention is only to remove the "Jira" name?

Yes @metmajer , the intention is to remove the Jira name. Since those tests could be (eventually) defined elsewhere, and I just want to avoid "tying" a method in a more "core" position with a use-case name. In the current config, and if you use jira, the mandatory test will come from jira. No changes were made in the behavior.

clemensutschig:
I like it @martsec - but for me  - there is more ...

https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/TestStage.groovy#L118

is a dupe of

https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/BuildStage.groovy#L76

clemensutschig:
@martsec  @metmajer  - what do we do with this now?

martsec:
Since it has been open for a while, and has little to no impact, we could close it. @clemensutschig  @metmajer

PR #352, opened 2020-05-25
Fix various smaller issues

michaelsauter:
* Adds debug statement when project / component is auto-resolved (there was a related failure which I could not reproduce now, but at least one would see more info when debug is enabled now)
* Makes overriding of `projectKey` and `projectName` consistent (otherwise e.g. `ods-doc-gen-svc` QS fails because the behaviour of `sonar-scanner` and `cnesreport.jar` was inconsistent). On a side note, I realised that overriding  `projectKey` is, technically speaking, a backwards incompatible change. However, my understanding is that you want to enforce it, right @clemensutschig?
* Adds a check if the target directory exists. @clemensutschig do you remember why we changed it from `out`, which is what I had initially, to the value of `componentId`? If it doesn't need to be `componentId`, then I actually propose to make a further change and switch back to `out` to be really safe.

### DIFF
--- a/src/org/ods/component/Pipeline.groovy
+++ b/src/org/ods/component/Pipeline.groovy
@@ -431,6 +431,7 @@ class Pipeline implements Serializable {
     }
 
     private void amendProjectAndComponentFromOrigin(Map config) {
+        logger.debug("Amending project / component based on Git origin URL")
         def block = {
             def origin
             try {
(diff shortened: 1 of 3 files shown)


clemensutschig:
> * Adds a check if the target directory exists. @clemensutschig do you remember why we changed it from `out`, which is what I had initially, to the value of `componentId`? If it doesn't need to be `componentId`, then I actually propose to make a further change and switch back to `out` to be really safe.

in the fork case - there was an error, as the directory already existed, I believe ...

PR #356, opened 2020-05-27
Check openshift resources

michaelsauter:
Improves error messages and adds a check for the build config. This is helpful in case sth goes wrong - which is the "standard" case if you try out the prov app quickstarter which doesn't ship the resources required. Users have to do that manually.


clemensutschig:
codenarc :)

PR #358, opened 2020-05-28
Refactor orchestration Project class

### LINKED ISSUE #292: Refactor orchestration Project class
The `Project` class is rather similar to the `Context` of the `component` and `quickstarter` packages. I think it would be good to rename `Project` to `Context`.

Further, once this is done, it would be nice to extract a new class (named `Project`) which deals with the project configuration (loading metadata etc.)

@metmajer Thoughts?

hrcornejo:
Closes #292 

- Renamed "Project" class to "Context" class
- Extracted utility to retrieve metadata, from "Context" class to a new "Project" class


hrcornejo:
FYI @jorge-romero

clemensutschig:
I am confused - I would have extracted all Jira specifics into a class, say 'Jira Project' - that would have inherited from a simple context object - that contains all the magic that is there even w/o jira ... 

for me - the change does not make too much sense right now, but maybe I just don't get it.

hrcornejo:
@clemensutschig the change is not related with Jira. It just a rename of Project class and extract the functionality of read metadata (project metadata -repos, capabilities,... -) to a new class

martsec:
Actually @hrcornejo now that we are at it, we could put the jira-related project details to a class that inherits from a more general "project" one.  And clean a bit the Project/Context part. What do you think? I have always found confusing the think of asking if the jira service is enabled or not.

clemensutschig:
@martsec  - I totally agree,.. I would create a thin context (that has all NO jira pieces in there ..) and a JiraContext? that has all the jira magic in it ... then we initialize once the right guy and are done :)

metmajer:
@martsec @hrcornejo @clemensutschig there is definitely a range of concerns being covered in this class. Renaming alone won't cut it. Let's do this right please and separate those concerns if your goal is to create clean code :)

hrcornejo:
Move the discussion to the issue

metmajer:
Closing for inactivity. Needs a structured approach with a clear separation of concerns.

PR #363, opened 2020-05-29
Fix credentials handling

### LINKED ISSUE #112: proper special characters management within ODS jenkins shared library and bash scripts
This issues is created with the purpose to assure all possible special characters are properly handled within all the steps in ODS jenkins shared library and bash script calls.

See: Auto Clone Environment fails during curl download #109

henrjk:
closes #112 

This must be delivered with an associated PR in ods-core to keep auto cloning working. 

Here a summary of the changes:

- Use `curl --header "Authorization:..."` instead of `--user` to avoid issues with quoting and escaping. I did not have the time to move this to Unirest now.  

- Instead of using `oc create secret` with `--from-literal=password=${password}` the yaml template is written as a file so that there cannot be any issues with weird usernames and passwords as the in the yaml these are base64 encoded.

- Before invoking clone scripts setup a git credential helper with access to the credentials. The details are handled by `git credential-store < file-with-credentials` ,  which handle all details in how these are store and updated. The git code updates or adds an entry and leaves unrelated entries alone. 

Additional notes:
- Credential helpers are also set in another area. I considered changing these to use `GitCredentialStore.setup` but think it should be delivered in a separate PR. Also I am not sure how to test mro scenarios.
- Similar I would propose to consider moving code which uses netrc to enable git operations to instead use `GitCredentialStore.setup`.

### DIFF
--- a/src/org/ods/component/Pipeline.groovy
+++ b/src/org/ods/component/Pipeline.groovy
@@ -4,12 +4,14 @@ import org.ods.services.GitService
 import org.ods.services.BitbucketService
 import org.ods.services.OpenShiftService
 import org.ods.services.ServiceRegistry
+import org.ods.util.GitCredentialStore
 import org.ods.util.ILogger
 import org.ods.util.IPipelineSteps
 import org.ods.util.PipelineSteps
 import org.ods.services.JenkinsService
 import org.ods.services.NexusService
 import groovy.json.JsonOutput
+import org.ods.util.AuthUtil
 
 class Pipeline implements Serializable {
 
@@ -23,10 +25,10 @@ class Pipeline implements Serializable {
     private final IPipelineSteps steps
     private IContext context
     private boolean notifyNotGreen = true
-    private boolean ciSkipEnabled  = true
+    private boolean ciSkipEnabled = true
     private boolean displayNameUpdateEnabled = true
     private boolean localCheckoutEnabled = true
-    private boolean bitbucketNotificationEnabled  = true
+    private boolean bitbucketNotificationEnabled = true
 
     Pipeline(def script, ILogger logger) {
         this.script = script
@@ -35,7 +37,8 @@ class Pipeline implements Serializable {
     }
 
     // Main entry point.
-    @SuppressWarnings(['NestedBlockDepth', 'AbcMetric', 'CyclomaticComplexity', 'MethodSize'])
+    @SuppressWarnings(['NestedBlockDepth', 'AbcMetric', 'CyclomaticComplexity',
+        'MethodSize', 'GStringExpressionWithinString'])
     def execute(Map config, Closure stages) {
         logger.debug "-> ODS Component pipeline setup, debug mode? ${logger.debugMode}"
         if (!!script.env.MULTI_REPO_BUILD) {
@@ -85,7 +88,7 @@ class Pipeline implements Serializable {
                         // in VERY rare (> 7 parallel slaves, sometimes the env.X returns null)
                         def wtfEnvBug = 'null/'
                         if (config.image?.startsWith(wtfEnvBug)) {
-                            script.node ('master') {
+                            script.node('master') {
                                 config.image = config.image.
                                     replace(wtfEnvBug, "${script.env.DOCKER_REGISTRY}/")
                                 logger.warn ("Patched image via master env to: ${config.image}")
@@ -272,7 +275,7 @@ class Pipeline implements Serializable {
     def setupForMultiRepoBuild(def config) {
         logger.info '-> Detected multirepo orchestration pipeline build'
         config.localCheckoutEnabled = false
-        config.displayNameUpdateEnabled  = false
+        config.displayNameUpdateEnabled = false
         config.ciSkipEnabled = false
         config.notifyNotGreen = false
         config.sonarQubeBranch = '*'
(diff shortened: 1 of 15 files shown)


clemensutschig:
Like.. except the build Error..

src/org/ods/component/Pipeline.groovy
```diff
@@ -355,26 +356,34 @@ class Pipeline implements Serializable {
...
                 def branchName = "${script.env.JOB_NAME}-${script.env.BUILD_NUMBER}-${context.cloneSourceEnv}"
                 logger.info "Calculated branch name: ${branchName}"
                 def scriptToUrls = context.getCloneProjectScriptUrls()
                 // NOTE: a for loop did not work here due to https://issues.jenkins-ci.org/browse/JENKINS-49732
                 scriptToUrls.each { scriptName, url ->
-                    script.sh(script: "curl --fail -s --user ${userPass} -G '${url}' -d raw -o '${scriptName}'")
+                    script.echo "curl --fail -s -G '${url}' -d raw -o '${scriptName}'"
+                    script.sh(script: """set +x
+                        curl --fail -s \\
+                        --header \"${AuthUtil.header(AuthUtil.Scheme_Basic, script.USERNAME, script.PASSWORD)}\" \\
```
michaelsauter, line 373: Can you make this `script.env.USERNAME` etc.? I think reading from `env` is the "right way" - and it allows to use an interface for `script` (`IPipelineSteps`).

src/org/ods/component/Pipeline.groovy
```diff
@@ -355,26 +356,34 @@ class Pipeline implements Serializable {
...
+                        -G '${url}' \\
+                        -d raw -o '${scriptName}'""")
                 }
+                GitCredentialStore.setup(script, "https://${context.bitbucketHost}",
```
michaelsauter, line 377: There should be `context.bitbucketUrl`? That would be better instead of assuming the scheme here.

src/org/ods/services/BitbucketService.groovy
```diff
@@ -176,15 +183,34 @@ class BitbucketService {
...
+          kind: Secret
+          type: kubernetes.io/basic-auth
+          metadata:
+            name: ${tokenSecretName}
+            labels:
+              credential.sync.jenkins.openshift.io: 'true'
+        """.stripIndent()
+    }
+
+    static final String BB_TOKEN_SECRET = ".bb-token-secret.yml"
```
michaelsauter, line 202: I guess CodeNarc would want this at the top of the class definition.

src/org/ods/orchestration/Stage.groovy
```diff
@@ -71,6 +71,7 @@ class Stage {
                 def unstashStartTime = System.currentTimeMillis();
                 script.unstash("wholeWorkspace")
                 script.echo "Unstashing workspace took ${System.currentTimeMillis() - unstashStartTime}ms"
+                // should use GitCredentialStore but I am not sure how to test mros.
```
michaelsauter, line 74: Would be nice to do this as well. I can help with how to test this, at least how to do the manual test. Every run of the orchestration pipeline will call this as it's needed to push a commit to the remote.

henrjk:
So far I (and others apparently) have been using double quoted strings (GStrings) routinely as to not distract by thinking about the differences to String. So perhaps we could disable rule UnnecessaryGString?

On the other hand currently we have GStringExpressionWithinString(enabled: false) which I think should probably be enabled.

Also I do not find the traditional sequence of methods public before others as that helpful in many cases. I often rather see private or package methods used by the the public method implementation. So I would not mind disabling rule PublicMethodsBeforeNonPublicMethods as well. 

@michaelsauter what do you think?

michaelsauter:
@henrjk I don't mind about the setting of `UnnecessaryGString` (however with the caveat at the end of this comment).

`GStringExpressionWithinString(enabled: false)` is only disabled because of a false positive - we should probably disable just for that case and enable everywhere else, agreed.

Regarding `PublicMethodsBeforeNonPublicMethods` I would argue that your comment "traditional sequence of methods" shows that this is the common way to do things. In the end, style can be discussed endlessly. I believe we should just stick with whatever is most common in the community ... therefore I'd keep it as is. In line with ["Gofmt's style is no one's favorite, yet gofmt is everyone's favorite."](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=8m43s)

michaelsauter:
@henrjk Yes that is the false positive, and as far as I'm aware that is not a bug - it should be evaluated in the launched pod, not in the one defining the container template.

src/org/ods/component/Pipeline.groovy
henrjk: I addressed this

henrjk:
I made the suggested fixes and changes for codenarc, but I need to test this version as well again including for mro.

henrjk:
@michaelsauter I like a tool enforced style better than getting lost in long discussions about style, even if I dislike some settings. 
 
But we still have hundreds of allowed prio3 violations and some rules disabled, we are not yet in that state and can define the style that makes sense for the shared lib. 
 
My vote would be to disable UnnecessaryGString but enable GStringExpressionWithinString and mark the false positive explicitly.

At any rate these style changes should happen separately from this PR, so that I only tried to correct those which I had introduced before.

src/org/ods/services/BitbucketService.groovy
```diff
@@ -1,9 +1,15 @@
 package org.ods.services
 
 import org.ods.util.ILogger
+import com.cloudbees.groovy.cps.NonCPS
+import org.ods.util.AuthUtil
 
+@SuppressWarnings('PublicMethodsBeforeNonPublicMethods')
```
michaelsauter, line 7: Is this even needed now? Looks like the order adheres to "public before non-public" now ...

henrjk: Without it it complains about: 

Map<String, String> createUserToken() {
[MSG]The public method createUserToken in class org.ods.services.BitbucketService is declared after a non-public method

And then is over the count,  I guess I would have to diff reports in order to figure out where another prio 3 issue came in.

michaelsauter: OK, then we leave it like it is ...

henrjk:
This should now be ready to pull in except that I first will have to ensure that the PR https://github.com/opendevstack/ods-core/pull/601/ will be ready to be merged as well.
 
Summary comments are in commit 44b229335.

PR #375, opened 2020-06-10
fix port in git service URL

georgfedermann:


clemensutschig:
@henrjk  - this did not work in case a port was present - which we uncovered during tests with @georgfedermann

henrjk:
@clemensutschig Thank you for the swift fix!

PR #393, opened 2020-07-01
Re-integrate release branches into main branch

### LINKED ISSUE #367: Orchestration Pipeline: Merge released code back into main branch
Follow up from #364:

> Upon successful deploy to prod merge release/ version branch back to Master (or what was configured in 'metadata.yml' of rm - this is the new starting point for a new Change imho.
I guess the best time to do - is before we go to Q - Q (and P) are holy, to if the merge does not succeed we can break early.

I think we are actually with two distinct issues here. For the sake of simplicity, we assume that the configured branch is `master`.

1. If a developer makes a change only in a release branch, the change would not be included in the next release branch unless explicitly merged to `master` before that next release branch is created. Only in rare cases this actually is desired: if the changed code only applies to code in the release branch, but does not apply to `master`. In general, we want the orchestration pipeline to integrate the changes in the release branch back into `master`. This should probably happen in assemble mode in D, for the reason that any code change on a release branch would need to be assembled again in D. Between Q and P, no code change is allowed. Note that merging the release branch into `master` might fail when there are merge conflicts. In that case, the user must solve them manually, push the merge commit to `master`, and then start assembly again.

2. Code that was on production should not be "lost" the next time we go to production. This should be covered by (1), but it would be good to verify that the release branch contains all code currently on Q/P when deploying a new release branch to Q/P. The check is crucial as one might need to integrate changes from one release branch to another - just integrating into `master` doesn't cut it. You might have a hot fix in a currently released branch when someone else already created a new release branch from `master`. 

A couple of open topics:

- Right now, the orchestration pipeline does not "know" what is in Q/P, making (2) hard to implement. The closest we've got is going through the tags, but it's not enough as we don't know which version was the one prior to the one we are deploying at the moment. Therefore, the release manager needs to keep track of what has been deployed, so that subsequent deploys can look up this information. 
- If there are commits done in the repos during deployment to Q and P, then we also need to re-integrate when deploying to Q and P ...
- When would re-integrate be done best? If we do it at the start of the pipeline, it is "safe" but we might block deploying. If we do it at the end of the pipeline it is "unsafe" but we don't block. If we only needed to re-integrate once (assembly in D) the issue would be smaller.

FYI @martsec @clemensutschig @metmajer Any comments? Ideas?

michaelsauter:
In the finalize stage (assembly), we merge back into the main branch.
That ensures master has all changes of the release branch, ensuring that
newer releases do not accidentally loose changes done on earlier release
branches. However, if two concurrent releases are being worked on, then
it can still happen that the newer release branch does not contain
changes done in the older release branch. This is mitigated by a check
at the pipeline start (init stage). Every promotion to Q or P checks if
the commits that are to be deployed are descendants of commits already
deployed to Q and P. In case of Q, this might be OK (concurrent releases
fighting over Q are allowed), so the build is marked only as unstable.
In case of P, the build fails.

To enable the functionality described above, there is now a directory
"ods-state" in the release manager repository, which, on the master
branch, contains up-to-date information which version, change and
repository commits are deployed in which environment at the moment.

Apart from those changes, there are quite a few other improvements / bug
fixes which were identified in the process:

* Remove repo.data.odsBuildArtifact in favour of repo.data.openshift
* Only run resurrection logic in assemble mode
* Change resurrection detection to look at latest deployed commit in environment
  instead of looking for diff since last deployment descriptor commit.
  It also does not check for deployment version anymore. The better
  check would be to compare the exports, but that proved to be difficult
  with the agent setup we have right now. In any case, the export is
  always happening in the finalize stage, even when the component was
  resurrected so skipping this check seems fine to me.
* Change format of deployment descriptor (new top-level “deployments” key)
* Extracted logic of deployment descriptor in its own class
* Extract logic of deploying/finalizing ODS components in its own class
* Move more Git tasks into GitService
* Move hardcoded name of image puller secret to project to make it easier in the
  future to cusomtize this
* Speed up retrieval of pod data (previously it needed at least two API
  calls, now just one)
* Push all commits together in the finalize stage - this helps to
  prevent inconsistent tags between repositories. However, it still
  isn't an atomic transaction. To prevent pipelines from getting stuck
  (as the tag of a repository is not writable because it already
  exists), the Git tag used for promotion now includes the Jenkins build
  number of the assembling job.
* Introduce new pipeline parameter 'rePromote'. Using this flag it is
  possible to promote a tag to Q again that was previously already
  deployed. Re-deployment to P is never possible.

Closes #367.

### DIFF
--- a/src/org/ods/orchestration/FinalizeStage.groovy
+++ b/src/org/ods/orchestration/FinalizeStage.groovy
@@ -8,8 +8,11 @@ import org.ods.services.BitbucketService
 import org.ods.services.OpenShiftService
 import org.ods.services.GitService
 import org.ods.util.PipelineSteps
+import org.ods.util.IPipelineSteps
 import org.ods.util.Logger
 import org.ods.util.ILogger
+
+import groovy.json.JsonOutput
 
 class FinalizeStage extends Stage {
 
@@ -26,6 +29,7 @@ class FinalizeStage extends Stage {
         def os = ServiceRegistry.instance.get(OpenShiftService)
         def util = ServiceRegistry.instance.get(MROPipelineUtil)
         def bitbucket = ServiceRegistry.instance.get(BitbucketService)
+        def git = ServiceRegistry.instance.get(GitService)
         ILogger logger = ServiceRegistry.instance.get(Logger)
 
         def phase = MROPipelineUtil.PipelinePhases.FINALIZE
@@ -51,23 +55,28 @@ class FinalizeStage extends Stage {
         def agentCondition = project.isAssembleMode && repos.size() > 0
         runOnAgentPod(agentCondition) {
             // Execute phase for each repository - here in parallel, all repos
-            Map allRepos = [ : ]
+            Map repoFinalizeTasks = [ : ]
             util.prepareExecutePhaseForReposNamedJob(
                 phase, repos, preExecuteRepo, postExecuteRepo).each { group ->
-                allRepos << group
+                repoFinalizeTasks << group
             }
+            repoFinalizeTasks.failFast = true
+            script.parallel(repoFinalizeTasks)
 
-            allRepos.failFast = true
-            script.parallel(allRepos)
-
-            // record release manager repo state
             if (project.isAssembleMode && !project.isWorkInProgress) {
-                util.tagAndPushBranch(project.gitReleaseBranch, project.targetTag)
+                integrateIntoMainBranch(steps, git)
             }
+
+            gatherCreatedExecutionCommits(steps, git)
+
+            if (!project.buildParams.rePromote) {
+                pushRepos(steps, git)
+                recordAndPushEnvState(steps, logger, git)
+            }
+
             // add the tag commit that was created for traceability ..
-            GitService gitUtl = ServiceRegistry.instance.get(GitService)
             logger.debug "Current release manager commit: ${project.gitData.commit}"
-            project.gitData.createdExecutionCommit = gitUtl.commitSha
+            project.gitData.createdExecutionCommit = git.commitSha
         }
 
         if (project.isAssembleMode && !project.isWorkInProgress) {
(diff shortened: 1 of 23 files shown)


metmajer:
@michaelsauter I'll need a few days to dig through. Quick comment: some teams use "dev" as an integration branch. Will need to check that such custom flows remain supported (will be if you're using the branches configured in metadata.yml).

michaelsauter:
@metmajer I know this is a lot to digest. Sorry for this. One reason that the diff is huge is that I moved some code out of `MROPipelineUtil` to `DeployOdsComponent` and `FinalizeOdsComponent`. To your question: I use the branches configured in metadata (`repo.branch`). Only for the release manager repository, `master` is hardcoded (as it is already hardcoded in the pipeline config) ... potentially we could read that from the branch of the pipeline run, but I am not sure we really want this ... so I kept it simple with `master` being the main branch of the RM repo.

metmajer:
@michaelsauter makes sense, and the component branches were what I was thinking of only. Thanks for the quick update, I'll try my best to have a look by the end of the week.

src/org/ods/orchestration/FinalizeStage.groovy
```diff
@@ -115,4 +124,107 @@ class FinalizeStage extends Stage {
...
+
+        if (project.isWorkInProgress) {
+            git.pushRef('master')
+        } else {
+            // We don't need to merge, we simply commit the env file. That
+            // avoids unnecessary merge conflicts.
+            git.switchToOriginTrackingBranch('master')
+            git.checkoutAndCommitFiles(
+                project.gitReleaseBranch,
+                [project.envStateFileName],
```
martsec, line 219: In case we need more files to be merged back in the MRO repo, this should be the place to add it, right?

michaelsauter: In theory, yes. But note that I'm actually *not* merging the release branch of the release manager repository back into the `master` branch. I'm directly committing one file (the env state) on `master`. This is done to avoid merge conflicts. Please check if you actually want to have a merge of the release manager ... I didn't do it out of fear it might create unnecessary merge conflicts (e.g. in `dev` state) ... and at the same time, I did not see a use case where one really needs the merge back into the main branch.

martsec: Yes, in fact this sort of works with the plan I had to store the jira data for a given version. Saving one file per version to avoid potential merge conflicts as much as possible. So actually it looks like adding the file there would be enough to bring the information to master branch. Great!

martsec:
lgtm. Great job!

Issue #391, opened 2020-06-28
For agents image configuration - check availability of image & set timeout to connect


sino92:
**Is your feature request related to a problem? Please describe.**
When the ods-pulled image isn't present in a reachable image stream Jenkins keeps looking for it forever. 

**Describe the solution you'd like**
To avoid this, we can introduce a timeout during `podTemplate` or `containerTemplate` configurations. The timeout functionality can be found in the kubernetes [plugin](https://github.com/jenkinsci/kubernetes-plugin) with propertyName: `slaveConnectTimeout`

**Describe alternatives you've considered**
An additional feature would be to check the availability of the entered image name/version. I.e. `jenkins-slave-maven:4.x.` doesn't exist in ODS. Upon receiving this, we can log a warning message saying its not present in ODS library but it might be a non-ODS image present in developer's private registry. If that's not the case at least the logs will help the developer pin down the source of the problem: the timeout happened because the image doesn't exist.

clemensutschig:
@sino92  - love this .. can you make this happen (at least part 1) for ODS 3 ... this would be great!

michaelsauter:
That would be great! Also, if the pod fails to come up, exposing the events would be great! See #353 - the code to do that is 99% there already. If you choose to add this as well, feel free to reach out to me.

sino92:
@michaelsauter Thanks for letting me know that newer Jenkins version displays the pod log output. 
Here's an example of what we get in the logs for a timeout. Will catch the error and display a message about the lack of the searched image for now.
```
bric8.kubernetes.client.KubernetesClientTimeoutException: Timed out waiting for [10000] milliseconds for [Pod] with name:[pod-57752291-0efa-4e27-9c31-d7e37e011fad-31mfv-kjgkd] in namespace [foo-cd].
    at org.csanchez.jenkins.plugins.kubernetes.AllContainersRunningPodWatcher.await(AllContainersRunningPodWatcher.java:130)
    at org.csanchez.jenkins.plugins.kubernetes.KubernetesLauncher.launch(KubernetesLauncher.java:131)
    at hudson.slaves.SlaveComputer.lambda$_connect$0(SlaveComputer.java:291)
    at jenkins.util.ContextResettingExecutorService$2.call(ContextResettingExecutorService.java:46)
    at jenkins.security.ImpersonatingExecutorService$2.call(ImpersonatingExecutorService.java:71)
[...]

PR #395, opened 2020-07-02
Add timeout to podtemplate of component timeout

sino92:
#391

### DIFF
--- a/src/org/ods/component/Pipeline.groovy
+++ b/src/org/ods/component/Pipeline.groovy
@@ -207,6 +207,7 @@ class Pipeline implements Serializable {
                 containers: config.podContainers,
                 volumes: config.podVolumes,
                 serviceAccount: config.podServiceAccount,
+                slaveConnectTimeout: 240 // in seconds
             ) {
                 script.node(config.podLabel) {
                     try {


src/org/ods/component/Pipeline.groovy
```diff
@@ -207,6 +207,7 @@ class Pipeline implements Serializable {
                 containers: config.podContainers,
                 volumes: config.podVolumes,
                 serviceAccount: config.podServiceAccount,
+                slaveConnectTimeout: 15
```
michaelsauter, line 210: Is that minutes or seconds ? If seconds, I think it's too short, if minutes, too long I'd say 😄 10 minutes would still be plenty I think.

sino92: thats seconds. how about a min?

michaelsauter: OK. I think I need to understand the meaning of the setting better then. When does the timeout start? When creating the pod, or when the pod is running? If it starts when the pod is created, it needs to be longer than one min I think. If it starts once it is running, then it can be one min maybe. That said, I'd err on it being longer - say 2-3 mins.

clemensutschig: so on slow clusters pull image and start > 1 minute ...

sino92: Countdown begins before the pod is initiated. If the agent isn't connected to pod, then the pod never starts, the timeout takes place. Here's countdown [block](https://github.com/jenkinsci/kubernetes-plugin/blob/35682a458529e42e94c2a905b21f935dfaca7758/src/main/java/org/csanchez/jenkins/plugins/kubernetes/KubernetesLauncher.java#L144)..
Now that I remember, cd projects that are loaded with builds were taking around a min to connect. Maybe I am just too impatient to wait 4 mins for the pod to connect to agents, I'd stop the pipeline thinking somethings wrong.

sino92: Since you guys vouch for 4 mins, I take you've experienced successful connections that happened past a min or two

PR #411, opened 2020-07-10
Allow DeploymentConfigs with containers pointing to remote images

### LINKED ISSUE #407: Rollout should not fail when containers point to images from Docker Hub directly
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy#L93 fails (on master). It should show that there is a problem, but not fail.

michaelsauter:
For a component, that is OK and should not fail. The orchestration
pipeline still has a check that all images are owned, which is unrelated
to the fixed check here.

Fixes #407.

FYI @georgfedermann @clemensutschig This is the error you ran into when you named the component `golang`. While in that case, you did not actually want to deploy an image from DockerHub, generally it is a valid use-case to deploy something from DockerHub. In that case, you just want to rollout, but not touch the image (e.g. not tag it with `latest`). Therefore the fix here is needed.

### DIFF
--- a/src/org/ods/services/OpenShiftService.groovy
+++ b/src/org/ods/services/OpenShiftService.groovy
@@ -487,9 +487,10 @@ class OpenShiftService {
         def urlParts = url.split('/').toList()
 
         if (urlParts.size() < 2) {
-            throw new RuntimeException(
-                "ERROR: Image URL ${url} must have at least two parts (repository/reference)"
-            )
+            logger.debug "Image URL ${url} does not define the repository explicitly."
+            imageInfo.repository = ''
+        } else {
+            imageInfo.repository = urlParts[-2]
         }
 
         if (urlParts.size() > 2) {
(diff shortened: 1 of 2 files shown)


src/org/ods/services/OpenShiftService.groovy
```diff
@@ -487,9 +487,10 @@ class OpenShiftService {
         def urlParts = url.split('/').toList()
 
         if (urlParts.size() < 2) {
-            throw new RuntimeException(
-                "ERROR: Image URL ${url} must have at least two parts (repository/reference)"
-            )
+            logger.debug "Image URL ${url} does not define the repository explicitly."
```
renedupont, line 490: Was the previous error message not more precise in case the URL does contain the repository but not the reference (for whatever reason)? Or is that not possible for some reason?

michaelsauter: Previously calling this method required the image to look like `<repository>/<name>` (`name` is called `reference` in the error message). If the image did not have two parts, an exception was thrown. Now we also allow `<name>` - in that case the debug message is printed, telling us that the repository is not defined.

michaelsauter: And yes, there always has to be a name ...

PR #420, opened 2020-07-21
Allow general fork

michaelsauter:
* Makes testing a little easier as one can point the quickstarter to a repository on the Bitbucket instance
* Pushes the branch that exists locally, instead of the one given as a param. The stage to push to the remote is not exposed, so users can't set the branch anyway ... better to just take the branch that has been created.


clemensutschig:
@Michaelsauter - this is a must fix for 3..

PR #421, opened 2020-07-21
Switch pod label generation to uuid

sino92:
https://github.com/opendevstack/ods-provisioning-app/issues/540

### DIFF
--- a/src/org/ods/quickstarter/Pipeline.groovy
+++ b/src/org/ods/quickstarter/Pipeline.groovy
@@ -128,7 +128,7 @@ class Pipeline implements Serializable {
             ]
         }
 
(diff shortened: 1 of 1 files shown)


src/org/ods/quickstarter/Pipeline.groovy
```diff
@@ -128,7 +128,7 @@ class Pipeline implements Serializable {
             ]
         }
 
-        def podLabel = "quickstarter-${config.sourceDir}-${config.projectId}-${config.componentId}"
+        def podLabel = "pod-${UUID.randomUUID().toString()}"
```
michaelsauter, line 131: Actually, it would be good to still distinguish it from other pods visually in the UI, so maybe:
```suggestion
        def podLabel = "qs-${UUID.randomUUID().toString()}"
```

PR #430, opened 2020-07-30
Fix for #429

martsec:
Solution found: Filtering-off the `ods-test` repos from the `pushRepos` and `integrateIntoMainBranch` closures. 

Please @michaelsauter check this out and see if it makes sense.  I've tested the official release to D and the promotion to Q and it seems to work well. Even for the git tags (where there was another bug when tried to add the same tag twice in ods-test type)

Output of release to dev
fix-bug-rm-merge-back-test-component-1

Release manager component
fix-bug-rm-merge-back-test-component-label-release-manager

Code component
fix-bug-rm-merge-back-test-component-label-ods-code

Test component
fix-bug-rm-merge-back-test-component-label-ods-test

Fixes #429

PR #431, opened 2020-07-31
Remove special tagging logic for ods-test type

michaelsauter:
Alternative to #430, which should achieve the same thing, with the following differences:

* It will fail if the tag exists already on the remote. This is the same behaviour as for `ods` components then. It should fail as the tag is unique per pipeline run so if it already exists, something is fishy.
* We apply the same fix for the make/shell type components
* Log if make/shell does nothing
* We end up with less code as we treat `ods-test` not as a special case, but the same as `ods` components, and can delete two methods from `GitService`

### DIFF
--- a/src/org/ods/orchestration/util/MROPipelineUtil.groovy
+++ b/src/org/ods/orchestration/util/MROPipelineUtil.groovy
@@ -337,16 +337,6 @@ class MROPipelineUtil extends PipelineUtil {
                     } else if (repo.type?.toLowerCase() == PipelineConfig.REPO_TYPE_ODS_TEST) {
                         if (name == PipelinePhases.TEST) {
                             executeODSComponent(repo, baseDir)
-                        } else if (this.project.isPromotionMode && name == PipelinePhases.FINALIZE) {
-                            if (!this.project.buildParams.rePromote) {
-                                this.steps.dir(baseDir) {
-                                    git.tagAndPush(this.project.targetTag)
-                                }
-                            }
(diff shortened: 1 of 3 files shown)


src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -337,16 +337,6 @@ class MROPipelineUtil extends PipelineUtil {
...
-                                this.steps.dir(baseDir) {
-                                    git.tagAndPush(this.project.targetTag)
-                                }
-                            }
-                        } else if (this.project.isAssembleMode && !this.project.isWorkInProgress && name == PipelinePhases.FINALIZE) {
-                            this.steps.dir(baseDir) {
-                                git.tagAndPushBranch(this.project.gitReleaseBranch, this.project.targetTag)
-                            }
                         } else {
                             this.logger.debug("Repo '${repo.id}' is of type ODS Test Component. Nothing to do in phase '${name}' for target environment'${targetEnvToken}'.")
```
martsec: This debug message won't be true anymore.

michaelsauter: I think it actually better reflects what is happening. There is nothing that is being "done" in the "finalise" phase. Only some git management later on, which I would consider auxiliary. In my eyes, this aligns better with e.g. how we view document generation: it also says "nothing to do" for phases were we actually generate docs - but the doc generation is also happening outside this method.

clemensutschig:
@martsec what are we doing Not? This One or yours.. ?

michaelsauter:
@clemensutschig @martsec  will run this PR to see if there are any issues on his side.

PR #433, opened 2020-07-31
Do not fail when secret cannot be created

### LINKED ISSUE #432: Concurrency issue in component library bitbucket token creation
**Describe the bug**
8 concurrent quickstarter tests will fail with 
```
+ oc -n unitt-cd create -f .bb-token-secret.yml
Error from server (AlreadyExists): error when creating ".bb-token-secret.yml": secrets "cd-user-bitbucket-token" already exists
```

**To Reproduce**
run quickstarter tests in parallel with -p 8

**Expected behavior**
No error - if token exists.. use it

 - OpenDevStack 3

michaelsauter:
We might not be able to create it because it has been created already by
another job at the same time. This is super unlikely to happen in real
life, but does happen during parallel tests.

Fixes #432.

The script errors later anyway if the Jenkins credential cannot be
found (see https://github.com/opendevstack/ods-jenkins-shared-library/blob/e63d0976d42aad08fddfebfbd8fbdbad34effbf0/src/org/ods/services/BitbucketService.groovy#L240-L248).

### DIFF
--- a/src/org/ods/services/BitbucketService.groovy
+++ b/src/org/ods/services/BitbucketService.groovy
@@ -296,10 +296,14 @@ class BitbucketService {
             file: BB_TOKEN_SECRET,
             text: secretYml
         )
-        script.sh """
-            oc -n ${openShiftCdProject} create -f ${BB_TOKEN_SECRET};
-            rm ${BB_TOKEN_SECRET}
+        try {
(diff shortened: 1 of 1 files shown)


clemensutschig:
@georgfedermann - want to try..  ?

PR #438, opened 2020-08-14
Fix branch prefix matching for Snyk and Sonar Stage classes

tobnee:
When calling:

`odsComponentStageScanWithSnyk(context, [snykAuthenticationCode: snykAuthenticationCode, failOnVulnerabilities: false, branch: 'develop, release/, master'])`

within a `release/foo` branch, the documentation suggests, that a scan should be done but it is not.

### DIFF
--- a/src/org/ods/component/Stage.groovy
+++ b/src/org/ods/component/Stage.groovy
@@ -36,10 +36,12 @@ abstract class Stage {
             return true
         }
         // Check if prefix (e.g. "release/") is allowed
-        eligibleBranches.each { eligibleBranch ->
+        if (eligibleBranches.any { eligibleBranch ->
             if (eligibleBranch.endsWith('/') && branch.startsWith(eligibleBranch)) {
                 return true
             }
+        }) {
+            return true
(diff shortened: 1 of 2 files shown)


src/org/ods/component/Stage.groovy
```diff
@@ -36,10 +36,12 @@ abstract class Stage {
             return true
         }
         // Check if prefix (e.g. "release/") is allowed
-        eligibleBranches.each { eligibleBranch ->
+        if (eligibleBranches.any { eligibleBranch ->
```
michaelsauter, line 39: Right!

Maybe we can shorten now though to:
```groovy
if (eligibleBranches.any { it.endsWith('/') && branch.startsWith(it) } {
  return true
}
```

clemensutschig:
cool - that needs to go both into 3.x and master - once comments addressed @tobnee :D

michaelsauter:
@tobnee Perfect. Could you also open this PR against `master` so that it does not get lost in ODS 4?

Issue #425, opened 2020-07-29
Wrong build status fetched on parallel builds


segfault16:
**Describe the bug**
If two branches are built at the same time the wrong build information status is fetched and results in a failed pipeline, e.g.

````
+ oc -n biob-dev start-build frontend --from-dir frontend --follow
Uploading directory "frontend" as binary input for the build ...
.
Uploading finished
[Pipeline] echo
build.build.openshift.io/frontend-592 started
Receiving source from STDIN as archive ...
````
After the docker image was successfully built:
````
Push successful
[Pipeline] }
[Pipeline] // timeout
[Pipeline] sh (Get lastVersion of BuildConfig frontend)
[biob-cd-biobase-core-844] Running shell script
+ oc -n biob-dev get bc/frontend -o 'jsonpath={.status.lastVersion}'
[Pipeline] sh (Get phase of build frontend-593)
[biob-cd-biobase-core-844] Running shell script
+ oc -n biob-dev get build frontend-593 -o 'jsonpath={.status.phase}'
````
and eventually the pipeline times out.

The pipeline should not check the status of the last version of the buildconfig (593 in this case), but the build that was triggered by this pipeline (592 in this case).

Occurs on current master.

clemensutschig:
Not Sure if we can Target this for 3 .. which will release end of the week.. @michaelsauter ?

michaelsauter:
Thanks for reporting!

This is definitely a problem and needs to be fixed, but I don't believe it is super crucial:

* It seems to be an edge case as I haven't seen other reports about this
* Jenkins Pipeline runs within one pipeline are sequential, so this situation does not occur when there is a 1:1 mapping between branches and environments - it can only happen if multiple branches are mapped to the same env

Issue #410 might be affected by this.

I would target this for 3.1 or even 4.0.

michaelsauter:
Some additional things to consider when attempting to "fix" this:

* How would changing this affect interaction with the orchestration pipeline?
* What should we do regarding tagging the built image with `latest`?
* What should happen with the deployment? Skip? Let it fail?

Issue #437, opened 2020-07-09
Build bails out when certain characters are in the commit message


serverhorror:
**Describe the bug**

**To Reproduce**
1. Commit and Push something with a message of
   (remove the backslashes, GitHub is not letting me format this correctly)
    ```
    this is the subject
    
    \\```
    some text here
    \\```
    ```
1. Go to the jenkins build
1. see it fail

**Expected behavior**
Build completes successfully

**Affected version (please complete the following information):**
 - OpenShift: don't know
 - OpenDevStack 2.x

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
+ oc patch bc component-dbseed --type=json --patch '[{"op": "replace", "path": "/spec/source", "value": {"type":"Binary"}},{"op": "replace", "path": "/spec/output/to/name", "value": "component-dbseed:30-a97586dd"},{"op": "replace", "path": "/spec/output/imageLabels", "value": [
        {"name":"ods.build.source.repo.url","value":"https://bitbucket.example.com/scm/project/project-component-dbseed.git"},{"name":"ods.build.source.repo.commit.sha","value":"a97586dd7709347fb31ca760e6219ad50a7e9d62"},{"name":"ods.build.source.repo.commit.msg","value":"project-143 make scripts executable in git ``` git update-index --chmod=+x .\scripts\preprocess.sh git update-index --chmod=+x .\scripts\postprocess.sh ```"},{"name":"ods.build.source.repo.commit.author","value":"Martin Marcher (36151+serverhorror@users.noreply.github.com)"},{"name":"ods.build.source.repo.commit.timestamp","value":"2020-07-09 13:03:24 +0200"},{"name":"ods.build.source.repo.branch","value":"feature/project-143-db-seed-container"},{"name":"ods.build.jenkins.job.url","value":"https://jenkins.example.com/job/project-cd/job/project-cd-component-dbseed-143/30/"},{"name":"ods.build.timestamp","value":"Thu Jul 09 11:03:41 UTC 2020"},{"name":"ods.build.lib.version","value":""}
      ]},{"op": "replace", "path": "/spec/strategy/dockerStrategy", "value": {"buildArgs": [
[...]

PR #442, opened 2020-08-25
component buildstage: fix concurrent builds - master

segfault16:
This resolves issue #425 for building docker images if there is more than one branch building on the same target environment. 
~~Log follow somehow doesn't work but I'm not sure if it does with the original implementation to be honest.~~ (Edit: Log comes in batches)


clemensutschig (changes requested):
Please fix codenarc findings..

michaelsauter:
Given we fix the build problem, what solution would you think is appropriate for deploying? I guess with your PR here, the Jenkins job would still fail later on in the job because only one deployment can succeed, right?

michaelsauter:
> Log follow somehow doesn't work but I'm not sure if it does with the original implementation to be honest.

What do you mean? Log output from the build works with the current implementation. Or does it not work for you?

segfault16:
@michaelsauter Can't say anything about the deployment since I'm only using the pipeline for building docker images, sorry. In the Jenkins console there is no "live" output of the `oc build --follow`, only if the build is finished

segfault16:
Can anyone help on those failing tests, I'm totally lost...

michaelsauter:
@segfault16 Interesting, the log output seems to be streamed for me (appears in batches but not everything at once).

I think we should think about deployment though before we merge this.

clemensutschig:
> @segfault16 Interesting, the log output seems to be streamed for me (appears in batches but not everything at once).
> 
> I think we should think about deployment though before we merge this.

I'll take a look at this today ... I am still trying to find out what we are solving here ... I could envision we capture the real tag behind latest .. and before we tag? we verify again .. (and fail if that latest has changed)?

segfault16:
@clemensutschig Thanks for looking into this again. The most obvious bug is that the build version is fetched after the build is finished. If there are running multiple builds in parallel for this component there is like a race condition which I fixed in this PR by getting the build version directly after the build has started and use this in the pipeline instance. 

This has been running smoothly in our environment now since August.

clemensutschig:
> @clemensutschig Thanks for looking into this again. The most obvious bug is that the build version is fetched after the build is finished. If there are running multiple builds in parallel for this component there is like a race condition which I fixed in this PR by getting the build version directly after the build has started and use this in the pipeline instance.
> 
> This has been running smoothly in our environment now since August.

I am just fixing the tests around this .. but - what is supposed to happen if the build config is setting latest .. then you get overwriting images there ..?!

clemensutschig:
also closes opendevstack/ods-jenkins-shared-library#437

clemensutschig:
@michaelsauter  - help me understand where you see deployment issues (in the case one runs on latest - well, there is not much we can do about it in the first place - welcome to concurrency :))

clemensutschig:
@michaelsauter  - this one is ready to merge? or?

src/org/ods/component/Context.groovy
```diff
@@ -577,7 +577,7 @@ class Context implements IContext {
 
     private String retrieveGitCommitMessage() {
         script.sh(
-            returnStdout: true, script: 'git log -1 --pretty=%B HEAD',
+            returnStdout: true, script: "git log -1 --format='%f' HEAD",
```
michaelsauter, line 580: Sorry to ask, but would you mind taking this out? It's very hard to find this change in the PR here, as it is completely unrelated. I'd rather have it separately so that it is easy to link to from the changelog, and easy to revert if need be.

michaelsauter (approved):
Apart from the Git message change which I'd love to have separately, 👍

src/org/ods/component/Context.groovy
michaelsauter: @clemensutschig I can also do that if you want me to do it ...

clemensutschig: downvote - it's linked .. and I'll backport in one shot ... ok :)?

PR #447, opened 2020-09-15
Extract issue ID also from commit message

michaelsauter:
The Jira issue ID can be given in either branch name or commit message.
Putting the ID into the commit message has the advantage that it is
still accessible when the commit is accessed from a different branch,
and that it is possible that every commit is connected to a Jira ticket.

~As the change made in this commit prefers the commit message over the
branch name, it is a breaking change in so far if users expect the
branch name to overrule any issue references in commit messages.~

*Update: reversed the order, which makes the change non-breaking.*

@gerardcl @henrjk @renedupont What are your thoughts regarding which piece of information should "win"? I've implemented my suggestion but happy to change if you think otherwise.

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
@@ -166,7 +166,7 @@ The `context` object contains the following properties:
 | Git commit time in RFC 3399.
 
 | issueId
-| Jira issue ID if any present in the branch name (e.g. `123` from branch `feature/FOO-123-bar-baz`).
+| Jira issue ID if any present in either commit message or branch name (e.g. `123` from commit message `FOO-123: Bar` or branch `feature/FOO-123-bar`). If the issue ID is present in both, the branch name has precedence.
 
 | openshiftHost
 | OpenShift host - value taken from `OPENSHIFT_API_URL`.
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -397,7 +397,14 @@ class Context implements IContext {
     }
 
     String getIssueId() {
-        GitService.issueIdFromBranch(config.gitBranch, config.projectId)
+        if (!config.containsKey("issueId")) {
+            config.issueId = GitService.issueIdFromBranch(
+                config.gitBranch, config.projectId
+            ) ?: GitService.issueIdFromCommit(
+                config.gitCommitMessage, config.projectId
+            )
+        }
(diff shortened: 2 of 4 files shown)


michaelsauter:
ping @henrjk @gerardcl @renedupont

renedupont:
I like it, especially combined with the commit message checker that is available as a bitbucket server plugin. Easy to enforce it for traceability.

Does your commit message matcher look only for the beginning of the message or the whole message? If you want to look in the whole message I would change the order to first check the branch name, then the commit message. I would assume it is more likely that there are several jira issue ids in one commit message than in a branch name.

gerardcl:
hi! sorry for the delay

I also like this approach to be in! definitely needed when working within highly regulated environments.

If it is really a breaking change, couldn't we make it configurable (to be able to define what is prioritized)?

In addition to this conversation, in our current project we also automated (thanks to @m-apsolon ) adding it into the git commit message via git hooks (`prepare-commit-msg` hook), so if a branch is called `feature/PROJECTX-123-foo-issue` or `PROJECTX-123-foo-issue` then then, when creating a commit message it will automatically take care to add it like: "[PROJECTX-123] the commit message"

Here the script:

```bash
#!/bin/bash
FILE=$1
MESSAGE=$(cat $FILE)
TICKET=[$(git rev-parse --abbrev-ref HEAD | grep -Eo '^(\w+/)?(\w+[-_])?[0-9]+' | grep -Eo '(\w+[-])?[0-9]+' | tr "[:lower:]" "[:upper:]")]
if [[ $TICKET == "[]" || "$MESSAGE" == "$TICKET"* ]];then
  exit 0;
fi

echo "$TICKET $MESSAGE" > $FILE
```

michaelsauter:
@renedupont @gerardcl Thanks for feedback!

I think I'll reverse the precedence now:

* That way, its not a breaking change
* As @renedupont pointed out, if the issue reference can appear somewhere in the commit message, maybe someone is using branch feature/FOO-123-foobar, but commits something like "cherry-pick: FOO-456" on that branch. would be weird if the commit has precedence in that case ... another edge case might be if users do NOT require every commit to include the issue, but they use Jira to create the branches. In that case an issue reference somewhere in the commit message should not trump the branch name.

@gerardcl @m-apsolon Thanks for sharing, this is great!

michaelsauter:
@gerardcl @renedupont @henrjk @m-apsolon Updated. Please review again. Thanks!

renedupont (approved):

gerardcl (approved):
🚀

PR #450, opened 2020-09-29
configurable emailext recipients and default to empty list

### LINKED ISSUE #449: jenkins sends failed build status notifications to undesired recipients
**Describe the bug**
When having `notifyNotGreen` enabled (default setting) Jenkins sends failed build notification emails to non expected recipients. The issue seems related to be using too much recipients options when defining recipients list in the library. See:

```
to: script.emailextrecipients([
                [$class: 'CulpritsRecipientProvider'],
                [$class: 'RequesterRecipientProvider'],
                [$class: 'UpstreamComitterRecipientProvider']
            ])
```

**To Reproduce**
1. Using ODS 3.x, make a jenkins pipeline fail
2. E-mails are sent to much more emails like: noreply@github.com <noreply@github.com>, unknown@boehringer-ingelheim.com <unknown@boehringer-ingelheim.com>

**Expected behavior**
The e-mails recipients should be less wide by default and also configurable from a user point of view.

 - OpenShift: 3.11
 - OpenDevStack 3.x, master

gerardcl:
fixes #449 

The conclusion is that in order to ensure that there is never an email sent to unexpected recipients we need to default to an empty list of recipients. Moreover, if users are configuring Jenkins SMTP (which is required) they will also need to define in the `odsComponentPipeline` context the list of the emails such:
```groovy
...
odsComponentPipeline(
  imageStreamTag: 'ods/jenkins-agent-base:3.x',
  emailextRecipients: ['email1@foo.bar', 'email2@foo.bar'],
  branchToEnvironmentMapping: [
    'master': 'dev',
    // 'release/': 'test'
  ]
) { context ->
...
```

Documentation of `emailextRecipients`context parameter has been added, but will add better documentation within another PR for fixing issue 346

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
@@ -50,7 +50,10 @@ Available options are:
 | Serviceaccount to use when running the pod.
 
 | notifyNotGreen
-| Whether to send notifications if the build is not successful.
+| Whether to send notifications if the build is not successful. Enabled by default.
+
+| emailextRecipients
+| Notify to this list of emails when `notifyNotGreen` is enabled. It is empty by default.
 
 | branchToEnvironmentMapping
 | Define which branches are deployed to which environments, see <<_git_workflow_branch_to_environment_mapping,Git Workflow / Branch to Environment Mapping>>
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -113,6 +113,9 @@ class Context implements IContext {
         if (!config.containsKey('imagePromotionSequences')) {
             config.imagePromotionSequences = ['dev->test', 'test->prod']
         }
+        if (!config.containsKey('emailextRecipients')) {
+            config.emailextRecipients = null
+        }
         if (!config.groupId) {
             config.groupId = "org.opendevstack.${config.projectId}"
         }
@@ -158,6 +161,10 @@ class Context implements IContext {
 
     void setDebug(def debug) {
         config.debug = debug
+    }
+
+    List<String> getEmailextRecipients() {
+        config.emailextRecipients
     }
 
     String getJobName() {
--- a/src/org/ods/component/IContext.groovy
+++ b/src/org/ods/component/IContext.groovy
@@ -8,6 +8,10 @@ interface IContext {
     // Get debug mode
     boolean getDebug()
 
+    // Jenkins email extension recipients list
+    // docs: https://www.jenkins.io/doc/pipeline/steps/email-ext/
+    List<String> getEmailextRecipients()
+
     // Get the location of the xmlunit results
     String getTestResults()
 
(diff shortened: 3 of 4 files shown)


metmajer (commented):
Thank you @gerardcl for taking this further! Did you test the pipeline with an empty list of recipients? In this case, the `to:` field in `script.emailext` would be an empty String, which may not be valid. Looks good otherwise.

gerardcl:
hi @metmajer ! yes I did :)
Basically an empty list ends up being an empty string which is then handled by the emailext plugin and it does not send anything. But it shows a message saying that it is discarding sending emails due to empty list, see:

```
[Pipeline] { (odsPipeline error)
[Pipeline] echo
WARN: [overleaf] ***** Finished ODS Pipeline for overleaf (with error) ***** (took 46259 ms)
[Pipeline] echo
WARN: Error: hudson.AbortException: trigger error
[Pipeline] withCredentials
Masking supported pattern matches of $USERNAME or $TOKEN
[Pipeline] {
[Pipeline] sh (Set bitbucket build status via API)

+ curl --fail --silent --request POST --header 'Authorization: Bearer ****' --header 'Content-Type: application/json' --data '{"state":"FAILED","key":"******","name":"b1a93af8","url":"https://jenkins-****-cd.*****openshiftapps.com/job/****-cd/job/*****-cd-overleaf-master/29/"}' https://bitbucket****com/rest/build-status/1.0/commits/b1a93af8c789d2877e405b3c842c32edd38666a5
[Pipeline] }
[Pipeline] // withCredentials
[Pipeline] emailext
An attempt to send an e-mail to empty list of recipients, ignored.
[Pipeline] }
[Pipeline] // stage
[Pipeline] echo
Attempting to stash testresults from location: 'build/test-results/test'
[Pipeline] sh (Moving test results to system location: build/test-results/test)
+ mkdir -p build/test-results/test build/test-results/test
+ cp -rf 'build/test-results/test/*' build/test-results/test
+ true
cp: cannot stat 'build/test-results/test/*': No such file or directory
[Pipeline] sh (Counting test results in build/test-results/test)

+ ls -la 'build/test-results/test/*.xml'
+ wc -l
ls: cannot access build/test-results/test/*.xml: No such file or directory
[Pipeline] echo
No xUnit results found!!
[Pipeline] }
[Pipeline] // node
[Pipeline] }
[Pipeline] // podTemplate
[Pipeline] End of Pipeline
ERROR: trigger error
Finished: FAILURE
```

Let me know your thoughts!

michaelsauter:
@gerardcl Cool!

What I'm still missing is a hint how I would specify e.g. `CulpritsRecipientProvider` from emailext? How would I do this? Can we make at least an example? Also, while I believe it will not affect many people, this is a breaking change in the `3.x` line which we should avoid in general. I'm in favour to do the exception here, but we should point it out somewhere!

gerardcl:
hi @michaelsauter , yes! totally makes sense

My intention was to provide better docs on this topic via this PR https://github.com/opendevstack/ods-core/pull/828 

Within the change provided one cannot add emailext provider classes, only specific emails. So, yes, it is a breaking change but won't break the CICD pipelines. Users will be able to have better control now and define exactly who/what has to receive failing job notifications.

PR #453, opened 2020-10-06
Fix reference of docker registry in custom agents

serverhorror:


gerardcl (approved):
LGTM!

michaelsauter (approved):
👍 

One question: Do you also define the `dockerRegistry` variable before the `node` block? I think I always had it like this, wondering if needed or if the version you have works as is.

PR #456, opened 2020-10-09
Document skipping builds

michaelsauter:


renedupont (approved):
the following alternatives are possible: 
'[ciskip]', '[ci-skip]' and '[ci_skip]'
not sure if we should add this to the doc or leave it out on purpose.

michaelsauter:
Left it out on purpose to make this as simple as possible. Good to catch those variants too though in case someone gets it wrong from memory!

PR #457, opened 2020-10-09
Set shared lib ref to value of ODS_IMAGE_TAG by default

michaelsauter:
Addresses https://github.com/opendevstack/ods-core/issues/773.

It tries to be as non-invasive as possible. Going forward, quickstarters
should use `@agent_image_tag@` and `@shared_library_ref@` in their
`Jenkinsfile.template` files. Those values are filled with the value of
`ODS_IMAGE_TAG` by default, which represents the new "ODS version as
consumed by customers".

Follow-up from https://github.com/opendevstack/ods-core/pull/831.


gerardcl (approved):
👍

PR #462, opened 2020-10-16
create PR and receive default reviewers

renedupont:
Adds functionality to create PRs and receive default reviewers to bitbucket service.

Background info:
We recently had the case that we needed a jenkins stage that syncs a file from one source repo to another target repo whenever that file changes. Furthermore, we wanted to create a PR towards the target repo instead of just pushing to it. To fully automate it, we also want to determine the reviewers of that PR via configuration in the Jenkinsfile or just take the default reviewers when not configured.

In this Github PR here we do not add that sync stage, we just add the 'create PR' and 'get default reviewer' methods that we needed for that sync stage to the bitbucket service because they could be of general interest next to the already existing getPullRequest etc. methods.


gerardcl (dismissed):
🚀 LGTM!

michaelsauter:
Awesome! Can you add a small section about this to https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/component-pipeline.html#_interacting_with_bitbucket?

renedupont:
done @michaelsauter  : https://github.com/opendevstack/ods-jenkins-shared-library/pull/462/commits/22e3faaae320f63505b0c63d040f83511d2389d3

renedupont:
@michaelsauter ok I took out the number to prevent this from happening again :)

michaelsauter (approved):

src/org/ods/services/BitbucketService.groovy
```diff
@@ -153,7 +240,7 @@ class BitbucketService {
                 label: "Post comment to PR#${pullRequestId}",
                 script: """curl \\
                   --fail \\
-                  --silent \\
+                  -sS \\
```
metmajer, line 243: @michaelsauter looking at some of the PRs on master which are not in 3.x to understand what we'll merge into soon. Wasn't -sS source of some issues recently?

michaelsauter: No this is an improvement - the issues were that `-S` was missing :)

`-s` is the same as `--silent`, and `-S/--show-error` enables error output, see `man curl`:

> When used with -s, --silent, it makes curl show an error message if it fails

Issue #458, opened 2020-10-13
Clarify Repository Types


serverhorror:
In `docs/modules/jenkins-shared-library/pages/orchestration-pipeline.adoc` it is not clear that `type: ods` *must* have JUnit XML results.

PR #459, opened 2020-10-13
Clarify requirement for XML results

serverhorror:
#fix #458

PR #464, opened 2020-10-19
Clarify repository types

serverhorror:
Introduces edits in the master branch of ODS

### DIFF
--- a/docs/modules/jenkins-shared-library/pages/orchestration-pipeline.adoc
+++ b/docs/modules/jenkins-shared-library/pages/orchestration-pipeline.adoc
@@ -67,6 +67,8 @@ repositories:
 
 This type designates ODS components designed for _code development_. Such repositories are based on quickstarters whose names start with `be-`, `ds-`, or `fe-`, for _backend_, _data science_, and _frontend_, respectively. This is the default type.
 
+If you use this type ODS expects to find JUnit XML test results. If you do not have any test results the pipeline will fail. If you are deploying something where JUnit XML test results are not available consider using <<Repository Type: ods-service>>.
+
 ==== Repository Type: ods-service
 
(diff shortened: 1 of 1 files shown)


serverhorror:
In reference to #459

michaelsauter:
@serverhorror Could you update this PR to only include the changes introduced by 719bee2? Otherwise I think this will break stuff ... (the alias thingy in the docs)

metmajer:
@serverhorror agree with Michael that this PR contains too many changes. May I ask you about the use-case? What was the need to switch to the ods-service type? Did this component contain code?

serverhorror:
@michaelsauter  When you rely on to magic from the IDE... sorry about that. This should have been only the cherry-pick from #459. I think that's better now.

fix #458

serverhorror:
@metmajer There's nothing that could be unit tested here. No code just pulling in an image and adapting it so that the clients requirements are met (e.g. one is unable to roll-out images directly from the upsream repository).

metmajer:
@serverhorror we should consider the necessity that code components have to provide Unit tests (at least in our enterprise context). This is why the pipeline fails. A general recommendation to use the `ods-service` type is misleading. Let's keep this PR open and add an improved description.

serverhorror:
@metmajer Your call. Just be aware that it already is in ODS3. Best to close this PR, revert #459 and reopen #458 then.

Just keep in mind that, while a noble goal, not everything can (or should be *unit* tested). Yes, there are regulatory requirements and they do make sense. It is, however, not the case for all parts of all organizations and this simply documents the current state.

michaelsauter (approved):

michaelsauter:
I would be fine with the addition.

metmajer:
Fine, too. Let's not overcomplicate things. Compliance related information should not be handled here.

PR #468, opened 2020-10-20
 deploy: add configuration for pod findretry - 3.x

### LINKED ISSUE #465: allow parameters for timeout and retries for getPodDataForDeployment(String rc)
There are cases when the transition from an image to a container might take longer than the hardcoded 5 times 12 seconds retries.
Therefore it would make sense to keep the default 5 times 12 seconds (=60s) delay initially however it should be configurable within a pipeline definition.

https://github.com/opendevstack/ods-jenkins-shared-library/blob/0d49d03b77042a39ce364a7ce5150af5504972a4/src/org/ods/services/OpenShiftService.groovy#L433-L446

```
.....
[Pipeline] echo
2020-10-01 02:49:35Z Could not find 'running' pod for deployment=xxxxxxxxxxxxxxxxxx - waiting
[Pipeline] sleep
Sleeping for 12 sec
[Pipeline] }
[Pipeline] // stage
.........
.........
[Pipeline] End of Pipeline
java.lang.RuntimeException: Could not find 'running' pod for deployment=myproject-32
	at org.ods.services.OpenShiftService.getPodDataForDeployment(OpenShiftService.groovy:445)
	at org.ods.component.RolloutOpenShiftDeploymentStage.run(RolloutOpenShiftDeploymentStage.groovy:117)
	at org.ods.component.Stage.execute(Stage.groovy:23)
```

clemensutschig:
closes #465  for 3.x

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
@@ -87,6 +87,9 @@ Available options are:
 
 | openshiftRolloutTimeout
 | Deprecated in 3.x! Please use option `deployTimeoutMinutes` on `odsComponentStageRolloutOpenShiftDeployment`.
+
+| testResults
+| Configurable location for xunit test results, in case the build does not put them into `build/test-results/test`.
 |===
 
 == Pipeline Context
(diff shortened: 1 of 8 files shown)


michaelsauter (approved):

felipecruz91:
Thank you very much for your prompt action, @clemensutschig. I've a minor comment:

Shouldn't the [MROPipelineUtil.groovy](https://github.com/opendevstack/ods-jenkins-shared-library/blob/227cc112cca2bfdf6b5292ba38964d3bf7d5cb0d/src/org/ods/orchestration/util/MROPipelineUtil.groovy#L486) have the flexibility to benefit from the `retries` param as well?

Currently it ends up invoking the `getPodDataForDeployments` function that has a hard-coded value of `5` for the [number of retries](https://github.com/opendevstack/ods-jenkins-shared-library/blob/b1372c8e55d2c1f642103a937c20a3d6876e2f9a/src/org/ods/services/OpenShiftService.groovy#L577).

michaelsauter:
@felipecruz91 I think in this spot this isn't really needed. The hardcoded `5` is used where there should be an existing deployment, so no need to wait for anything. In theory, `1` should be enough there.

felipecruz91:
@michaelsauter Ah, I see it now, thanks for the clarification!

PR #471, opened 2020-10-21
Rename release branch to release/${version}

### LINKED ISSUE #470: Release Manager - Branch naming convention has changed in ODS 3.x
Hi,

I've noticed a slight difference in the branch naming convention when deploying a component to DEV using Release Manager between ODS 2.x and 3.x:

### ODS 2.x
`release/${version}`, i.e: `release/1.0`

image

### ODS 3.x
`release/ods-generated-${version}`, i.e: `release/ods-generated-${version}`

image

However, I haven't found where in the official documentation this information is. This is the documentation I have checked:

- https://www.opendevstack.org/ods-documentation/opendevstack/3.x/quickstarters/release-manager.html
- https://www.opendevstack.org/ods-documentation/opendevstack/3.x/update-guides/3x.html

I'd really appreciate if someone could tell me where I can find that information.

**Additional context**
This is the only issue I found with occurrences to "ods-generated": opendevstack/ods-core#484

Update: This [section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/orchestration-pipeline.html#_source_code_organisation) of the documentation needs to be updated for 3.x from:

> When a version is given, source code will be taken from a branch release/$VERSION in each repository.

to:

> When a version is given, source code will be taken from a branch release/ods-generated-$VERSION in each repository.

/ cc @metmajer

felipecruz91:
Closes #470


src/org/ods/services/GitService.groovy
```diff
@@ -74,13 +74,16 @@ class GitService {
         } else if (branch.startsWith('release/')) {
             def list = branch.drop('release/'.length()).tokenize('-')
             "${list[0]}-${list[1]}"
+        } else if (branch.startsWith('release-ods-generated/')) {
```
metmajer, line 77: @michaelsauter not sure about this code portion. Seems like the block in L74 is a left-over before `ODS_GIT_TAG_BRANCH_PREFIX` was introduced? Think we'll only ever need one block that refers to a constant.

metmajer (changes requested):
@felipecruz91 please have a second look at the original issue. @michaelsauter and I have shared additional thoughts.

src/org/ods/services/GitService.groovy
michaelsauter: AFAIk this is only used by components, to extract a Jira issue ID (from Atlassian-suggested branch patterns). The code path is not used from the orchestration pipeline, so we don't need any changes here.

metmajer:
@felipecruz91 any progress?

michaelsauter:
I have some thoughts around the Git branches at the moment.

We decided to create a release branch per version so that:

1. work can continue on `master` while a release is being prepared for installation
2. old releases can receive bugfixes without polluting them with changes from `master`

However, it seems that having separate branches per release is significant overhead for many, causes confusion (e.g. where to make changes so that it takes effect), and requires more work in the orchestration pipeline (merging back from release to `master`).

Therefore, I've been thinking if we couldn't make release branches optional. Meaning: by default, the orchestration pipeline would grab the source code from `master` (or whatever is the configured branch in the release manager repo). However, users could create release branches by themselves, and if such a branch (e.g. `release/VERSION`) existed, the orchestration pipeline would grab the code from that branch instead. This would allow users to still achieve (1) and (2) above if they do have the specific need for it - but it's an opt-in, not something that users need to learn upfront.

If we did this change, then we would not need this PR, but could just drop those codepaths. @felipecruz91 @metmajer Thoughts on this?

felipecruz91:
@michaelsauter I think it makes sense but, we need to improve how this gets executed. From my perspective, it could be a bit confusing to have both: a file in the Release Manager repository where users are expected to define _what_ branch the orchestration pipeline will use to grab the source code from (e.g, `master`) and _also_ a and non-explicit/declarative behaviour (e.g. if a branch `release/VERSION` exists, the orchestration pipeline will use this branch) that is not stated anywhere in the Release Manager source code.

Not sure how but, if possible, I'd like to have both approaches in a declarative way in the Release Manager source code to avoid having "hidden" or "defaults" behaviours from the user.

michaelsauter:
Hmm good point. I'll think about this.

src/org/ods/services/GitService.groovy
```diff
@@ -201,7 +204,7 @@ class GitService {
             extensions: extensions,
             userRemoteConfigs: userRemoteConfigs,
         ])
-    }
+        }
```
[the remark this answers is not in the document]
felipecruz91: Fixed.

src/org/ods/services/GitService.groovy
felipecruz91: Reverted this change.

src/org/ods/services/GitService.groovy
```diff
@@ -74,13 +74,16 @@ class GitService {
...
+            def list = branch.drop('release-ods-generated/'.length()).tokenize('-')
+            "${list[0]}-${list[1]}"
         } else {
             branch
         }
     }
 
     static String getReleaseBranch(String version) {
-        "release/${ODS_GIT_TAG_BRANCH_PREFIX}${version}"
+        "release-ods-generated/${version}"
```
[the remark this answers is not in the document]
felipecruz91: Using constants now.

src/org/ods/services/GitService.groovy
```diff
@@ -80,7 +81,7 @@ class GitService {
     }
 
     static String getReleaseBranch(String version) {
-        "release/${ODS_GIT_TAG_BRANCH_PREFIX}${version}"
+        "${ODS_GIT_BRANCH_PREFIX}${ODS_GIT_TAG_PREFIX}${version}"
```
metmajer, line 84: @felipecruz91 I guess this should not include the TAG prefix here

metmajer (changes requested):
Well done Felipe, there is one thing to be changed.

test/groovy/org/ods/services/GitServiceSpec.groovy
```diff
@@ -0,0 +1,22 @@
...
+        given:
+        def version = "0.0.1"
+        def script = new PipelineScript()
+        def service = new GitService(script, new Logger(script, false))
+
+        when:
+        def releaseBranch = service.getReleaseBranch(version)
+
+        then:
+        releaseBranch == "release/0.0.1"
```
felipecruz91, line 19: @metmajer @michaelsauter Please double check this is the expected value.

metmajer: 💯

metmajer (approved):
Great first contribution, Felipe!

metmajer:
@felipecruz91 please squash and merge your code into `master`. Once done, please do a `git cherry-pick $sha` inside the `3.x` branch and push.

PR #476, opened 2020-10-23
Component pipeline: better logging, less confusion with xunit results

### LINKED ISSUE #475: Improve output when no unit tests are found
Currently we output something like this:

```
ls: cannot access build/test-results/test/*.xml: No such file or directory
[Pipeline] echo
No xUnit results found!!
```

This looks very much like an error, but in fact is not. Especially in cases where the pipeline has an encountered an error before, the junit result collection is often it is mistaken for the actual error.

Maybe there is a way to make the output less confusing?

clemensutschig:
closes #475 for master

### DIFF
--- a/src/org/ods/services/JenkinsService.groovy
+++ b/src/org/ods/services/JenkinsService.groovy
@@ -22,17 +22,22 @@ class JenkinsService {
             xUnitResultDir = customXunitResultsDir.trim()
         }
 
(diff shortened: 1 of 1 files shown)


clemensutschig:
the diff is super small (just 5 lines) - but for some reason dos2unix changed now all linefeeds I guess :-8

michaelsauter (approved):

PR #490, opened 2020-11-18
Fix variable references

michaelsauter:
Noticed these during a refactor on master where I added `TypeChecked` annotations (still to be proposed as a PR).

I guess the code paths in question have never been executed so far on 3.0.


renedupont (approved):

clemensutschig (approved):
oops

felipecruz91 (approved):
Probably a test case executing this particular code path would be very beneficial 😁

PR #492, opened 2020-11-18
Pass project to every method of OpenShiftService

michaelsauter:
* Experience shows that we do interact with more than one project
* Allows to separate build environment and deploy environment
* Paves the way to deploy to multiple projects in the same pipeline

To verify I added `@TypeChecked` annotations, ran unit tests and a quickstarter test. Anything else I should test? I'm going to cover more areas with tests when I change functionality, so if I overlooked something then it'll show up there as well.

@clemensutschig As you had it originally :) I wanted to have `project` as an instance variable so that it easy to use, but for where we're going, we need greater flexibility. So now `project` needs to passed to pretty much every method ...

FYI @fkressmann That will make your use case even easier ... and there is more to come in that area ... I'll ping you with more details.


michaelsauter:
@renedupont @clemensutschig ping

I'd like to get this in to be able to base my further work on master.

clemensutschig (approved):

Issue #496, opened 2020-11-23
[component pipeline] Allow to deploy to multiple clusters


michaelsauter:
**Is your feature request related to a problem?**
It would be nice to deploy to multiple clusters (and within each cluster to multiple namespaces). This is a typical scenario, and currently users have to build it themselves.

**Describe the solution you'd like**
`odsComponentStageRolloutOpenShiftDeployment` should be able to deploy to multiple clusters/namespaces. For this, it first needs to change from rolling out one deployment (identified by `resourceName`) to rolling out all deployments for the selector. Then, it needs to be able to read "environment configuration", that specifies for each environment the clusters / namespaces to deploy to. The configuration could/should be largely identical to https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/orchestration-pipeline.html#_environments, but it should be able to specify multiple clusters. Further, the configuration cannot be sourced from a repo/file, but must either be inline in the `Jenkinsfile` (which works well for one component) or it could be sourced from a `ConfigMap` in the `*-cd` project (which would be the better fit for multiple components with same/similar configuration).

**Describe alternatives you've considered**
The orchestration pipeline is also missing this functionality. One could add the functionality there, and require users wanting to deploy to multiple clusters to use the orchestration pipeline. I would like to avoid this because:

* the orchestration pipeline is complex and imposes a certain process that is not required for "just" deploying to multiple clusters
* the orchestration pipeline bundles all components together, and there are good organisational and technical reasons to deploy individually (e.g. less risk, higher speed)

**Additional context**
We also need to take care of importing images to the target clusters.

metmajer:
Related to https://github.com/opendevstack/ods-jenkins-shared-library/issues/389

metmajer:
As discussed with @michaelsauter, the following topics need to be addressed:

- **D/P + P vs. D + Q/P**: Will deployments into different P-environments differ and, if yes, in which ways? Do we need a Q environment next to each P, or is the setup (D/Q + P clusters) sufficient? Think of requirements depending on the geographic location (language, location-specific features, etc.)
- **Atomicity**: Do we require rollbacks if the application fails to deploy on a single cluster? Can we avoid rollbacks by implementing idempotence?
- **TIP/TIR documents**: Assumption: TIR for P to contain the P-cluster locations, but there's still only a single document generated for P.
[...]

Issue #390, opened 2020-06-26
Orchestration Pipeline: Support multiple replicas


michaelsauter:
The orchestration pipeline only deals with one replica of a pod. This somehow defeats the purpose of OpenShift, allowing to scale horizontally easily. Even when users do not need to scale massively right now, I would see it as a common best practice in the industry to have "two of everything" to avoid locking yourself into a situation where it only works with one instance.

Affected areas of the shared lib:

* documents
* checking for running pods after deployment and collecting information from them
* ?

FYI @clemensutschig @metmajer I'm certain we will get requests for that soonish :)

metmajer:
Thanks @michaelsauter, you are a mind reader. 2 instances is also the minimum to allow for 0-downtime upgrades of application components. +5 from me :-)

clemensutschig:
Michael you could try. I guess it will work.. docs will just show one pod i guess..

michaelsauter:
@metmajer I think the upgrade downtime is not directly related. In OpenShift, you can have rolling updates, which start a 2nd pod before terminating the old one. If you are careful with readiness/liveness probes, you can have zero downtime without running 2 pods continuously.

I'm not going to start working on it now. I have to finish the other stuff (merging back into main branch, checking before release that all previously released commits are part of the new release, and a few loose ends) first. I think I won't have the capacity to work on this for v3 ... let's see how the other stuff progresses.

martsec:
The TIR docs should be adapted, since with the introduction of Cloud and other non openshift components, IMO the best way would be to provide a section where the different usecases inject whatever html data they want/need to show. That way, adding more information about all the pods created would only mean a change in the shared library instead on the template. A good candidate would be section 2.2 of TIR. Right now, the table is hardcoded in the template, but for re-usability reasons, it should not be. 
![image](https://user-images.githubusercontent.com/6894585/87677563-33f73d00-c77a-11ea-989b-380fd3d4dd44.png)

PR #500, opened 2020-11-23
Support Deployment resources in component pipeline

### LINKED ISSUE #295: [component pipeline] Support Deployment resources
Right now, the library only supports working with `DeploymentConfig` resources. However, one can also use `Deployment` resources in OpenShift, and it would be good to figure out if / how we want to support that.

michaelsauter:
Closes #295.

Also deals with multiple pods (replicas) in the OpenShiftService, laying
some groundwork for #390.

Note that due to `rollout restart` being introduced in Kubernetes 1.15,
Deployment resources are not supported in OpenShift 3.11 at the moment.
This could change in future PRs if we change how we trigger rollouts
(specifically, by not using the `latest` tag but instead update the pod
template to different image tags for every rollout). Further, use of
`Deployment` resources is not (yet) possible when the orchestration
pipeline is being used.

Also note that the introduction of the config option `deploymentKind` in
`odsComponentStageRolloutOpenShiftDeployment` will likely be temporary.
#496 will require to refactor the stage so that it can deal with the whole
component, not individual Deployment or DeploymentConfig resources.

### DIFF
--- a/test/groovy/vars/OdsComponentStageRolloutOpenShiftDeploymentSpec.groovy
+++ b/test/groovy/vars/OdsComponentStageRolloutOpenShiftDeploymentSpec.groovy
@@ -28,17 +28,18 @@ class OdsComponentStageRolloutOpenShiftDeploymentSpec extends PipelineSpockTestB
       componentId: 'bar'
   ]
 
-  def "run successfully without Tailor"() {
+  def "run successfully without Tailor [DeploymentConfig]"() {
     given:
     def c = config + [environment: 'dev', targetProject: 'foo-dev', openshiftRolloutTimeoutRetries: 6]
     IContext context = new Context(null, c, logger)
     OpenShiftService openShiftService = Stub(OpenShiftService.class)
-    openShiftService.resourceExists(*_) >> true
-    openShiftService.getLatestVersion(*_) >> 123
-    openShiftService.rollout(*_) >> "${config.componentId}-124"
+    openShiftService.resourceExists('foo-dev', 'DeploymentConfig', 'bar') >> true
+    openShiftService.resourceExists('foo-dev', 'Deployment', 'bar') >> false
+    openShiftService.getRevision('foo-dev', 'DeploymentConfig', 'bar') >> 123
+    openShiftService.rollout('foo-dev', 'DeploymentConfig', 'bar', 123, 5) >> "bar-124"
     // test the handover of the poddata retries
-    openShiftService.getPodDataForDeployment(_, _, 6) >> [ deploymentId: "${config.componentId}-124" ]
-    openShiftService.getImagesOfDeploymentConfig (*_) >> [[ repository: 'foo', name: 'bar' ]]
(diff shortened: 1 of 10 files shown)


michaelsauter:
Opening this up for review now to get some comments early if you have. I have tested components with both DeploymentConfig and Deployment resources. Right now I'm looking into how I can extend the tests to show both variants work - unfortunately it is hard to "prove" without a real cluster to work against.

test/groovy/vars/OdsComponentStageRolloutOpenShiftDeploymentSpec.groovy
```diff
@@ -28,17 +28,18 @@ class OdsComponentStageRolloutOpenShiftDeploymentSpec extends PipelineSpockTestB
...
-    openShiftService.getLatestVersion(*_) >> 123
-    openShiftService.rollout(*_) >> "${config.componentId}-124"
+    openShiftService.resourceExists('foo-dev', 'DeploymentConfig', 'bar') >> true
+    openShiftService.resourceExists('foo-dev', 'Deployment', 'bar') >> false
+    openShiftService.getRevision('foo-dev', 'DeploymentConfig', 'bar') >> 123
+    openShiftService.rollout('foo-dev', 'DeploymentConfig', 'bar', 123, 5) >> "bar-124"
     // test the handover of the poddata retries
-    openShiftService.getPodDataForDeployment(_, _, 6) >> [ deploymentId: "${config.componentId}-124" ]
-    openShiftService.getImagesOfDeploymentConfig (*_) >> [[ repository: 'foo', name: 'bar' ]]
+    openShiftService.getPodDataForDeployment('foo-dev', 'DeploymentConfig', 'bar-124', 6) >> [[ deploymentId: "bar-124" ]]
```
felipecruz91, line 41: Out of curiosity - why has it changed to a double array?

felipecruz91 (approved):

test/groovy/vars/OdsComponentStageRolloutOpenShiftDeploymentSpec.groovy
michaelsauter: Note that this is actually not a double array, but `List<Map>`, whereas previously it was just `Map`. 

The reason this is a list now is because one deployment can have multiple pods ("replicas"). I would consider this a pretty typical use case, which we should support eventually in the docs. The issue tracking this is #390. In this PR, I'm already dealing with multiple pods properly from a technical perspective. Previously, we would only look at the first pod, not all of them, which is not fully correct. Now we consider all pods when gathering the data, but the orchestration pipeline then only uses the first item in the list to put it into the docs (and throws away information about any further pods).

felipecruz91: Interesting! Thanks for the explanation Michael

clemensutschig (approved):
Like it!

PR #502, opened 2020-11-23
Feature/add branch to sonar cnes

### LINKED ISSUE #491: Sonarqube report - code analysis always based on application's master branch
Currently in ODS 3.x and ODS 2.x, If we use Docgen for generate documentation for a quickstarter using a branch different than master in the quickstarter, a document generated (SSDS document) will contain sonarqube analysis info from quickstarter's master branch instead of branch used by quickstarter.

For example, if we use docgen for generate documentation for a python quickstarter concretely from quickstarter develop branch, the resulting SSDS will contain in "Code Analysis" section sonarqube reports made from quickstarter master branch instead of develop branch (which was the choosen branch in Docgen).

We also noticed that, in this case docgen, generates successfully the sonarqube scan succesfully for develop branch.

**For explaining this case, we'll provide the following example:** 
* Docgen with only one application
* python application with two branches:
   * master (contains 7 lines of code in main application)
   * develop (contains 9 lines of code in main application)

**Setup used for reproduce this case:**

**Release manager metadata.yml file:**
metadata

**Sonar .properties file (generates in execution time the branch used, project name, etc)**
sonar-config

**Custom stage for configure sonar config in execution time:**
sonar-prepare

**We also make sure our sonar config is generated in the proper time:**
stage_declaration_order

**We also checked OpenDevStack docs for this task:**
https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/orchestration-pipeline.html#_automated_generation_of_compliance_documents
and applyed the following param to our jenkinsfile:

sonarqube-branch

**After running Docgen, we see that our sonarqube scan was performed properly in application develop branch:**
sonarqube-ui-main

**We can check the develop branch scan report:**
sonar-develop-coverage

**If we check the resulting SSDS Code Analysis section, we see the resulting report is coming from latest master branch codescan:**

SSDS

**master branch code scan:**
sonar-master-coverage

How we can solve this?

Best Regards

renedupont:
Fixes #491 

with this PR the following will happen:
- every git branch will have its own sonar branch when running through the pipeline (if sonarqube edition is higher than community and the respective env var is set)
- using `sonar.branch.name` in the sonar-projects.properties has no effect since it is automatically set in the pipeline to the git branch name (same behaviour as for `sonar.projectKey` and `sonar.projectName`)
- cnesreporter will use the `sonar.branch.name` value as input for its `-b` branch option (if sonarqube edition is higher than community and the respective env var is set) 

also see https://github.com/opendevstack/ods-core/pull/890 which goes along with this PR

### DIFF
--- a/src/org/ods/services/SonarQubeService.groovy
+++ b/src/org/ods/services/SonarQubeService.groovy
@@ -14,7 +14,7 @@ class SonarQubeService {
         script.readProperties(file: filename)
     }
 
-    def scan(Map properties, String gitCommit, Map pullRequestInfo = [:], boolean debug = false) {
+    def scan(Map properties, String gitCommit, Map pullRequestInfo = [:], String sonarQubeEdition, boolean debug = false) {
         withSonarServerConfig { hostUrl, authToken ->
             def scannerParams = [
                 "-Dsonar.host.url=${hostUrl}",
@@ -22,8 +22,10 @@ class SonarQubeService {
                 '-Dsonar.scm.provider=git',
                 "-Dsonar.projectKey=${properties['sonar.projectKey']}",
                 "-Dsonar.projectName=${properties['sonar.projectName']}",
-                "-Dsonar.branch.name=${properties['sonar.branch.name']}",
             ]
+            if(sonarQubeEdition != 'community') {
+                scannerParams << "-Dsonar.branch.name=${properties['sonar.branch.name']}"
+            }
             if (!properties.containsKey('sonar.projectVersion')) {
                 scannerParams << "-Dsonar.projectVersion=${gitCommit.take(8)}"
             }
--- a/src/org/ods/services/SonarQubeService.groovy
+++ b/src/org/ods/services/SonarQubeService.groovy
@@ -54,7 +54,7 @@ class SonarQubeService {
 
     def generateCNESReport(String projectKey, String author, String sonarBranch, String sonarQubeEdition) {
         withSonarServerConfig { hostUrl, authToken ->
-            def branchParam = sonarQubeEdition == 'community' ? '' : "-b $sonarBranch"
+            def branchParam = sonarQubeEdition == 'community' ? '' : "-b ${sonarBranch}"
             script.sh(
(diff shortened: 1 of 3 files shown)


clemensutschig (approved):

jssnull (approved):

src/org/ods/services/SonarQubeService.groovy
```diff
@@ -48,16 +52,18 @@ class SonarQubeService {
         }
     }
 
-    def generateCNESReport(String projectKey, String author) {
+    def generateCNESReport(String projectKey, String author, String sonarBranch, String sonarQubeEdition) {
         withSonarServerConfig { hostUrl, authToken ->
+            def branchParam = sonarQubeEdition == 'community' ? '' : "-b $sonarBranch"
```
michaelsauter, line 57: Doesn't this need to be `"-b ${sonarBranch}"`? At least for consistency we should do that ...

michaelsauter (changes requested):

felipecruz91:
@renedupont In what case the `sonar.branch.name` variable in the sonar-projects.properties is used? Or is it not used anymore as the pipeline sets it automatically to the git branch name?

src/org/ods/services/SonarQubeService.groovy
renedupont: For a single param this is enough but I changed it nevertheless now for consistency. Btw. I saw other cases where we also do it that way, e.g. https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Pipeline.groovy#L308

renedupont:
@felipecruz91 In the initial description of the issue #491 it was used to create a 'branch' on sonarqube server. With this PR this is not necessary anymore as it will be automatically set by the pipeline, hence setting it in sonar-projects.properties will have no effect anymore.

michaelsauter (approved):

clemensutschig:
@renedupont  - this needs a backport to 3.x because it's a bug from 3.0

Issue #410, opened 2020-07-10
[component pipeline] Build-only environment


michaelsauter:
**Is your feature request related to a problem?**
The option `branchToEnvironmentMapping` maps branches to environments. If the current branch does not match any environment, no target project in OpenShift is selected, and the stage to build an OpenShift image as well as the rollout stage are skipped. Unfortunately, this means there is no feedback on whether the container image can be built on those branches. And the only way to get that feedback, is to also rollout the image. Before #400, building images for every branch lead to a lot of waste as many images would have been built, but not reused. Now that we can reuse images, it would be nice to specify that images should be built but not deployed.

**Describe the solution you'd like**
We could add a prefix to the environment, like this:
```
branchToEnvironmentMapping: [
  "master": "dev",
  "*": "buildonly:dev"
]
```

Which would tell the pipeline to build & deploy to `dev` on `master` branch, and just build in `dev` for all other branches.

**Describe alternatives you've considered**
* Don't make any change and require that users create a separate env so that they can build & deploy there - that would give the benefit that they also run the image, which is even more proof that "it works". But likely the deployments from different branches would fight over this env, which can lead to weird errors. Also, it would require and extra env - the solution above avoids this
* Some other way of specifying that an env is "build only" ... but so far the proposed one feels best to me

**Additional context**
* This should not interfere with the orchestration pipeline as there the environment is anyway set by the release manager - and would always be build & deploy
* We would need to extend the component context with a flag that tells the pipeline if only build should happen. Should be a relatively small change ...

FYI @gerardcl @oalyman @renedupont @henrjk @clemensutschig @metmajer Thoughts?

henrjk:
Cool would be if one could connect to a private openshift (for example running locally on a developer machine) for local builds and one would still observe build failures etc. However this would be a much bigger effort.

Meanwhile this looks good to me. Some questions I wonder about: 

Will the image build during buildonly go to the registry? 
If it is could this cause inadvertent deployments based on latest tag?

michaelsauter:
@henrjk Interesting idea with the local builds. As you say this would likely be bigger and therefore a separate discussion.

> Will the image build during buildonly go to the registry?

I would say yes. Otherwise we would need to patch the `BuildConfig` to avoid any output image - and I'm not sure that is even possible.

> If it is could this cause inadvertent deployments based on latest tag

[...]

PR #513, opened 2020-11-30
Remove automatic versioned dev env creation

michaelsauter:
The option to use a versioned env still exists, but the creation has to
be manual now. This cicrumvents issues with `self-provisioner`
permissions and missing `admin` permissions of the `jenkins`
serviceaccount.

Going forward, this allows us to build images in the `*-cd` project,
which is easier to manage, and frees us from requiring the `*-dev` env
to be on the same cluster. This can be handy if e.g. you want to have
build infrastructure on a non-internet facing cluster, but deploy into
an internet-facing cluster.

This change is related to https://github.com/opendevstack/ods-jenkins-shared-library/pull/512.


metmajer:
@michaelsauter why do you see a need to remove this? Actually, with EDP 1.1 supporting application version management, versioned developments (and actually also testing environments) would be necessary to allow the parallel development of versions. E.g. a product increment in parallel to a fix version of what is currently in production. Both could be developed and tested without interfering with each other. Was this implementation causing any issues?

michaelsauter:
@metmajer Sorry, the reasons are a bit hidden in the linked issues, so I'm putting them here as well:

* It doesn't work anyway in a typical, regulated environment where the `jenkins` serviceaccount does not have `self-provisioner` rights. Meaning no current EDP project is actually using it.
* Even if `jenkins` did have enough rights, it's not `admin` on the `*-cd` project, which means it cannot add roles. Therefore, it cannot give the newly created namespace `image-puller` rights to the  `*-cd` project. However, this is required for #517. 

And I believe #517 to be crucial for several reasons:

* It simplifies the authoring of templates/charts, because BuildConfig/ImageStream do not need to be specified
* It allows allows to separate build and runtime clusters (right now you need to have DEV/TEST on the same cluster than CD)
* Images only need to be pulled from the *-cd project. If you allow multiple DEV/TEST namespaces, it quickly becomes difficult to understand / verify from where image promotion should happen
* It brings us one step closer to a future model in OpenShift 4. I expect us to drop the idea of a `BuildConfig` completely and move to building via `buildah` or `kaniko`. These would always build in `*-cd` and then just push the images to the right place.
* It allows developers to build images even for branches which they do not deploy, to verify if image building works (see #410).

Apart from those reasons, please note that the PR does NOT remove the functionality to have separate dev environments. It just removes the functionality to have them generated on-the-fly. If someone with appropriate permissions creates the separate, versioned dev env beforehand, the orchestration pipeline will use it.

@metmajer does that help?

michaelsauter:
If we do not want to remove the auto-dev-env functionality, we could require that `jenkins` has, next to `self-provisioner` permissions, also `admin` permissions in `*-cd`. That way we can unblock #517 and keep this functionality. Just noting this here as an option, but I still think removing is the better option.

henrjk:
@metmajer it appears that the parallel development would still work as one could create a fix-dev and feature-dev environments in advance so your use case of independent development of these would still be supported. Am I wrong?

metmajer:
@michaelsauter @henrjk we're in the middle of a release which is always time-consuming. Please keep this open for now so we can make a decision once there's enough time to discuss the implications and options. I would be in favor of keeping and enhancing the status quo.

michaelsauter:
@metmajer I understand that you want to take time on this. On the other hand, I'd also would like to continue with #517 to be able to take the next step towards multi-cluster deployment (which is implemented as a proposal ready for review / feedback). Do you think you could live with my suggestion above?

> If we do not want to remove the auto-dev-env functionality, we could require that jenkins has, next to self-provisioner permissions, also admin permissions in *-cd. 

If you would be ok with this so that we can move forward with #517, I would like to do this. If you are still unsure, then we wait.

metmajer:
@michaelsauter understood! @segator do you see any implications compliance-wise if the Jenkins user also has admin permissions in *-cd?

What I would love to see in the future is that every parallel development stream (e.g. 1.1 and 2.0) have their own `dev-$version` and `qa-$version` namespace which are automatically provided through the EDP. If this is not enough, users could still use the component pipelines' `branchToEnvironmentMapping` functionality to make use of additional environments. However, in the latter case, we would not provide the necessary environments (namespaces) automatically.

michaelsauter:
@metmajer @segator Note that the `self-provisioner` permissions are also not given by default and need to be granted explicitly. I am not aware that this is the case anywhere right now. Also note that the `jenkins` serviceaccount as `admin` permissions on the created project already.

I realised that there is probably yet another way to get around this. We have a `skopeo` binary in the base agent image. That can move images from one registry to another, provided it has a credentials for the source registry and credentials for the target registry. The `jenkins` serviceaccount should have both credentials (it can pull in `*-cd` and it can push into the created project because it is admin there). So technically, it should be possible to implement this tagging of images via `skopeo`, which bypasses the need to grant admin permissions in `*-cd`. It feels a little odd to use `skopeo` for sth. that could be done via `oc` if the permissions would exist, but it should work.

segator:
Thanks for the explanation, it make sense to me partially.

but I have a question, 
why we need to create NS/Project by version? 
we should be able to deploy multiple  instances of our dev(features branches) application in the same NS. maybe the problem now is that the deployment is not managed as a package, but introducing helm/kustomize or any similar tool that allow us dynamic rendering of the OC resources (Deployment,configmaps,routes...) we should be able to deploy on the same NS, then jenkins SA Self-provisioner is not needed(something that in our company we will never have) and I really would like to see in a near foture environment per feature branch, allowing e2e testing on isolated environment on Pull Requests to Master.

About the docker images I think it's good approach to build them always on the cd project and eventually push them to destination Project/NS and cluster as needed. This allow the capability to have our dev/test environments in diferents clusters so meaning will be more similar to prod behaviour.

michaelsauter:
> we should be able to deploy multiple instances of our dev(features branches) application in the same NS

Agree, we should explore this for the release manager! Creating new namespaces on the fly always comes with escalated privileges (one way or another) which we should try to avoid.

Note that multiple instances of a component are already possible right now in the component pipeline if the repo controls the OpenShift templates, but it is a bit cumbersome to setup. Helm will make this easier, because it has the explicit concept of a "release" and gives templates access to this release object.

metmajer:
@segator you are adding a new topic into the mix. Let's try to find a solution for the initial topic discussed here. Since we enable our users to develop application versions in parallel on the EDP, the automated creation of namespaces `*-dev-$version` and `-test-$version` is still very useful and should not go away. @michaelsauter had questions and suggestions regarding Jenkins user permissions. Can you please comment?

michaelsauter:
@metmajer I think @segator meant the "multiple instances within one namespace" approach as a replacement of the "multiple namespaces" approach. It should be possible to install two instances (= two versions) of an application in the same namespace.

metmajer:
@michaelsauter those approaches are not equal though. When a team uses the EDP to develop a product increment v2.0 and a fix version 1.1, it makes no sense to deploy them into the same namespace.

michaelsauter:
> it makes no sense to deploy them into the same namespace

Why is having two versions in the same namespace worse than having two versions in two namespaces?

I would not rule out @segator's suggestion immediately ... I think we should experiment with this and compare both approaches to see what is the best fit.

metmajer:
> > it makes no sense to deploy them into the same namespace
> 
> Why is having two versions in the same namespace worse than having two versions in two namespaces?

Because all changes being actively developed would land in the same target namespace. A team would need to be very careful not to override an installation in D (think developer preview and product demos) of another version. Also, any version that is currently in D/Q and on its way to P (but might stay there for a while because of manual testing) will block a bugfix that has to advance quickly into P.

segator:
> @michaelsauter understood! @segator do you see any implications compliance-wise if the Jenkins user also has admin permissions in ***-cd**?

Sorry I forgot to answer that on my initial message.
No, there are no issues with that, we can set Main Jenkins SA on every ***-cd**

> it should in the future go into the **-dev-$version** namespace

I see your point Martin, probably is safest than using multiple **$versions** in the same NS but hope this not against OC packaged apps in a sooner future, I  dream on see in our git repos our helm/kustomize definitions, as how ODS works right now there are no clear "source of truth" for OC resources, as the user can modify their resources on OC without any tracking in git.

if we use NS per $version how we will manage PR/Feature Environments  ***-dev-$commitId** ***-dev-$branchName**?

michaelsauter:
> I dream on see in our git repos our helm/kustomize definitions, as how ODS works right now there are no clear "source of truth" for OC resources

Right now you already should have the templates in the repo, and then those are your source of truth. I would highly discourage the use of the auto-export feature. In my eyes it goes completely counter to what the release manager is trying to achieve (= traceability). I would be in favour of dropping the auto-export feature to force people to define the resources. However, this is a completely separate topic, let's continue the discussion on this in https://github.com/opendevstack/ods-provisioning-app/issues/650.

> A team would need to be very careful not to override an installation in D (think developer preview and product demos) of another version

Understood. This is of course a concern, and we would need to find figure out if we can prevent this. ... another concern with multiple version instances in one namespace is that it will likely conflict as some point with quota restrictions.

> Also, any version that is currently in D/Q and on its way to P (but might stay there for a while because of manual testing) will block a bugfix that has to advance quickly into P.

I don't understand this comment. If we were to allow multiple versions within one namespace, the bugfix can be deployed to D/Q/P without being impacted / blocked by the other instances in D/Q/P.

That said, let's get back to the original issue, and leave out potential alternatives for now:

* I want to unblock #517 as that paves the way for a few really great additions to the component pipeline as noted in that issue (e.g. #535)
* #517 requires that images can be pulled from `*-cd` namespace
* As a consequence, the orchestration pipeline "auto-dev-env" feature would either need to be removed, or it would require that `image-puller` permissions on `*-cd` are granted for the new project, which requires `admin` permissions

Given what I read above, it sounds like @metmajer you are not comfortable with removing the feature. However, it seems like we can, for the projects that need it, give `self-provisioner` and `admin` permissions to the `jenkins` serviceaccount. Granting the permissions is a manual step by a cluster admin (as it is the case right now). Please note though that currently no project on our main cluster has even `self-provisioner` permissions so no one uses the feature anyway.

My proposal now is that I open a PR that gives `image-puller` permissions on `*-cd` for the newly created, versioned dev project. Then I'll close this PR and we can go ahead with #517.

@metmajer @segator Let me know if you are OK with this or not. I would love to move forward with #517 but I understand you are busy and it's ok by me if you need some time to think this through.

segator:
@metmajer  you have last word.

But here my toughts :
I think best should use multiple $versions on the same NS.

Creating NS per version will be hard to maitain(who will check/remove old envs?) Quota limitation per project is losing sense. infrastructure resources will be harder to control as quotas won't work. What about PR Environments, will we create one NS per PR or even branch?

Nevertheless if we are not ready for this transition @michaelsauter  suggestion is the way to go.
> My proposal now is that I open a PR that gives image-puller permissions on *-cd for the newly created, versioned dev project. Then I'll close this PR and we can go ahead with #517.

self-provisioner on specific project jenkins SA is not an option,.

michaelsauter:
> self-provisioner on specific project jenkins SA is not an option

but this is the whole premise on which this feature is built. That's why I originally opened the PR: because self-provisioner is not given to each jenkins SA anyway, I believe we don't need code that auto-creates an environment. The PR still allows to use an existing, versioned dev env. It just doesn't create one (because in reality, it never can anyway because cluster admins do not want this - and for good reasons I think).

metmajer:
@michaelsauter @segator I guess we reached the limits of written communication. Let's talk soon. @segator the issue you see with high resource consumption can be mitigated using an automated scale-to-0 approach.

metmajer:
Clarified the expectations with @michaelsauter. Absolutely makes sense to remove the anyway dead code. When a `version` parameter is provided != `WIP` then versioned namespaces can still be used. @michaelsauter will try to investigate if the same behavior we have for versioned `dev` namespaces can easily be ported to the `test` namespaces as well, so we can support parallel feature/fix development as long as the namespaces have been provided manually.

metmajer (approved):

PR #514, opened 2020-11-30
Deploy all Deployment(Config) resources of component together

michaelsauter:
* Avoids having to specify the `resourceName` config option to target
  multiple Deployment/DeploymentConfig resources
* Avoids having to specify the `deploymentKind` config option
* Important prerequisite for #496 as we only want to login to every
  cluster once, then deploy everything there in one go.
* This is also a better fit with Tailor/Helm, which update all
  resources with a certain label / which are part of a release, not
  individual resources.

### DIFF
--- a/src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
+++ b/src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
@@ -19,11 +19,8 @@ class RolloutOpenShiftDeploymentStage extends Stage {
         JenkinsService jenkins,
         ILogger logger) {
         super(script, context, config, logger)
-        if (!config.resourceName) {
-            config.resourceName = context.componentId
-        }
-        if (!config.deploymentKind) {
-            config.deploymentKind = OpenShiftService.DEPLOYMENTCONFIG_KIND
+        if (!config.selector) {
+            config.selector = "app=${context.projectId}-${context.componentId}"
         }
         if (!config.imageTag) {
             config.imageTag = context.shortGitCommit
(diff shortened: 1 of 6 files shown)


src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -94,28 +90,51 @@ class RolloutOpenShiftDeploymentStage extends Stage {
...
+
+                def podData = rolloutDeployment(resourceKind, resourceName, originalVersion)
+
+                if (!rolloutData.containsKey(resourceKind)) {
+                    rolloutData[resourceKind] = [:]
+                }
+                rolloutData[resourceKind][resourceName] = podData
+                // TODO: Once the orchestration pipeline can deal with multiple replicas,
+                // update this to store multiple pod artifacts.
+                // TODO: Potential conflict if resourceName is duplicated between
```
gerardcl, line 120: Should we add a warning message for this case? although it could be checked beforehand also.
Thinking more on this edge case, shall we just fail because it shouldn't happen?

gerardcl (approved):
LGTM!

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
michaelsauter: I thought about this but have not yet reached a conclusion how to deal with it / where to fail. Maybe this can also be solved by making an update to the orchestration pipeline (as in, where it expects artefacts). Therefore I'm going to leave this as is, and revisit later.

PR #516, opened 2020-12-01
Fix get the information for history

jorge-romero:
Get different information for tracking issues(versioned) and the issues to create the history, that needs the previous Documents


metmajer (approved):
Cheers @jorge-romero! Just a reminder for us that over the next months we need to create unit tests for the collective changes made in ODS 3.1 to have a safety net for the needed reduction of technical debt.

PR #517, opened 2020-12-02
Build images in *-cd project

### LINKED ISSUE #497: [component pipeline] Build images in *-cd projects
**Is your feature request related to a problem?**
We run the Jenkins Pipeline in the `*-cd` project, but trigger builds for the components in the `*-dev` and `*-test` namespaces. This has a few downsides:

* We need a `BuildConfig` in both `*-dev` and `*-test`, and need to keep them in sync
* We need `*-dev` and `*-test` projects on the cluster which hosts the `*-cd` project. If we would also build the image in the `*-cd` project, we would not need `*-dev` and `*-test` projects at all on the same cluster. This is relevant for #496. It would also open up the possibility of having a "build cluster" and a "runtime cluster". 
* We need to have resolved an environment to actually build the image at all, see #410. Moving the image building to the `*-cd` project solves #410 in a very different, easy way.

Build image in the `*-cd` project, and then move the image during deployment to the relevant target cluster/project.

Keep as it is.

**Additional context**
All quickstarters use the same `BuildConfig` and `ImageStream` definition, and we patch the `BuildConfig` in the build stage to e.g. have the right labels. Therefore, we could easily create those resources in the `*-cd` project if they don't exist yet. The component authors then don't have to care about the definition at all. Further, in the future we will probably not have `BuildConfig` resources at all, and e.g. use OpenShift pipeline tasks that use tools such as Buildah or Kaniko to build images.

michaelsauter:
Closes #497.

Last step before #496 :) In that PR, I'm going to extend the pipeline to allow deployment to external clusters, and to deploy to multiple namespaces in those. Together with this PR, this means we could have one cluster for building (just `*-cd`) and any number of "deployment clusters" (e.g. one for dev/test and one for prod). Each team starts out with cd/dev/test in one cluster, can then configure the environments/clusters as needed.


gerardcl (dismissed):
awesome!
LGTM!

comments:
- with this implementation will be easy to upgrade/migrate to if having an existing project, which is nice!
- branch to env mapping is now really telling where to deploy to 👍 
-

michaelsauter:
@gerardcl Thanks for approving! And yes, this should be 

I have updated the PR slightly:

* `master` has been updated with `selector` etc. so this is not part of this PR anymore
* Instead of relying on image labels on `BuildConfig` and `ImageStream` in order to figure out which images to pull from cd into the target project, we now use the information in the build artefacts. I believe this is a little better suited for the `ds-ml-service`, which changes the selector in the rollout stage, and also probably a better fit for monorepos, if they want to conditionally building some images (e.g. only when folder content changed).

henrjk (approved):
This LGTM.

michaelsauter:
@gerardcl @henrjk I added another small update 53f38fa, which adds the image reference to the build artefacts when built previously. Otherwise it would not be moved to the target project./

henrjk (dismissed):

gerardcl (dismissed):

michaelsauter:
@gerardcl @henrjk as discussed, I added b38cb4f to fix the rshiny issue - but specifying the resources explicitly is a good idea anyway ...

gerardcl (dismissed):
as a followup action we can cleanup ds-rshiny once this is merged
🚀

PR #519, opened 2020-12-02
Add discontinuations to RA doc history

jorge-romero:
Provide the discontinuations to the method that generates the history for RA


metmajer (approved):
Thanks @jorge-romero !

PR #522, opened 2020-12-02
Feature/gh issue 434

### LINKED ISSUE #434: Snyk creates many subproject for Gradle based repos
**Describe the bug**
Gradle creates many sub-projects for Gradle based repositories. This seems to be caused by Jenkins running every pipeline in a folder named after the pipeline (e.g. named `foo-cd-bar-123`) - this folder name is picked up for Gradle based projects by the Snyk CLI and passed as the "sub project" parameter. This does not happen for NPM based repos.

For more information, see https://support.snyk.io/hc/en-us/articles/360003817357-Snyk-for-Java-Gradle-Maven-#UUID-3d91eb09-1fcf-e973-89ff-bf24ab0ddfe2.

> Gradle sub-projects
>
> Gradle build can consist of several sub-projects, where each sub-project has its own build.gradle, while the root project is the only one that also includes a settings.gradle file. Sub-projects depend on the root project, but can be configured otherwise.
>
> By default, Snyk CLI scans only the current project (the project in the root of the current folder), or the project that is specified by --file=path/to/build.gradle):
>
> To scan all projects at once (recommended), use the --all-sub-projects flag:	
>   `snyk test --all-sub-projects`
> Note: Each of the individual sub-projects appears as a separate Snyk project in the UI.
> To scan a specific project (for example, myapp):
>  `snyk test --sub-project=myapp`

**Affected version**
OpenDevStack: master, but probably also 2.x

Discovered by @fkressmann

### LINKED ISSUE #521: Allow additional CLI options for Snyk Stage
Currently, we specify the cli options for `snyk test` and `snyk monitor` hardcoded in `SnykService`. We set `--org`, `--file`, `--severity-threshold` and `--all-sub-projects`. For one, `--all-sub-projects` is only of interest for gradle multi projects and is also causing issues (see #434). Also, the snyk cli is constantly changing and we don't really control which version of the cli is used in a project, because this can be set as an env var of the jenkins base-image (`SNYK_DISTRIBUTION_URL`). 

To avoid having cli version issues and others, we could just let the user set the official [cli options](https://support.snyk.io/hc/en-us/articles/360003812578-CLI-reference) in the Jenkinsfile by himself like this:
```
    odsComponentStageScanWithSnyk(context, ['severityThreshold'     : 'medium',
                                            'failOnVulnerabilities' : false,
                                            'snykAuthenticationCode': snykAuthenticationCode,
                                            'additionalOptions': ['all-sub-projects':'',
                                                                  'json':'',
                                                                  'show-vulnerable-paths':'all'
                                                                 ],
                                            ],
    )
```

renedupont:
Fixes #434 
Closes #521

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
@@ -40,4 +40,7 @@ Available options:
 
 | severityThreshold
 | Severity threshold for failing. If any found vulnerability has a severity equal or higher to the threshold, the snyk test will return with a failure status. Possible values are `low`, `medium`, `high`. Defaults to `low`.
+
+| additionalOptions
+| Additional options for the Snyk CLI. Please refer to the official Snyk CLI reference for possible options and don't forget to take the CLI version of your ODS installation into account. The value of `additionalOptions` must be a map in which the key is the official option name and the value a possible option value. If there is no option value, set it to the empty string `''`. Example: ['all-sub-projects':'', 'show-vulnerable-paths':'all']
 |===
--- a/docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
@@ -41,6 +41,6 @@ Available options:
 | severityThreshold
 | Severity threshold for failing. If any found vulnerability has a severity equal or higher to the threshold, the snyk test will return with a failure status. Possible values are `low`, `medium`, `high`. Defaults to `low`.
 
-| additionalOptions
-| Additional options for the Snyk CLI. Please refer to the official Snyk CLI reference for possible options and don't forget to take the CLI version of your ODS installation into account. The value of `additionalOptions` must be a list in which the entries have the official option name and a possible value. Example: ['--all-sub-projects', '--show-vulnerable-paths=all']
+| additionalFlags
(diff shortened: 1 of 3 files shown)


docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
```diff
@@ -40,4 +40,7 @@ Available options:
 
 | severityThreshold
 | Severity threshold for failing. If any found vulnerability has a severity equal or higher to the threshold, the snyk test will return with a failure status. Possible values are `low`, `medium`, `high`. Defaults to `low`.
+
+| additionalOptions
+| Additional options for the Snyk CLI. Please refer to the official Snyk CLI reference for possible options and don't forget to take the CLI version of your ODS installation into account. The value of `additionalOptions` must be a map in which the key is the official option name and the value a possible option value. If there is no option value, set it to the empty string `''`. Example: ['all-sub-projects':'', 'show-vulnerable-paths':'all']
```
michaelsauter, line 45: Did you consider making that a `List<String>` that would avoid the need to explain and set the empty string. So the example would be: `['--all-sub-projects, '--show-vulnerable-paths=all']` - what do you think? If you like that, I'd consider naming this `additionalFlags` ...

michaelsauter (commented):
Generally 👍 

Nice solution. Just one tiny question about the UX ...

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithSnyk.adoc
renedupont: good idea, changed it!

michaelsauter (approved):

PR #523, opened 2020-12-03
(component) application domain missing param - master

clemensutschig:
will fix part of opendevstack/ods-provisioning-app#646

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -499,11 +499,13 @@ class Context implements IContext {
 
     String getOpenshiftApplicationDomain () {
         if (!config.environment) {
+            logger.debug('Could not get application domain, as no environment is available')
             return ''
         }
         if (!this.appDomain) {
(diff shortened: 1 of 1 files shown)


src/org/ods/component/Context.groovy
```diff
@@ -499,11 +499,13 @@ class Context implements IContext {
...
     String getOpenshiftApplicationDomain () {
         if (!config.environment) {
+            logger.debug('Could not get application domain, as no environment is available')
             return ''
         }
         if (!this.appDomain) {
             logger.startClocked("${config.componentId}-get-oc-app-domain")
-            this.appDomain = ServiceRegistry.instance.get(OpenShiftService).applicationDomain
+            this.appDomain = ServiceRegistry.instance.get(OpenShiftService).
+                getApplicationDomain(config.environment)
```
michaelsauter, line 508: Should be `getApplicationDomain(config.targetProject)` I think.

michaelsauter (changes requested):

PR #524, opened 2020-12-03
Fix/display epics in csd

jorge-romero:
Provide all the information needed to display the epic information in the CSD


metmajer (approved):
Thanks @jorge-romero!

PR #525, opened 2020-12-03
Fixed NullPointerException when creating TRC document

jafarre-viewnext:


metmajer:
@jafarre-viewnext this PR does not make any sense. Every test case needs to have the property set. A null value would mean that a test case without a level would be a valid data record, which it is not. The error is not in the shared library, but in the data source. I do not approve.

jafarre-viewnext:
Closing this PR as the final fix is totally different.

Issue #533, opened 2020-12-08
Support of gitmodules


SHoen:
**Is your feature request related to a problem? Please describe.**
I would like to use gitmodules to resolve and split dependencies which are used in multiple components.

**Describe the solution you'd like**
If .gitmodules are configured the jenkins should init and update them together with the sourcecode of the repo

**Describe alternatives you've considered**
No alternatives are considered 

**Additional context**
https://git-scm.com/docs/gitmodules#_examples

serverhorror:
Strongly opposed to that. `subtree` is a superior alternative to submodules and will always work with a "simple" git clone

SHoen:
I agree that if you just include some dependency subtree is the way to go. But if you also own the component and want also to push code back gitmodules are the better choice. 
I would say, depends on your usecase.

PR #534, opened 2020-12-08
Feature - support gitmodules (master)

SHoen:
This is a PR to discuss the ticket **Support of gitmodules #533**

Tested on a Repo with gitmodules and on one without gitmodules.

### DIFF
--- a/src/org/ods/services/GitService.groovy
+++ b/src/org/ods/services/GitService.groovy
@@ -150,9 +150,7 @@ class GitService {
             userRemoteConfigs: userRemoteConfigs,
         ]
         if (isAgentNodeGitLfsEnabled()) {
-            gitParams.extensions = [
-                [$class: 'GitLFSPull']
-            ]
+            gitParams.extensions << [$class: 'GitLFSPull']
         }
         script.checkout(gitParams)
     }
--- a/src/org/ods/services/GitService.groovy
+++ b/src/org/ods/services/GitService.groovy
@@ -137,6 +137,15 @@ class GitService {
             $class: 'GitSCM',
             branches: [[name: gitCommit]],
             doGenerateSubmoduleConfigurations: false,
+            extensions: [[
+                    $class: 'SubmoduleOption',
+                    disableSubmodules: false,
+                    parentCredentials: true,
+                    recursiveSubmodules: true,
+                    reference: '',
+                    trackingSubmodules: false],
+                   [$class: 'CleanBeforeCheckout'],
+                   [$class: 'CleanCheckout']],
             submoduleCfg: [],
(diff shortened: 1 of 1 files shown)


clemensutschig:
@SHoen - cool. How do you use / invoke it.. ?

SHoen:
@clemensutschig  
You have a .gitmodules file where you define the modules. 

For example two modules. You define the name, the path where it is linked in the "main" repo and ref to the repos of the modules.  
 The file looks like this->

 [submodule "libfoo"]
	path = include/foo
	url = git://foo.com/git/lib.git

[submodule "libbar"]
	path = include/bar
	url = git://bar.com/git/lib.git


With "branch" you can point to specific branches if you need.

clemensutschig:
So no need for any configuration of the jenkinsfile? Very nice

SHoen:
Yep, no need to change the jenkinsfile. If there is a .gitmodules file in the repo, it will init and update the modules. If you don't have one, it just do normal checkout.

clemensutschig (approved):

clemensutschig:
@SHoen wonderfull

michaelsauter:
@SHoen Nice!

There are some other places which checkout code, too:
```
→ ag GitSCM
vars/odsOrchestrationPipeline.groovy
70:            $class: 'GitSCM',

src/org/ods/orchestration/InitStage.groovy
453:            $class: 'GitSCM',

src/org/ods/services/GitService.groovy
137:            $class: 'GitSCM',
199:            $class: 'GitSCM',
```

Maybe have a look at them as well / unify it?

metmajer:
@SHoen please consider @michaelsauter's advice. Otherwise, the Release Manager lacks support for Git submodules. Thanks for providing this fix, Steffen 👍

SHoen:
@michaelsauter  I had a look into in. 

> → ag GitSCM
vars/odsOrchestrationPipeline.groovy
70:            $class: 'GitSCM',
-> adding the functionality would be helpful. 
     Localbranch extension is used, I need to investigate why this is not used in the gitservice.
    Perhaps it makes sense to add it there as well

src/org/ods/orchestration/InitStage.groovy
453:            $class: 'GitSCM',
-> checkout of Release manager repo ( support of gitmodules not necessary)
     I think usage of the gitservice would make sense anyway

src/org/ods/services/GitService.groovy
137:            $class: 'GitSCM',
199:            $class: 'GitSCM',
-> is used by the other components (most times with localbranch extension) except of  pipeline.groovy 
-> is used also in MRO Pipeline Util together with RelativeTargetDirectory and another time withLocalBranch

I will create a PR as proposal. Just need to figure out how to test the changes, especially with the release manager.

PR #537, opened 2020-12-10
Implement Helm deploy

### LINKED ISSUE #495: [component pipeline] Allow deployment via Helm
**Is your feature request related to a problem?**
We support deployment of OpenShift Templates via [Tailor](https://github.com/opendevstack/tailor) at the moment. We do not have support for [Helm](https://helm.sh) because a) it used to require a privileged Tiller in v2 and b) it did not work with all OpenShift resources (such as BuildConfig, Route) in OpenShift v3.11. That said, OpenShift 4.3+ has full support for Helm, and offers a nice integration of Helm releases. Because of this and the widespread use of Helm in the community, we should support Helm in ODS.

Add Helm as an alternative to Helm in ODS 4. Recommend Tailor for OpenShift 3.11, and Helm for OpenShift 4. Remove support for Tailor in ODS 5, and rely on Helm only then.

[Kustomize](https://kustomize.io) is also a popular, tightly integrated way to specify Kubernetes resources. Potentially we should support this as well, but I see Helm as having higher priority.

[Operators](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) offer another way of packaging Kubernetes applications, and therefore have some overlap with Helm. However operators are more suited towards services which have a complex lifecycle (e.g. databases). Helm is probably a better fit for typical applications.

**Additional context**
In terms of process, Tailor and Helm are quite similar: 1) render resources, 2) apply resources. Therefore it should be relatively easy to integrate.

However, Helm has a few concepts that Tailor does not have, and we need to think about how we deal with those. E.g. the concept of "release", "package" and "rollback" are not present in Tailor.

On the other hand, Tailor has a few concepts that Helm does not have, e.g. diffing (could potentially be provided by [helm-diff](https://github.com/databus23/helm-diff)), secrets encryption (could potentially be provided by [helm-secrets](https://github.com/zendesk/helm-secrets)), preserving paths (not sure if there is a way to do this) and exporting live configuration (also not sure if there is a way to do this - but maybe that is a good thing? 😉).

For users who currently use OpenShift Templates, [template2helm](https://github.com/redhat-cop/template2helm) should provide a nice way to convert to Helm templates.

To support Helm, we need to look at two related, but separate topics: support in the component pipeline, and support in the orchestration pipeline. We should start with the component pipeline first.

FYI @segfault16 @serverhorror @gerardcl @henrjk @segator @metmajer @clemensutschig

michaelsauter:
Allows to use a chart in the repository to deploy via Helm (closes #495). I have implemented only very few options for now. I think we could start with this in `master` and review if we need more/different options once we have first users. The PR is a draft for now until we have gathered some feedback on the general direction, then I'll finish up the implementation/tests.

A comment towards general strategy: my view is that we should support Helm as an alternative to Tailor for the time being. Tailor is better suited for OpenShift 3.11, but for OpenShift 4, I would like us to transition to Helm. If you want to know more what this means and which implications this has, I'm starting to collect thoughts on this at https://github.com/opendevstack/tailor/wiki/Migrating-from-Tailor-to-Helm.

Would love to get your thoughts on the integration / strategy / implementation ... all comments welcome :)

FYI @segfault16 @serverhorror would especially appreciate your feedback as users of Helm if this would fit your use case

### DIFF
--- a/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
@@ -23,6 +23,21 @@ Available options:
 | deployTimeoutRetries
 | Adjust retries to wait for the pod during a rollout (defaults to 5). 
 
+| chartDir
+| Directory of Helm chart (defaults to `chart`).
+
+| helmReleaseName
+| Name of the Helm release (defaults to `context.componentId`). Change this value if you want to install separate instances of the Helm chart in the same namespace. In that case, make sure to use `{{ .Release.Name }}` in resource names to avoid conflicts.
+
+| helmValues
+| Map of key/value pairs to pass as values (by default, the key `imageTag` is set to the config option `imageTag`).
+
+| helmValuesFiles
+| List of paths to values files (empty by default).
+
+| helmAdditionalFlags
+| List of additional flags to be passed verbatim to the `helm` binary (empty by default).
+
 | openshiftDir
 | Directory with OpenShift templates (defaults to `openshift`).
 
--- a/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
@@ -27,28 +27,31 @@ Available options:
 | Directory of Helm chart (defaults to `chart`).
 
 | helmReleaseName
-| Name of the Helm release (defaults to `context.componentId`). Change this value if you want to install separate instances of the Helm chart in the same namespace. In that case, make sure to use `{{ .Release.Name }}` in resource names to avoid conflicts.
+| Name of the Helm release (defaults to `context.componentId`). Change this value if you want to install separate instances of the Helm chart in the same namespace. In that case, make sure to use `{{ .Release.Name }}` in resource names to avoid conflicts.  Only relevant if the directory referenced by `chartDir` exists.
 
 | helmValues
-| Map of key/value pairs to pass as values (by default, the key `imageTag` is set to the config option `imageTag`).
+| Map of key/value pairs to pass as values (by default, the key `imageTag` is set to the config option `imageTag`). Only relevant if the directory referenced by `chartDir` exists.
 
 | helmValuesFiles
-| List of paths to values files (empty by default).
+| List of paths to values files (empty by default). Only relevant if the directory referenced by `chartDir` exists.
 
 | helmDefaultFlags
-| List of default flags to be passed verbatim to to `helm upgrade` (defaults to `['--install', '--atomic']`). Typically these should not be modified - if you want to pass more flags, use `helmAdditionalFlags` instead.
+| List of default flags to be passed verbatim to to `helm upgrade` (defaults to `['--install', '--atomic']`). Typically these should not be modified - if you want to pass more flags, use `helmAdditionalFlags` instead. Only relevant if the directory referenced by `chartDir` exists.
 
 | helmAdditionalFlags
-| List of additional flags to be passed verbatim to to `helm upgrade` (empty by default).
+| List of additional flags to be passed verbatim to to `helm upgrade` (empty by default). Only relevant if the directory referenced by `chartDir` exists.
 
 | helmDiff
-| Whether to show diff explaining changes to the release before running `helm upgrade` (`true` by default).
+| Whether to show diff explaining changes to the release before running `helm upgrade` (`true` by default). Only relevant if the directory referenced by `chartDir` exists.
+
+| helmPrivateKeyCredentialsId
+| Credentials name of the private key used by helm-secrets (defaults to `<PROJECT>-cd-helm-private-key`). The fingerprint must match the one specified in `.sops.yaml`. Only relevant if the directory referenced by `chartDir` exists.
 
 | openshiftDir
 | Directory with OpenShift templates (defaults to `openshift`).
 
 | tailorPrivateKeyCredentialsId
-| Credentials name of the secret key used by Tailor (defaults to `<PROJECT>-cd-tailor-private-key`). Only relevant if the directory referenced by `openshiftDir` exists.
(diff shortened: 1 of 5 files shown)


michaelsauter:
One limitation with the current implementation is that there is no support for automatically decrypting secrets. I need to look into this going forward, but maybe we do not need to have integration with `helm-secrets` out-of-the-box initially.

segator:
questions:
* How the user can defined diferent values.yml files depending on the environment we are going to deploy? maybe we can use name convention? **$envName-values.yml** if not the user will need to code this logic in case they have diferent configuration in any environment.
* how we are going to version the charts?

Thanks!

michaelsauter:
@segator Thanks for looking over this. Good questions!

> How the user can defined diferent values.yml files depending on the environment we are going to deploy? maybe we can use name convention? $envName-values.yml if not the user will need to code this logic in case they have diferent configuration in any environment

Right now the answer would be:
```
odsComponentStageRolloutOpenShiftDeployment(context, [helmValuesFiles: ["${context.environment}-values.yml"]])
```
I'm not sure about some build-in logic. I have not found a "standard" naming for Helm values file. If there is one, please point me to it!

> how we are going to version the charts

I think we should not care about the chart version. Versioning charts seems to be a feature mostly targeted at sharing charts with others. Since ODS is consuming a chart from a local folder, I think the chart version does not matter. I would leave it up to the component author(s) to decide if and when they want to change the chart version. As far as I understand, the chart version does not play a specific role during `helm upgrade`. Do you know of any implication there? From what I understood from @segfault16, they do not increment the chart version at all.

segator:
regarding especific configuration per environment as far as I know helm doesn't have anything, for that we use **kustomize** or **helmfile** tools, but I imagine some kind of name convention in our shared library.

**default behavirour**
```
odsComponentStageRolloutOpenShiftDeployment(context)
```
then helm or chart folder with
```
helm\values.yml
helm\prod-values.yml
helm\test-values.yml
helm\dev-values.yml
```
then with helm we can always load the values.yml and the specific env values depending on the branch we are working.
the user then could in case this logic doesn't work for them override this logic with what you said
**overrided**
```
odsComponentStageRolloutOpenShiftDeployment(context, [helmValuesFiles: ["${context.environment}-values.yml"]])
```

as far as I know ```helm upgrade``` does not apply upgrade in case target version is same than installed unless you  use ```--force```

Also will be nice to use the ```--atomic``` parameter to rollback in case something fails, but here will be very important users use  **readinessprobe** on their deployment resources, or this parameter then is useless.

michaelsauter:
@segator Yes I was thinking about those default names I'm not yet sure I want to commit to them :)

Also thanks for the hint about `--force` and `--atomic`!

`--atomic` is described as `if set, upgrade process rolls back changes made in case of failed upgrade. The --wait flag will be set automatically if --atomic is used`, which sounds like something we would want. @henrjk @gerardcl @segfault16 agree?

About `--force`, I'm not so sure. It is described as `force resource updates through a replacement strategy`, but searching for this feature led me to https://github.com/helm/helm/issues/1958, which specifically talks about immutable fields. I believe it is a good default to prevent modification of immutable fields. I'd opt for leaving it up to the user to add this flag via `helmAdditionalFlags`.

segator:
**--force** I think will be needed if we don't care about the version helm chart, if we someday we decide to split the pipelines (one to build/publish the helm package (on Nexus?) and another for deploying) then we are safe to use versioning.(Something that in the future will be needed when people start to have dependencies on other company applications and we want reusability)

serverhorror:
I think the `--atomic` are good defaults but I would definitely like to be able to override them.

Especially the **readinessProbe** (or any probe for that matter) currently has problems in some OpenShift installations and isn't quite useable. That's even worse than it sounds.

michaelsauter:
0ebc03c sets `--atomic` by default now, but allows to override that.

@serverhorror I never had issues with the readiness probes - as they are part of Kubernetes, I doubt there is an OpenShift issue here ... Are you sure you have configured them correctly? :)

@segator I don't get how `--force` interacts with chart versioning. That said, your comment sounds like you are thinking about future potential use, and I suggest we take small steps for now and see how users begin to use this.

michaelsauter:
Opening this for review now. I believe this is good enough for a first draft. Also, we have some time until the next major release so we can still improve this over the next months.

segfault16 (approved):

segfault16:
@michaelsauter LGTM, all the functionality that we require is in 👍  Nice to see helm coming to ODS

Regarding versioning of charts: For applications using a single helm chart it's not really necessary to do so as they usually also update their values.yaml if something changes. For shared helm charts I'd consider it necessary. Especially if you have breaking changes in the interface (i.e. the values structure).
Another case where you'd want to version the charts is if you maintain your components in different git repos with different charts for the individual components and plugging them together by using subcharts into one application chart. This could be an approach for the orchestration pipeline I guess. 

On a related matter: Next step could be including a helm registry into ODS for re-using components throughout projects. Our team could supply initial charts for oauth2-proxy and matomo ready for use in OC. Beginning of next year we'll also tackle improved monitoring.

michaelsauter:
@segfault16 Thanks for your feedback!

I've added integration of `helm-diff` in 088a43b (on by default). It's a bit annoying that diffing is implemented by a plugin which means that not all params are supported. For now, let's keep diffing simple and when people run into issues, they can disable diffing, and share alternative approaches.

gerardcl (dismissed):
LGTM! good first steps to helm! 👍

michaelsauter:
@gerardcl Thanks!

I experimented with `helm-secrets` today and it turned out to be easy to integrate, so I did this in 814a2cd. Also, if you're coming from Tailor, it is quite easy to migrate - https://github.com/opendevstack/tailor/wiki/Migrating-from-Tailor-to-Helm has a section at the bottom now about encrypted secrets.

BTW, one nice property of `helm-secrets` and `helm-diff` is that secrets are redacted in the logs:
```
michael-dev, go-helm-master, Secret (v1) has been added:
+ # Source: go-helm/templates/secret.yaml
+ apiVersion: v1
+ kind: Secret
+ metadata:
+   labels:
+     app: michael-go-helm-master
+   name: go-helm-master
+ data:
+   mysecret: '++++++++ # (6 bytes)'
+ type: Opaque
```

gerardcl (approved):

PR #538, opened 2020-12-10
Bugfix EDPC-306: Fix NotSerializableException while creating TCP document

jafarre-viewnext:


stitakis (approved):

metmajer (changes requested):
@jafarre-viewnext can you please make the method private and use a return statement?

jafarre-viewnext:
> @jafarre-viewnext can you please make the method private and use a return statement?

Done

metmajer (approved):
@jafarre-viewnext thanks for the fix. Those issues will keep arising depending on the dynamics of your system. Over time, we will have to make iterator-heavy code like this @NonCPS'ed to be safe, since those are not serializable.

PR #541, opened 2020-12-16
add logger usage and logger tests consistently to services - master

clemensutschig:
@renedupont - can you please check as well, especially the snyk part

### DIFF
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -50,7 +50,11 @@ class ScanWithSnykStage extends Stage {
             script.error 'Snyk auth failed'
         }
         if (logger.debugMode) {
-            config.additionalFlags += '-d'
+            if (config.additionalFlags) {
+                config.additionalFlags += '-d'
+            } else {
+                config.additionalFlags = ['-d']
+            }
         }
         logger.info 'Scanning for vulnerabilities with ' +
             "organisation=${config.organisation}, " +
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -56,7 +56,8 @@ class ScanWithSnykStage extends Stage {
             "buildFile=${config.buildFile}, " +
             "failOnVulnerabilities=${config.failOnVulnerabilities}, " +
             "severityThreshold=${config.severityThreshold}, " +
-            "additionalFlags=${config.additionalFlags ? config.additionalFlags.toListString() : ''}."
+            "additionalFlags=${config.additionalFlags ? config.additionalFlags.toListString() : ''}." +
+            "${logger.debugMode ? '-d' : ''}"
 
         boolean noVulnerabilitiesFound
 
--- a/src/org/ods/component/ScanWithSnykStage.groovy
+++ b/src/org/ods/component/ScanWithSnykStage.groovy
@@ -49,15 +49,16 @@ class ScanWithSnykStage extends Stage {
         if (!snyk.auth(config.snykAuthenticationCode)) {
             script.error 'Snyk auth failed'
         }
-
+        if (logger.debugMode) {
+            config.additionalFlags += '-d'
+        }
         logger.info 'Scanning for vulnerabilities with ' +
             "organisation=${config.organisation}, " +
             "projectName=${config.projectName}, " +
             "buildFile=${config.buildFile}, " +
             "failOnVulnerabilities=${config.failOnVulnerabilities}, " +
             "severityThreshold=${config.severityThreshold}, " +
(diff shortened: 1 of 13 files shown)


src/org/ods/services/JenkinsService.groovy
```diff
@@ -23,10 +23,9 @@ class JenkinsService {
         }
 
         logger.debug "Collecting (applicable) testresults from location: '${xUnitResultDir}'"
-        def logCommand = logger.debugMode ? '' : 'set +x'
         script.sh(
             script: """
-                ${logCommand}
+                ${logger.shellScriptDebugFlag}
```
michaelsauter, line 28: Hmm, this is a bit confusing for me. Everywhere else, debugging is opt-in, whereas here it is opt-out. Not all places use `${logger.shellScriptDebugFlag}`, so most have `-x` (the Jenkins default), and only the few places here set `+x` when debugging is off.

I guess much more effort would be needed to have this consistent everywhere (e.g. by making `PipelineSteps` aware of this and modifying the passed `script`).

I'd probably leave this like it is unless we go all-in (which I would not do right now).

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -301,7 +301,7 @@ class OpenShiftService {
 
     int startBuild(String project, String name, String dir) {
         steps.sh(
-            script: "oc -n ${project} start-build ${name} --from-dir ${dir}",
+            script: "oc -n ${project} start-build ${name} --from-dir ${dir} ${logger.OCPDebugFlag}",
```
michaelsauter, line 304: Love the addition. As a nitpick, could we use `logger.ocDebugFlag`? I think that would better reflect what the flag is for (the `oc` binary) and avoids `OCPD` all uppercase ;)

src/org/ods/services/SonarQubeService.groovy
```diff
@@ -58,6 +61,7 @@ class SonarQubeService {
             script.sh(
                 label: 'Generate CNES Report',
                 script: """
+                ${logger.OCPDebugFlag}
```
michaelsauter, line 64: This looks odd - is this a mistake?

src/org/ods/component/ScanWithSnykStage.groovy
```diff
@@ -56,7 +56,8 @@ class ScanWithSnykStage extends Stage {
             "buildFile=${config.buildFile}, " +
             "failOnVulnerabilities=${config.failOnVulnerabilities}, " +
             "severityThreshold=${config.severityThreshold}, " +
-            "additionalFlags=${config.additionalFlags ? config.additionalFlags.toListString() : ''}."
+            "additionalFlags=${config.additionalFlags ? config.additionalFlags.toListString() : ''}." +
+            "${logger.debugMode ? '-d' : ''}"
```
michaelsauter, line 60: Personally I think we don't need the extra info about the debug mode here ...

renedupont, line 60: I assume what you actually wanted to have is:

```
        if (logger.debugMode) {
            config.additionalFlags += '-d'
        }

        logger.info 'Scanning for vulnerabilities with ' +
            "organisation=${config.organisation}, " +
            "projectName=${config.projectName}, " +
            "buildFile=${config.buildFile}, " +
            "failOnVulnerabilities=${config.failOnVulnerabilities}, " +
            "severityThreshold=${config.severityThreshold}, " +
            "additionalFlags=${config.additionalFlags ? config.additionalFlags.toListString() : ''}."
```

src/org/ods/services/JenkinsService.groovy
clemensutschig: @michaelsauter  confused... what do you want, or not :)?

src/org/ods/services/OpenShiftService.groovy
clemensutschig: nitpick, thats right ... I'll change that

src/org/ods/services/SonarQubeService.groovy
clemensutschig: ahh. great catch!

src/org/ods/component/ScanWithSnykStage.groovy
clemensutschig: I think it's worthwhile, especially when calling external jars / tools

renedupont: sorry, I just tested it, it needs to be:
```
        if (logger.debugMode) {
            if (config.additionalFlags) {
                config.additionalFlags += '-d'
            } else {
                config.additionalFlags = ['-d']
            }
        }
```

renedupont (approved):
snyk part looks good now

michaelsauter (approved):

PR #544, opened 2020-12-17
Typed Options

### LINKED ISSUE #542: [component pipeline] Improve options handling
Each stage takes a `Map config`, which is used to configure the behaviour of the stage (e.g. on which branches it is executed, or which `dockerDir` to use or similar). A map was chosen initially because it is easy to use from the `Jenkinsfile`, but there are a few downsides related to this:

* Misspelled keys are not detected
* The type of the passed properties is not checked, which can lead to weird errors further down
* The stage class cannot be annotated with `@TypeChecked` because `config` values have type `Object` and `config` is used in so many places that type checking becomes basically impossible
* Modifying the `config` map in the stage has the potentially surprising side-effect that the map passed from the `Jenkinsfile` is altered
* The Antora documentation does not have any information which type each config option has

To fix this, I have the following proposal:

* Create specific "Options" classes for each stage (e.g. `BuildOpenShiftImageOptions.groovy`)
* Instantiate one "Options" object per stage with the `Map config`, e.g. `this.options = new BuildOpenShiftImageOptions(config)`
* Add GroovyDoc comment for each property of the "Options" class

Once this is in place, we can also automatically create the documentation for those options:
* Render GroovyDoc documentation (HTML files) via `./gradlew groovydoc`
* Extract name, type and description of each "Options" class property
* Render Asciidoctor template (`foo.adoc.tmpl`) with the extracted info into the final document (`foo.adoc`)

For the last part, I made a POC:

```go
package main

import (
	"fmt"
	"os"
	"strings"
	"text/template"

	"github.com/PuerkitoBio/goquery"
)

// render-adoc.go extracts information from groovydoc HTML pages, and renders
// the content into Antora templates. This process ensures that the final
// documentation reflects the source code accurately, without manual effort.
//
// Before running "go run render-adoc.go", ensure to run "./gradlew groovydoc".
// This should place HTML documents into "build/docs/groovydoc". From the pages
// describing the "Options" classes, the property information (name, type,
// description) is extracted, and then rendered into the Antora partials.

const (
	groovydocComponentPath = "build/docs/groovydoc/org/ods/component"
	antoraPartialsPath     = "docs/modules/jenkins-shared-library/partials"
)

func main() {
	stages := []string{"BuildOpenShiftImage"}
	for _, stage := range stages {
		renderDocs(stage)
	}
}

func check(e error) {
	if e != nil {
		panic(e)
	}
}

type option struct {
	Name        string
	Type        string
	Description string
}

type docData struct {
	Options []option
}

func extractType(rawType string) string {
	cleanedType := strings.Replace(strings.TrimSpace(rawType), " ", "&nbsp;", -1)
	cleanedType = strings.Replace(cleanedType, "java.util.", "", -1)
	cleanedType = strings.Replace(cleanedType, "java.lang.", "", -1)
	return cleanedType
}

func getOptions(filename string) []option {
	optionsDocFile, err := os.Open(filename)
	check(err)
	optionsDoc, err := goquery.NewDocumentFromReader(optionsDocFile)
	check(err)
	options := []option{}
	optionsDoc.Find("div.details ul.blockListLast").Each(func(i int, s *goquery.Selection) {
		optionType := extractType(s.Find("li.blockList h4").Nodes[0].FirstChild.Data)
		optionName := strings.TrimSpace(s.Find("li.blockList h4 strong").Text())
		optionDesc := strings.TrimSpace(s.Find("p").Text())
		options = append(options, option{Name: optionName, Type: optionType, Description: optionDesc})
	})
	return options
}

func renderTemplate(targetFilename string, data docData) {
	targetFile, err := os.Create(targetFilename)
	check(err)
	tmpl, err := template.ParseFiles(targetFilename + ".tmpl")
	check(err)
	err = tmpl.Execute(targetFile, data)
	check(err)
}

func renderDocs(stageName string) {
	options := getOptions(fmt.Sprintf("%s/Options.html", groovydocComponentPath))
	stageOtions := getOptions(fmt.Sprintf("%s/%sOptions.html", groovydocComponentPath, stageName))
	options = append(options, stageOtions...)
	data := docData{Options: options}
	renderTemplate(
		fmt.Sprintf("%s/odsComponentStage%s.adoc", antoraPartialsPath, stageName),
		data,
	)
}
```

michaelsauter:
Closes #542. Read that issue first to get an idea of the problem this PR solves and the general approach.

Docs can be generated via `make doc`, and GitHub Actions checks whether docs are up-to-date (= generated `.adoc` matches `.groovy` source).

As a result of the typed options (and switching from `def script` to `PipelineSteps steps`), all stages are annotated with `@TypeChecked` now, which should help to reduce errors further.

Some notes:
* During the conversion, I noticed access to the private property `localCheckoutEnabled` on `Context`. I changed that to look at `context.triggeredByOrchestrationPipeline`, which should be equivalent.
* I did not change that the `config` map is still altered in the constructor, maybe we can do that separately in another PR
* The options are ordered alphabetically now. Does this make sense or what you prefer it differently?
* The approach can probably also applied to the `Context` class, but let's take this one step at a time.

This is ready for review, but before merging I still need to do:

- [x] Run `make test-quickstarter` for all quickstarters
- [x] Run tests for upload to nexus / Snyk (maybe adding them to the quickstarter tests somehow?)


renedupont:
Nice! There is one thing I don't understand yet: There are constructor calls like `this.options = new ImportOpenShiftImageOptions(config)` but I couldn't find the constructor code, neither in the specific class nor in the Options class. Is there some implicit magic going on?

henrjk:
Hi @renedupont see https://blog.mrhaki.com/2018/06/groovy-goodness-add-map-constructor.html

henrjk:
Awesome change! LGTM

henrjk (approved):

renedupont (approved):

PR #546, opened 2020-12-17
Include a better filter for the technical specifications

jorge-romero:
When a tech spec does not have software of design specification it is render on the pdf, and if it has any of them, it is displayed on the proper section of the SSDS document

### DIFF
--- a/src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
+++ b/src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
@@ -834,7 +834,7 @@ class LeVADocumentUseCase extends DocGenUseCase {
 
         def componentsMetadata = SortUtil.sortIssuesByKey(this.computeComponentMetadata(documentType).values())
         def systemDesignSpecifications = this.project.getTechnicalSpecifications()
-            .findAll { it.systemDesignSpec }
+            .findAll { it.systemDesignSpec || !it.softwareDesignSpec }
             .collect { techSpec ->
                 [
                     key        : techSpec.key,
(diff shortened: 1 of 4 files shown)


metmajer:
Nice fix @jorge-romero! I have converted this to a draft until we have a unit test that covers the behavior. Cheers!

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -1298,7 +1299,7 @@ class LeVADocumentUseCase extends DocGenUseCase {
             def metadata = repo_.metadata
 
             def sowftwareDesignSpecs = component.getResolvedTechnicalSpecifications()
-                .findAll { it.softwareDesignSpec }
+                .findAll { it.softwareDesignSpec || !it.systemDesignSpec }
```
s2oBCN, line 1302: I think it.softwareDesignSpec is part of  !it.systemDesignSpec

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -834,7 +834,7 @@ class LeVADocumentUseCase extends DocGenUseCase {
 
         def componentsMetadata = SortUtil.sortIssuesByKey(this.computeComponentMetadata(documentType).values())
         def systemDesignSpecifications = this.project.getTechnicalSpecifications()
-            .findAll { it.systemDesignSpec }
+            .findAll { it.systemDesignSpec || !it.softwareDesignSpec }
```
s2oBCN, line 837: Do we want to find all that are different than "it.systemDesignSpec"? (because it.softwareDesignSpec != it.systemDesignSpec)
If the return something different than "softwareDesignSpec " then return var name should be different?

jorge-romero: We want that the information of the Tech Specs appear in the document even if they dont have provide data for the soft/system design, so

-  if no information is provided on those fields, the issues are rendered in any section that must appear. In that moment as no extra information in the TechSpec are provided we don't now if that specifies a System or Software Design Spec, but it must be render in the pdf
- if there's information on any or both of the fields, the information is rendered in the proper section:
  There are two sections in the document, in one the System design specs are listed and another where the Software Design Specs are listed
 
In the code we are checking that systemDesignSpec or softwareDesingSpec contains any value

metmajer (approved):

PR #551, opened 2020-12-18
Feature/issue 293

### LINKED ISSUE #293: Retrieve services via context instead of registry?
I was thinking if it would be good to remove the `ServiceRegistry`, and instead add getters to the `Context` to retrieve e.g. the `OpenShiftService`.

Pros:
- No separate registry to deal with. Makes testing a little easier
- Services can be lazy loaded (on first access). This avoids having to create the instances initially, and to do fancy stuff like https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/odsComponentStageScanWithSonar.groovy#L9-L17 to make testing possible
- It would allow Jenkinsfile authors to easily access a properly configured service to use in their own stages

Cons:
- Context gets more methods

@metmajer @clemensutschig @oalyman @renedupont Thoughts on this?

renedupont:
fixes #293 

The idea is to have in each service class a getOrCreate method that is getting called in a getter method in Context.groovy or directly in cases where the context does not yet exist (e.g. in Pipeline.groovy) or is in some way not available.

I did not get rid of the ServiceRegistry because it is still used in the orchestration part and also for the Logger (which we might want to also put into the context?).

btw. sorry, I don't know why github shows some files as completely changed even though that is not true.


s2oBCN:
I think that now we have two ways to get access to the OpenShiftService, one if we don't have the context already created using 
and another from the context (_src/org/ods/orchestration/DeployStage.groovy_):
`def os = ServiceRegistry.instance.get(OpenShiftService)`
and another from the context:
`this.openShift = context.getOpenShiftService()`
This is weird, from my point of view...

Proposal:
Why don't we just call the **getOrCreate** from the same **ServiceRegistry**.  We can use the _Class<T> type_ to know the registry type and then use an enum to centralize the creation of all the services (like https://medium.com/@roanmonteiro/clean-code-with-java-replace-the-logical-condition-using-enum-if-else-statements-898bd6a85327).
This way we will have always the same way of calling a service and in a real lazy way.

renedupont:
interesting idea and pattern, would be really nice to have just this one way of getting or creating the services. I fear the issue with your idea is though that it requires that all getOrCreate implementations have the same argument list, which is (mostly because of BitbucketService) unfortunately not the case.
I'll think about if there is a way to overcome that issue.

s2oBCN:
If you like the idea, later I can give you a schema of the solution, the param is not a problem :)

renedupont:
yes ok, thinking a bit further though I think the real issue is not just the params, but also that these params are only available after the context was assembled, see here: https://github.com/opendevstack/ods-jenkins-shared-library/blob/9c24a89eafecac14eb51c3fbe7ff43c98e7da901/src/org/ods/component/Pipeline.groovy#L70

s2oBCN:
Sorry for the delay!
Yes, there is no a perfect solution :)

I like this approach:
`package org.ods.services

import org.ods.component.IContext
import org.ods.util.ILogger

@Singleton
class ServiceFactory {
    private final registry = [:]

    BitbucketService getBitbucket() {
        def service = registry[Services.BITBUCKET] as BitbucketService
        if (!service){
            throw new IllegalArgumentException("There is no instance created of this service to be used")
        }        
    }
    
    BitbucketService getBitbucket(def script, def env, String project, String passwordCredentialsId) {
        def service = registry[Services.BITBUCKET] as BitbucketService
        if (!service){
            def c = readConfigFromEnv(env)
            service = new BitbucketService(script, c.bitbucketUrl, project, passwordCredentialsId, getLogger())
            registry[Services.BITBUCKET] = service
        }
        return service
    }
    
    BitbucketService getBitbucket(def script, IContext context) {
        def service = registry[Services.BITBUCKET] as BitbucketService
        if (!service){
            service = new BitbucketService(
                this,
                context.bitbucketUrl,
                context.projectId,
                context.credentialsId,
                getLogger()
            )
            registry[Services.BITBUCKET] = service
        }
        return service
    }
    
    ILogger getLogger(){
        def service = registry[Services.BITBUCKET] as T
        if (!service){
            ...
        }
        return 
    }
    
    ...
    
    def clear() {
        registry.clear()
    }
}`

`package org.ods.services;

enum Services {
    BITBUCKET,
    GIT,
    ...
    LOGGER;
}
`

The pros of this is that we have the factory object where all the services are created (without external "setters") and we don't have static methods. But, yes, this can be done with some variations too.

It's just an idea, I know that you already implemented your solution and this one is not perfect too. :)

renedupont:
hey, we discussed a bit further and came to the conclusion that there is currently no real good benefit from trying these things. I'll close it until the rest of the code allows some day some more benefetting simplifications in this regard.

Issue #555, opened 2021-01-07
"Dead" code or undocumented repo type?


michaelsauter:
I've come across https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/util/MROPipelineUtil.groovy#L343-L362.

That part is, to my understanding, only executed if the repo is not of type `ods`, `ods-service` or `ods-test` (based on the `if` conditions above). However, according to the docs (https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/docs/modules/jenkins-shared-library/pages/orchestration-pipeline.adoc#configuration) there are only 3 repo types: `ods`, `ods-service` and `ods-test`.

Which makes me wonder if the shell script / Makefile part is dead code and can be removed, or if there is an undocumented component type, or if I am just confused and don't understand the code.

@metmajer do you know more on this?

clemensutschig:
I think it's dead code (as we somewhere default the type to `ods` from what I remember ... )

``` (Project 1003ff)
            // Set repo type, if not provided
            if (!repo.type?.trim()) {
                repo.type = MROPipelineUtil.PipelineConfig.REPO_TYPE_ODS_CODE
            }
```

clemensutschig:
so we should introduce `makefile` and `shellscript` imho

metmajer:
You're right @michaelsauter. This stems from the first executor implementation which happened before the `ods` type got integrated. Then, `ods` became a default as indicated by @clemensutschig which rendered this block a dead code in the end. Believe that there are some use-cases that would benefit from this in one or the other way, but good for me if we bury it in the Git history for now.

PR #388, opened 2020-06-25
[orchestration] Add beta support for ods-infra type - master

nichtraunzer:
Resolves #281

PR #557, opened 2021-01-11
[orchestration] Move repo execution from util to each stage 

michaelsauter:
tl;dr

Currently what's happening in each stage (build, deploy, test, release, finalize) for each repo is defined in one huge method in `MROPipelineUtil`. This method is complicated to read, error prone and untested. Further it's "far away" from the stage, and the way it's called from each stage is quite complicated to trace IMO. The PR moves the code of what is happening for each repo into each stage class, making it much easier to follow what's going on.

Long version: 

This should make it much easier to follow what's going on. The previous
version was so difficult to understand that each time I looked at it, I
had to trace the flow again. This was even more complicated by the fact
that the previous version used incorrect method types (which Jenkins
ignores), and no tests were written for those methods, either.

The new version should be simpler to understand, and has tests. Further, all
stages are now type checked, and there is at least one test run to
verify the bare minimum.

Note that this PR led me to discover issue https://github.com/opendevstack/ods-jenkins-shared-library/issues/555. In this PR, the behaviour was kept as-is so far though ... 

The quickstarter test for the `release-manager` quickstarter passes.

### DIFF
--- a/src/org/ods/orchestration/phases/DeployOdsComponent.groovy
+++ b/src/org/ods/orchestration/phases/DeployOdsComponent.groovy
@@ -9,7 +9,6 @@ import org.ods.services.GitService
 import org.ods.services.OpenShiftService
 import org.ods.services.JenkinsService
 import org.ods.services.ServiceRegistry
-import org.ods.services.GitService
 import org.ods.orchestration.util.DeploymentDescriptor
 import org.ods.orchestration.util.MROPipelineUtil
 import org.ods.orchestration.util.Project
@@ -18,10 +17,10 @@ import org.ods.orchestration.util.Project
 @TypeChecked
 class DeployOdsComponent {
 
-    private Project project
-    private IPipelineSteps steps
-    private GitService git
-    private ILogger logger
+    private final Project project
+    private final IPipelineSteps steps
+    private final GitService git
+    private final ILogger logger
     private OpenShiftService os
 
     DeployOdsComponent(Project project, IPipelineSteps steps, GitService git, ILogger logger) {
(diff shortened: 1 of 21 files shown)


src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -32,55 +31,52 @@ class DeployOdsComponent {
     }
 
     @TypeChecked(TypeCheckingMode.SKIP)
-    public void run(Map repo, String baseDir) {
+    void run(Map repo) {
```
michaelsauter, line 34: NOTE: This method is basically unchanged. The huge diff is only due to the removal of `steps.dir(baseDir) {}`. Switching to the correct working dir is done by the surrounding code now to reduce code duplication.

src/org/ods/orchestration/phases/FinalizeOdsComponent.groovy
```diff
@@ -30,7 +29,7 @@ class FinalizeOdsComponent {
         this.logger = logger
     }
 
-    public void run(Map repo, String baseDir) {
+    void run(Map repo) {
```
michaelsauter, line 32: NOTE: This method is basically unchanged. The huge diff is only due to the removal of `steps.dir(baseDir) {}`. Switching to the correct working dir is done by the surrounding code now to reduce code duplication.

src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -85,34 +90,32 @@ class MROPipelineUtil extends PipelineUtil {
         }
     }
 
-    private void executeODSComponent(Map repo, String baseDir) {
-        this.steps.dir(baseDir) {
-            if (repo.data.openshift.resurrectedBuild) {
-                logger.info("Repository '${repo.id}' is in sync with OpenShift, no need to rebuild")
-                return
-            }
+    void executeODSComponent(Map repo) {
```
michaelsauter, line 93: NOTE: This method is basically unchanged. The huge diff is only due to the removal of `steps.dir(baseDir) {}`. Switching to the correct working dir is done by the surrounding code now to reduce code duplication.

src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -296,90 +299,74 @@ class MROPipelineUtil extends PipelineUtil {
...
             [
                 [ $class: 'RelativeTargetDirectory', relativeTargetDir: "${REPOS_BASE_DIR}/${repo.id}" ],
-                [ $class: 'LocalBranch', localBranch: "**" ]
+                [ $class: 'LocalBranch', localBranch: "**" ],
             ],
             [[ credentialsId: credentialsId, url: repo.url ]]
         )
     }
 
-    Set<Closure> prepareExecutePhaseForRepoNamedJob(String name, Map repo, Closure preExecute = null, Closure postExecute = null) {
```
michaelsauter, line 305: This monster is the problem that this PR addresses. Instead of having this huge, complicated block here in utils, it is now easy to read from each stage.

src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -296,90 +299,74 @@ class MROPipelineUtil extends PipelineUtil {
...
-        return [
-            repo.id,
-            {
-                this.executeBlockAndFailBuild {
-                    def baseDir = "${this.steps.env.WORKSPACE}/${REPOS_BASE_DIR}/${repo.id}"
-                    def targetEnvToken = this.project.buildParams.targetEnvironmentToken
+    void runCustomInstructionsForPhaseOrSkip(String phase, Map repo) {
```
michaelsauter, line 308: The "custom phase" part can likely be removed, see #555.

src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -296,90 +299,74 @@ class MROPipelineUtil extends PipelineUtil {
...
+            Map<String, Closure> parallelTasks = repoGroup.collectEntries { repo ->
+                [
+                    repo.id,
+                    {
+                        this.executeBlockAndFailBuild {
+                            if (preRepoExecute) {
+                                preRepoExecute(repo)
+                            }
+                            if (repoExecute) {
+                                this.steps.dir(repoBaseDir(repo)) {
```
michaelsauter, line 348: This is where `steps.dir(baseDir)` is now centrally applied, allowing it to be removed from all other places.

clemensutschig (approved):
wow @michaelsauter  - well done! massive complexity reduction

michaelsauter:
@clemensutschig Thanks! :)

@metmajer I'd love to have an acknowledgement whether you've seen this and are OK with the approach before I merge this ... no hurry though.

metmajer:
Hi @michaelsauter, I love the changes. Thanks for investing your time into this. I would favour if we could wait with the merge until 3.x has been merged into master. Unfortunately, instead of integrating often, we amassed a large, long-running release branch that is quite a bit behind master. I'll make sure we apply better practices going forward. I expect this to happen in the next days. Thanks!

michaelsauter:
@metmajer Cool. Let's wait then. I'm aware that this PR conflicts with https://github.com/opendevstack/ods-jenkins-shared-library/pull/388 as well. I'd prefer to merge this PR before #388, and once that is done, will open a PR into #388 to make it mergeable again.

metmajer:
@michaelsauter absolutely, #388 will follow. I don't expect the latter to close down so soon.

metmajer:
@michaelsauter we need to change plans quickly. The AWS Quickstarter arrived at a working state and we needed to make sure that the early access users don't have to wait until #388 gets merged. #570 has just been merged into branch `3.x`. The merge to `master` will follow in the next days.

michaelsauter:
OK, then I will adapt this PR once the other parts are stable. Please let me know when this is the case.

PR #558, opened 2021-01-13
fixed spelling mistake

stevegbrooks:
I assume it should be `odsQuickstarterPipeline` and not `odsQuickstgarterPipeline`


michaelsauter (approved):
Thanks 👍

PR #565, opened 2021-01-22
feat: improve skipping commits

SimonGolms:


src/org/ods/services/GitService.groovy
```diff
@@ -124,12 +124,17 @@ class GitService {
...
+     */
     boolean isCiSkipInCommitMessage() {
-        return script.sh(
-            returnStdout: true, script: 'git show --pretty=%s%b -s',
+        def gitCommitMessage = script.sh(
+            returnStdout: true, script: 'git show --pretty=%s -s',
             label: 'check skip CI?'
-        ).toLowerCase().replaceAll('[\\s\\-\\_]', '').contains('[ciskip]')
+        ).toLowerCase().replaceAll('[\\s\\-\\_]', '')
+        return gitCommitMessage.contains('[ciskip]') || gitCommitMessage.contains('[skipci]') || gitCommitMessage.contains('***noci***')
```
michaelsauter, line 137: It would be nice to have this tested. Looking at `GitServiceSpec` there is no existing test right now, so totally understand you did not add one. But it would be great to have this. Want to add one?

docs/modules/jenkins-shared-library/partials/odsComponentPipeline.adoc
```diff
@@ -414,4 +414,22 @@ def sayHello(def context) {
...
+[source,sh]
+skip
+----
+# skips pipeline (one-line)
+$ git commit -m "docs: update README [ci skip]"
+
+# skips pipeline (multi-line)
+$ git commit -m "docs: update README [ci skip]
+- add section installation
+"
```
michaelsauter, line 427: I think this example is not really needed. Having the first and the last is probably enough ... otherwise this gets a bit lengthy.

src/org/ods/services/GitService.groovy
SimonGolms: Sure, I'm happy to do it, but my groovy skills are based more on "copy&paste"-ish than real knowledge...
How do I call the function correctly, as I have only found the following reference for it and wondering if the missing prefix `is` intended?
https://github.com/opendevstack/ods-jenkins-shared-library/blob/4cdd980a979ffbf148b528647aa9dba94fa31307/src/org/ods/component/Pipeline.groovy#L325-L328:

michaelsauter: I was suggesting to simply add a test in `GitServiceSpec`. I think you do not need to go all the way to test integration with `Pipeline` (would be great of course but let's take small steps). If you look into `GitServiceSpec`, you should see an example that you can copy&paste ;)

michaelsauter (approved):
👌

michaelsauter (changes requested):
CodeNarc failed though, see the report at https://github.com/opendevstack/ods-jenkins-shared-library/suites/1881243145/artifacts/36727848.

SimonGolms:
> CodeNarc failed though, see the report at https://github.com/opendevstack/ods-jenkins-shared-library/suites/1881243145/artifacts/36727848.

Unfortunately, I do not know what to do with it. Is there anything to do here from my side?

clemensutschig:
@SimonGolms  - ps - you can run codeNarc also locally :)

src/org/ods/component/Pipeline.groovy
```diff
@@ -149,7 +149,8 @@ class Pipeline implements Serializable {
 
                     skipCi = isCiSkip()
                     if (skipCi) {
-                        logger.info 'Skipping build due to [ci skip] in the commit message ...'
+                        logger.info "Skipping build due to [ci skip], [skip ci] or ***NO_CI***" +
```
clemensutschig, line 152: replace `"` with `'` there is no resolution needed

clemensutschig (approved):

test/groovy/org/ods/services/GitServiceSpec.groovy
```diff
@@ -18,5 +18,73 @@ class GitServiceSpec extends SpecHelper {
...
+        """
+        def script = new PipelineScript()
+        def service = new GitService(script, new Logger(script, false))
+
+        when:
+        def isSkippingCommitMessage = service.isCiSkipInCommitMessage(gitCommitMessage)
+
+        then:
+        isSkippingCommitMessage == false
+    }
```
michaelsauter, line 89: I'm fine with this, but note that there is a way to condense this drastically, and make it easier to understand what you are actually testing ... using `where` "tables" ... see e.g. https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/test/groovy/org/ods/services/OpenShiftServiceSpec.groovy#L12-L31

SimonGolms: Nice! I was not aware of this before and adjusted the test cases respectively.

michaelsauter (approved):
Ace 👌

PR #569, opened 2021-01-29
Extract information from deltadocgen report that belongs to previous versions and use it to resolve references

jafarre-viewnext:


metmajer (approved):
Massive thanks for the long-nighter, well done @jafarre-viewnext!

PR #571, opened 2021-01-30
(orchestration) fail build after ALL unit tests have been executed - master

clemensutschig:
ok - the idea is pretty simple .. fail delayed if in `WIP` and only during `assemble` after all unit tests executed /in all repos - rather than fail fast (immediately)

@metmajer - please verify

### DIFF
--- a/src/org/ods/orchestration/BuildStage.groovy
+++ b/src/org/ods/orchestration/BuildStage.groovy
@@ -47,8 +47,9 @@ class BuildStage extends Stage {
                         [Project.TestType.UNIT],
                         data.tests.unit.testResults
                     )
-                    // we warn in any case ... (largely because the above call will
-                    // return immediatly when no jira adapter is configured.
+                    // we check in any case ... (largely because the above call will
+                    // return immediatly when no jira adapter is configured).
+                    // this  will set failedTests if any xunit tests have failed
                     util.warnBuildIfTestResultsContainFailure(data.tests.unit.testResults)
                 } else {
                     logger.info("[${repo.id}] Resurrected tests from run " +
@@ -82,8 +83,12 @@ class BuildStage extends Stage {
 
         // in case of WIP we fail AFTER all pieces have been executed - so we can report as many
         // failed unit tests as possible
-        if (project.isAssembleMode && project.isWorkInProgress && project.hasFailingTests) {
-            util.failBuild ("Unit tests failed, please look for repositoriy errors!")
+        // - this will only apply in case of WIP! - otherwise failfast is configured, and hence
+        // the build will have failed beforehand
+        def failedRepos = repos.flatten().findAll { it.data?.failedStage }
+        if (project.isAssembleMode && project.isWorkInProgress &&
+            (project.hasFailingTests || failedRepos.size > 0)) {
+            util.failBuild ("Failing build as repositoies contain errors!\nFailed: ${failedRepos}")
         }
     }
 
--- a/src/org/ods/orchestration/BuildStage.groovy
+++ b/src/org/ods/orchestration/BuildStage.groovy
@@ -47,6 +47,9 @@ class BuildStage extends Stage {
                         [Project.TestType.UNIT],
                         data.tests.unit.testResults
                     )
+                    // we warn in any case ... (largely because the above call will
+                    // return immediatly when no jira adapter is configured.
+                    util.warnBuildIfTestResultsContainFailure(data.tests.unit.testResults)
                 } else {
                     logger.info("[${repo.id}] Resurrected tests from run " +
                         "${repo.data.openshift.resurrectedBuild} " +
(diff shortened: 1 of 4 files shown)


src/org/ods/orchestration/BuildStage.groovy
```diff
@@ -76,6 +79,12 @@ class BuildStage extends Stage {
         }
         executeInParallel(executeRepos, generateDocuments)
         levaDocScheduler.run(phase, MROPipelineUtil.PipelinePhaseLifecycleStage.PRE_END)
+
+        // in case of WIP we fail AFTER all pieces have been executed - so we can report as many
+        // failed unit tests as possible
+        if (project.isAssembleMode && project.isWorkInProgress && project.hasFailingTests) {
```
clemensutschig, line 85: @metmajer extend with a Check against all repos .. if there is a failed stage..

clemensutschig: @metmajer  - I added what we talked about - please verify .. I don't have a good testcase at hand

src/org/ods/orchestration/BuildStage.groovy
```diff
@@ -47,6 +47,10 @@ class BuildStage extends Stage {
                         [Project.TestType.UNIT],
                         data.tests.unit.testResults
                     )
+                    // we check in any case ... (largely because the above call will
+                    // return immediatly when no jira adapter is configured).
+                    // this  will set failedTests if any xunit tests have failed
+                    util.warnBuildIfTestResultsContainFailure(data.tests.unit.testResults)
```
metmajer, line 53: @clemensutschig I am not quite sure about this one. Consider that the call to `jira.reportTestResultsForComponent` a few lines above invokes this method already. I think this to be superfluous.

metmajer (changes requested):
@clemensutschig contribution is most welcome and, although not ideal, I will not merge this change into `3.x`. This requires some time to share thoughts and do proper testing, and unfortunately, we cannot afford a single day of delay at this point in time.

Just to summarise the approach we had discussed:
- The correct behaviour of the Release Manager in the context of reporting failed tests should not depend on the failure mode of ODS components (exit code of test execution of both `0` and `!= 0` are fine).
- [`MROPipelineUntil.prepareExecutePhaseForRepoNamedJob`](https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/src/org/ods/orchestration/util/MROPipelineUtil.groovy#L307) calls `executeODSComponent` within `executeBlockAndFailBuild`. The latter throws `Throwable` when an component's pipeline emits a failure.
- When a Release Manager's `Stage` catches this `Throwable`, currently in [`Stage.execute`](https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/src/org/ods/orchestration/Stage.groovy#L41), components which have failures AND failing test results shall report these failures into Jira by calling [`JiraUseCase.reportTestResultsForComponent`](https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/src/org/ods/orchestration/usecase/JiraUseCase.groovy#L285).

Good with the approach?

src/org/ods/orchestration/BuildStage.groovy
clemensutschig: Martin.. the warnbuild is idempotent.. and does Not report anythibg to jira.. it parses the Junit Files .. and if failure sets hasfailing tests

clemensutschig:
> 
> Good with the approach?

And thats whats implemented in this pr.. 

Execute Component will during build Not throw but just set failure and warn.. in the build stage ... the last piece is the failure in Case of either failed Tests or other Component failures

src/org/ods/orchestration/BuildStage.groovy
```diff
@@ -76,6 +80,16 @@ class BuildStage extends Stage {
...
         levaDocScheduler.run(phase, MROPipelineUtil.PipelinePhaseLifecycleStage.PRE_END)
+
+        // in case of WIP we fail AFTER all pieces have been executed - so we can report as many
+        // failed unit tests as possible
+        // - this will only apply in case of WIP! - otherwise failfast is configured, and hence
+        // the build will have failed beforehand
+        def failedRepos = repos.flatten().findAll { it.data?.failedStage }
+        if (project.isAssembleMode && project.isWorkInProgress &&
+            (project.hasFailingTests || failedRepos.size > 0)) {
+            util.failBuild ("Failing build as repositoies contain errors!\nFailed: ${failedRepos}")
```
michaelsauter, line 91:
```suggestion
            util.failBuild("Failing build as repositories contain errors!\nFailed: ${failedRepos}")
```

michaelsauter (commented):
In general, looks good to me.

I think it is the right approach to handle this on the orchestration level, and not on the component level.

hrcornejo:
With unit tests failing:
image
The test are updated:
image
The behaviour is also correct with a green suite:
image
Verified over an updated 3.x branch with changes of this PR.

metmajer (approved):

PR #572, opened 2021-02-01
WIP: wrap execution of components in master node (master)

clemensutschig:
as we have seen in a couple of occasions (deploy stage :)) - we should "force" wrap the execution of a component on the master node.


clemensutschig:
the diff is rather big - because of it's intendation ... in reality it's just `steps.node` and the copy of the repo `steps.sh (cp ...`

src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -307,60 +307,70 @@ class MROPipelineUtil extends PipelineUtil {
...
-                        } else {
-                            this.logger.debug("Repo '${repo.id}' is of type ODS Service Component. Nothing to do in phase '${name}' for target environment '${targetEnvToken}'.")
-                        }
-                    } else if (repo.type?.toLowerCase() == PipelineConfig.REPO_TYPE_ODS_TEST) {
-                        if (name == PipelinePhases.TEST) {
-                            executeODSComponent(repo, baseDir)
+                    // force this to happen on master - if an agent is needed, the closure has to
+                    // take care about it. The reason for this is the closure to have access to files
+                    // from the component, e.g. jenkinsfile, shell scripts,...
+                    this.steps.node {
```
michaelsauter, line 319: Oh I think I misunderstood you earlier then.

I believe what you are doing here can't be done. `DeployOdsComponent#run` needs to be run on an agent node to deploy to external clusters. With your change, you'd go back to master, right?

Also note that running `node {}` has weird effects on the workspace it uses. I would avoid using if at all possible.

I would suggest not to make a change like this. I know this means that the whole `deploy` phase runs on an agent node then ... and therefore cannot execute a component's Jenkinsfile. This is a big limitation, but so far my understanding is that the AWS quickstarter can hook into the `build` phase instead.

TBH, the whole master/agent thing is a mess, and will cause us lots of issues. I don't see too many options going forward. The cleanest thing would probably be to move the switch to an agent into `DeployOdsComponent` (at the cost of doing this many times for many components, which is ugly as well).

We'll need to do more thinking around this ...

michaelsauter (changes requested):

src/org/ods/orchestration/util/MROPipelineUtil.groovy
clemensutschig: @michaelsauter  - the other option would be to move this into execute ods component - and make it as surgical as possible

michaelsauter: right that might work ...

clemensutschig: > right that might work ...

let me know if you are happi(er) with that approach

PR #575, opened 2021-02-04
add post comment function

longbowww:
Proposal for post Tag functionality to avoid manually interacting with GIT through jenkins.

### DIFF
--- a/src/org/ods/services/BitbucketService.groovy
+++ b/src/org/ods/services/BitbucketService.groovy
@@ -250,6 +250,35 @@ class BitbucketService {
         }
     }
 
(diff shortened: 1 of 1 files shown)


clemensutschig (approved):
@longbowww - very cool thanks much .. where and how do you use it?

@michaelsauter  - where do we use it?

longbowww:
I would like to use it for NPM modules that get pushed to Nexus.

While I can do build/npm publish steps on jenkins, what I missed was a convenient way to also git-tag my built versions to have a quick overview of what git commit belongs to what published npm version.

Release Mngr kind of does the same thing with version builds.

clemensutschig:
really nice .. thanks much @longbowww

PR #579, opened 2021-02-16
Fix discontinued issues appearing in the documents

jafarre-viewnext:


metmajer (approved):
Great job @jafarre-viewnext and @jorge-romero!

PR #588, opened 2021-02-23
Fix error getting Document Chapter heading number when versioning is disabled

jafarre-viewnext:
For backwards compatibility, when versioning is disabled, allow documentation chapters without heading number.


metmajer:

PR #590, opened 2021-02-23
Fix backwards compatibility error when tests do not have components

jafarre-viewnext:
For backwards compability, when versioning is disabled, allow tests without one component. Turned the error into a warning when the number of components is not exactly 1.


metmajer:

PR #592, opened 2021-02-23
Fix error trying to get "ProductRelease Version" field for projects without versioning

jafarre-viewnext:
Projects with the versioning capability disabled don't have a ProductRelease Version defined, and trying to get it results in an exception.
For backwards compability, it has been fixed by considering the absence of this field as an indication that vesioning is not enabled.


metmajer (approved):

metmajer:

PR #607, opened 2021-03-19
Add ods-saas-service component

metmajer:
Add a new `ods-saas-service` component type to manage the automated documentation of external, vendor-provided services.

### DIFF
--- a/src/org/ods/orchestration/util/MROPipelineUtil.groovy
+++ b/src/org/ods/orchestration/util/MROPipelineUtil.groovy
@@ -25,6 +25,7 @@ class MROPipelineUtil extends PipelineUtil {
 
         static final String REPO_TYPE_ODS_CODE = "ods"
         static final String REPO_TYPE_ODS_INFRA = "ods-infra"
+        static final String REPO_TYPE_ODS_SAAS_SERVICE = "ods-saas-service"
         static final String REPO_TYPE_ODS_SERVICE = "ods-service"
         static final String REPO_TYPE_ODS_TEST = "ods-test"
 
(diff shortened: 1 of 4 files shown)


src/org/ods/orchestration/util/MROPipelineUtil.groovy
```diff
@@ -339,6 +340,12 @@ class MROPipelineUtil extends PipelineUtil {
                         } else {
                             this.logger.debug("Repo '${repo.id}' is of type ODS Infrastructure as Code Component/Configuration Management. Nothing to do in phase '${name}' for target environment'${targetEnvToken}'.")
                         }
+                    } else if (repo.type?.toLowerCase() == PipelineConfig.REPO_TYPE_ODS_SAAS_SERVICE) {
+                        if (this.project.isPromotionMode && name == PipelinePhases.DEPLOY) {
```
clemensutschig, line 344: @metmajer ... i think none of this is needed... there is no jenkinsfile 
. ?

metmajer: @clemensutschig been thinking the same... but what if account creation, etc. is needed? Don't documents claim that there has to be a Jenkinsfile?

clemensutschig: If its that way... then this needs to happen in the build phase as well.. similar to ods infra..

metmajer: Wait a minute... if we would support account creation, etc., then we would need to document this in a TIR, I guess -- which we don't want to, right?

clemensutschig: In a saas case Everything is done by the vendor..  so i vouch to have nothing in there.. so an empty ^if^

clemensutschig:
ok - I am almost thru ... we have to talk about long component names for the SSDS .. which I changed now, and which make the pdf check fail ... @metmajer

PR #614, opened 2021-03-25
SSDS with PR info not SonarQube

hrcornejo:
Generate a CSV file with bitbucket info (commits and PR over default branch per every repo) to use later in the generation of SSDS document.


src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
```diff
@@ -0,0 +1,170 @@
...
+        def nextPage = true
+        def nextPageStart = 0
+        while (nextPage) {
+            def commits = bitbucketService.getCommitsForMainBranch(token, repo.repo, PAGE_LIMIT, nextPageStart)
+            if (commits.isLastPage) {
+                nextPage = false
+            } else {
+                nextPageStart = commits.nextPageStart
+            }
+            processCommits(token, repo, commits, file)
```
michaelsauter, line 55: If I read this right, this make one API request per page until all commits have been found, and for each commit, make another API request to find any pull requests.

This sounds like a very expensive operation - which will get more expensive over time. Did you calculate the impact on the Bitbucket Server?

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
```diff
@@ -0,0 +1,170 @@
...
+                    getReviewers(mergedPR.values[0].reviewers),
+                    mergedPR.values[0].links.self[(0)].href,
+                    commit.id,
+                    repo.repo)
+                writeCSVRecord(file, record)
+            }
+        }
+    }
+
+    private void writeCSVRecord(File file, Record record) {
```
michaelsauter, line 84: Why CSV? Why not JSON?

hrcornejo: @michaelsauter the requirement was a CSV, @metmajer can give you more reasons :)

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
hrcornejo: Yes, that is the process. And no, right now we don't have calculated the impact, we are making test to probe it.

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
metmajer: @hrcornejo the current implementation using a CSV file as a result is only a prototype. The final solution removes the CSV file again and puts the data into a table within the SSDS document's section "2.3 Code Reviews". Please put this PR into Draft mode and use the same branch for delivering the final solution. Thanks!

michaelsauter: Why not use a data structure that can be marshalled to JSON? That way you can upload this and process it further if need be.

metmajer: I don't know either, but I would suggest the same, @hrcornejo.

metmajer:
@hrcornejo the current implementation using a CSV file as a result is only a prototype. The final solution removes the CSV file again and puts the data into a table within the SSDS document's section "2.3 Code Reviews". Please put this PR into Draft mode, adjust the title to reflect the end goal, and continue to use this branch for delivering the final solution. Thanks!

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * Create a CSV file that contains the following records
+     * for every merge event into the integration branch of every ODS component:
+     * @return absolutePath of the created file
+     */
+    @SuppressWarnings(['JavaIoPackageAccess'])
+    String generateSourceCodeReviewFile() {
+        def file = new File("${steps.env.WORKSPACE}/${CSV_FILE}")
```
metmajer, line 34: We typically use some namespace directories to differentiate different types of files. Please avoid writing this file directly into env.WORKSPACE. See MROPipelineUtil class members for examples.

test/groovy/org/ods/orchestration/usecase/BitbucketTraceabilityUseCaseSpec.groovy
```diff
@@ -0,0 +1,121 @@
...
+    BitbucketService bitbucketService
+
+    def setup() {
+        log.info "Using temporal folder:${tempFolder.getRoot()}"
+
+        steps = new PipelineSteps()
+        steps.env.WORKSPACE = tempFolder.getRoot().absolutePath
+        logger = new LoggerStub(log)
+        project = buildProject(logger)
+        bitbucketServiceMock = new BitbucketServiceMock().setUp("csv").startServer(RECORD_WIREMOCK, BB_URL_TO_RECORD)
```
metmajer, line 50: Why not use the WireMock server utils in SpecHelper.groovy?

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        return result
+    }
+
+    private void processRepo(String token, Map repo, File file) {
+        def nextPage = true
+        def nextPageStart = 0
+        while (nextPage) {
+            def commits = bitbucketService.getCommitsForMainBranch(token, repo.repo, PAGE_LIMIT, nextPageStart)
```
metmajer, line 89: As a minor improvement, I suggest to change the method name from *Main* into *Integration*, since `repo.repo` names the integration branch.

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
metmajer: Hector, found out that we reach a limit at ~8k commits. The reason is not yet clear, since there is no error being shown. @hrcornejo @jorge-romero we should consider only including those commits that belong to the current product version. This requires keeping track of the commit SHA of the previous releases.

metmajer:
@hrcornejo @jorge-romero shall we "updraft" and proceed with the review?

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
hrcornejo: Created in new "review" folder

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
hrcornejo: Changed to: _getCommitsForIntegrationBranch_

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
michaelsauter: Would other endpoints help us instead? E.g. going through the merged pull requests instead, and then plucking the ones we want to use?

michaelsauter: Ah. I think there is another endpoint that gives you "PR actions" or something like this which does show the merged commit. I would have to look it up

hrcornejo:
Updated the templates in https://github.com/opendevstack/ods-document-generation-templates/pull/36

metmajer:
@hrcornejo is this ready to go? If yes, please resolve conflicts and let's get this merged.

hrcornejo:
@metmajer solved, pending approval to merge

src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
metmajer: @hrcornejo @michaelsauter we should definitely investigate this.

metmajer (approved):
Great job! I approve, but we need to ensure we integrate with Jira in the most efficient way. @michaelsauter has proposed an alternative REST API to be used. Let's follow up on this in the next iteration.

metmajer:
@hrcornejo please squash and merge into a single commit. Cheers!

PR #617, opened 2021-03-26
Implementation of Aqua Security stage

renedupont:
### This PR integrates an Aqua Security stage into the shared-lib.

**General Information:**
This stage runs only if an _imageRef_ was set in the image build stage, otherwise the stage is discontinued because it is assumed no new image to scan was build. It tries to retrieve it this way:
```
        Map<String, String> buildInfo =
            context.getBuildArtifactURIs().builds[options.resourceName] as Map<String, String>
```
After the scan, a Bitbucket Code Insight Report is added to the commit and can be seen in a PR. An HTML report is also archived as an artifact and can be accessed in Jenkins.

**This stage comes with these prerequisites that need to be configured beforehand:**
- A _jenkins agent image_ that has the aqua _scannercli_ included to trigger the scan process
- Credentials (username and password) in a `Secret` of an aqua platform user that has scanner rights
- A `ConfigMap` in the project-cd namespace that has these fields:
```
...
metadata:
  name: aqua
...
data:
  registry: <registry-name-in-aqua-platform>
  secretName: <secret-name-of-aqua-user-credentials>
  url: <aqua-platform-url>
```
_aqua.secretName_ defaults to `cd-user-with-password` if not set.
This `ConfigMap` is used to connect to the aqua platform during the scan process.


metmajer:
@renedupont some things to consider:

- Currently, the HTML report is archived in Jenkins only, but shall be archived in Nexus with a link made available inside the Bitbucket PR. Jenkins artefacts are far not considered reliable enough (short retention period).
- We recently discussed the need to globally enable/disable Aqua centrally on an OpenShift cluster (Aqua = a feature of the platform, not of a project). I haven't look at the entire code yet, but we need to ensure that Aqua scanning cannot be disabled at the project/component levels.

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -0,0 +1,122 @@
...
+            config.resourceName = context.componentId
+        }
+        this.options = new ScanWithAquaOptions(config)
+        this.aqua = aqua
+        this.bitbucket = bitbucket
+        this.openShift = openShift
+    }
+
+    protected run() {
+        Map connectionData = openShift.getConfigMapData(context.cdProject, "aqua")
```
michaelsauter, line 34: Could you add some documentation about this config map? How / where do I create it? What keys / values does it need?

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -0,0 +1,122 @@
...
+        // base URL of Aqua server.
+        String aquaUrl = connectionData['aqua.url']
```
michaelsauter, line 36: If I read this right, the name of the key is `aqua.url` - wouldn't `url` be enough as the scope is already the `aqua` `ConfigMap`?

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -0,0 +1,122 @@
...
+        if (!registry) {
+            steps.error "Please provide the name of the registry that contains the image of interest!"
+        }
+        // name of the credentials that stores the username/password of a user with access
+        // to the Aqua server identified by "aquaUrl", defaults to the cd-user
+        String secretName = connectionData['aqua.secretName']
+        String credentialsId = context.cdProject + "-"
+        if (!secretName) {
+            credentialsId = context.credentialsId
+            logger.info("No custom secretName was specified, continuing with default " +
```
michaelsauter, line 51: Maybe not entirely clear that this `secretName` can be set in the `ConfigMap`?

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -0,0 +1,122 @@
...
+        aqua.scanViaCli(aquaUrl, registry, imageRef, credentialsId, reportFile)
+        logger.infoClocked(options.resourceName, "Aqua scan (via CLI)")
+    }
+
+    private createBitbucketCodeInsightReport(String aquaUrl, String registry, String imageRef) {
+        String aquaScanUrl = aquaUrl + "/#/images/" + registry + "/" + imageRef.replace("/", "%2F") + "/vulns"
+        String title = "Aqua Security"
+        String details = "Please visit the following link to review the Aqua Security scan report:"
+        // for now, we set the result always to successful in the aqua stage
+        String result = "PASS"
```
michaelsauter, line 94: Why? Is this because we do not have the actual result here?

src/org/ods/services/AquaService.groovy
```diff
@@ -0,0 +1,64 @@
+package org.ods.services
+
+import org.ods.util.ILogger
+
+class AquaService {
```
michaelsauter, line 5: Please make this `@TypeChecked` :)

src/org/ods/services/AquaService.groovy
```diff
@@ -0,0 +1,64 @@
...
+            case 1:
+                script.error "An error occurred in processing the scan request " +
+                    "(e.g. invalid command line options, image not pulled, operational error)."
+                break
+            case 4:
+                script.error "The image scanned failed at least one of the Image Assurance Policies specified."
+                break
+            default:
+                script.error "An unknown status code was returned!"
+        }
```
michaelsauter, line 49: I'm wondering if we shouldn't move the responsibility to stop the pipeline to the stage? That would make the service easier to use in case you want to handle failure differently.

src/org/ods/services/BitbucketService.groovy
```diff
@@ -309,6 +309,50 @@ class BitbucketService {
...
+                script.sh(
+                    label: 'Create Bitbucket Code Insight report via API',
+                    script: """curl \\
+                        --fail \\
+                        -sS \\
+                        --request PUT \\
+                        --header \"Authorization: Bearer ${token}\" \\
+                        --header \"Content-Type: application/json\" \\
+                        --data '${payload}' \\
+                        ${bitbucketUrl}/rest/insights/latest/projects/${project}/\
```
michaelsauter, line 346: Is `latest` the right value? The other API calls in this service target `1.0`?

michaelsauter (commented):
Missing some tests as well :)

src/org/ods/services/BitbucketService.groovy
renedupont: Yes, the url in the doc https://docs.atlassian.com/bitbucket-server/rest/7.11.2/bitbucket-code-insights-rest.html#idm45774552049392 did not work, it took me half a day to find the working url in some PR of another project on github that integrated that. super strange.

src/org/ods/services/AquaService.groovy
renedupont: moved it

src/org/ods/services/AquaService.groovy
renedupont: done

src/org/ods/services/BitbucketService.groovy
michaelsauter: Oh Bitbucket.

Maybe worth a comment in the source file to document why it is like this ...

renedupont: I just checked again, they fixed it meanwhile, I adjusted accordingly.

src/org/ods/component/ScanWithAquaStage.groovy
renedupont: currently we don't have any image assurance policy discussed AFAIK, but I changed it now in a way that it should FAIL once we have implemented policies.

src/org/ods/component/ScanWithAquaStage.groovy
renedupont: added: ...was specified **in the ConfigMap**,...
the ConfigMap is explained now in the groovy doc

src/org/ods/component/ScanWithAquaStage.groovy
renedupont: changed it also for _registry_ and _secretName_

src/org/ods/component/ScanWithAquaStage.groovy
renedupont: added a adoc.tmpl file for documentation

src/org/ods/services/BitbucketService.groovy
metmajer: @renedupont are you able to list which Bitbucket version the URL works with and with which version it doesn't? We use a specific version in the corporate context which might differ from the one used at the BI-X.

renedupont: I think the initial confusion I had was that they have a wrong url in this how-to https://developer.atlassian.com/server/bitbucket/how-tos/code-insights/#create but not in the API doc. I changed it now from latest to 1.0 and I also compared for our bitbucket versions (which differ only in patch versions) and the url did not change.

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
+The "Aqua Security Scan" stage scans an image that was previously build in that same pipeline run.
```
michaelsauter, line 1:
```suggestion
The "Aqua Security Scan" stage scans an image that was previously built in that same pipeline run.
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+
+As a result, a BitBucket Code Insight entry is added to the git commit (in BitBucket) that basically
```
michaelsauter, line 3:
```suggestion
As a result, a Bitbucket Code Insight entry is added to the git commit (in Bitbucket) that basically
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+contains a link to the scan result on the Aqua platform. The BitBucket Code Insight entry can be seen in a Pull Request.
```
michaelsauter, line 4:
```suggestion
contains a link to the scan result on the Aqua platform. The Bitbucket Code Insight entry can be seen in a pull request.
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+A Pull Request in BitBucket shows the Code Insight of the latest commit of the PR.
```
michaelsauter, line 5:
```suggestion
The pull request in Bitbucket shows the Code Insight of the latest commit of the PR.
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  name: aqua
+...
+data:
+  registry: <registry-name-in-aqua-platform>
+  secretName: <secret-name-of-aqua-user-credentials>
+  url: <aqua-platform-url>
+----
+
+. `registry`: Refers to a name for the image registry given in the aqua platform by an aqua platform admin.
```
michaelsauter, line 19:
```suggestion
. `registry`: Refers to a name for the image registry given in the Aqua platform by an Aqua platform admin.
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+. `secretName`: Name of the `Secret` that contains the credentials of the aqua platform user that is used for executing the scan command. That user needs to have scanner rights.
```
michaelsauter, line 20:
```suggestion
. `secretName`: Name of a `Secret` that contains the credentials of the Aqua platform user that is used for executing the scan. That user needs to have scanner rights.
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+. `url`: Base URL of the aqua platform.
```
michaelsauter, line 21:
```suggestion
. `url`: Base URL of the Aqua platform (including scheme).
```

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
```diff
@@ -0,0 +1,43 @@
...
+
+Now, to use the stage, add it in your `Jenkinsfile`, e.g. like this:
+----
+) { context ->
+    ...
+    odsComponentStageScanWithAqua(context)
```
michaelsauter, line 27: Maybe good to show that this should be after the build stage, and if there is this `orElse` thing, it should be inside?

michaelsauter (commented):
Sorry for all the nitpicking!

Looks great overall 👌

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
metmajer: @michaelsauter @renedupont as mentioned above, I am not confident that scanning with Aqua should be a stage in the component's Jenkinsfile. Why? It's too easy to overcome. Rene, can you share your thoughts on my feedback above?

src/org/ods/services/BitbucketService.groovy
michaelsauter: @metmajer As I understand Atlassian versioning, v1.0 means Bitbucket Server/DataCenter, and v2.0 means Bitbucket Cloud. All of shared lib targets v1.0 of the API.

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
michaelsauter: I think it has to be in the component pipeline if you want feedback on pull requests. If you additionally want scanning to happen in the orchestration pipeline, it would need to be implemented in addition I think, not instead of the component pipeline scanning. Could be done, but maybe does not need to be done in this PR?

metmajer: Absolutely, scanning must happen in the component pipeline. However, I would like to consider security as a feature of the platform, and not as a feature of a component. In other words, I would not make Aqua scanning implemented in a dedicated function within the pipeline, but as an add-on to the build stage: build -> scanWithAqua. The latter should only be invoked though if Aqua is enabled for the current installation. Ideally, Aqua can be enabled/disabled by a cluster admin and does not depend on a config change and ensuing re-installation of ODS. Possible reasons are: future license expiry, breaking changes in Aqua, etc. which shall not block all builds. This was part of a previous discussion with Rene.

renedupont: I was under the assumption that your consideration to run that stage mandatorily was only meant for projects that use the orchestration pipeline, and not a general thing for every user of ODS. That is why I did not see it as part of this PR. I think when enforcing this stage to run in every build is something you want to add, this dedicated function can be used in the orchestration pipeline to do so.

metmajer: Hmm.. no, I was not seeing it this way. Aqua scanning is the implementation of a long standing security requirement which is definitely not optional for developers, but an integral part of building a safe application platform. Conceptually, it shall only be disable-able at the cluster level.

gerardcl: hi all, sorry to jump into, but I felt there is a misunderstanding on what the expectations would be, so I am listing from my point of view what I think we should have/distinguish:

- ODS/EDP component pipeline (team SLC): to have a stage on the CICD that clearly can tell you/teams how secure/healthy the component is (so to get feedback from Aqua: make use of the reports, and then if needed and possible even get into the reports on the Aqua side to dig more) and depending on the phase of the project work over it as deeply as needed (we know that not always one will really need to get blocked at this point -> MVP's, explorations, tests branches, demos,...) and even depending on the criticallity of the project auto-enforce actions (like failing the pipeline) 

- Platform/Cluster: to ensure corporate policies and governance requirements, such as avoiding deployments of images that are not compliant (if the project is on PROD maybe only...), enforcing by blocking or auditing non-permitted processes/behaviours...depending on the type of project, criticallity, processes,...

metmajer: I don't think we deviate much. Ever since I've been involved in this, scans when building have never been considered optional though. Where do you see the issue with Aqua scanning every build?

michaelsauter: @metmajer So you are envisioning the Aqua scan as part of the build stage, not as a separate stage? I like that.

How to turn Aqua on/off at cluster level is less clear to me. So far we do something similar with the SonarQube edition by writing information into the Jenkins image. Therefore, to change the setting, a new image needs to be built AND rolled out to all users. It would be nice to flip a switch centrally, but I think right now we do not have such a thing. It would need to be a service which can be asked by the pipeline whether Aqua scanning should be performed. The only "central" service we have under control is the provisioning app, and I think that's probably the wrong place. Next to "central" information, the "local" information whether Aqua scanning should be on/off could be presence of the `aqua` ConfigMap (and/or a key within that).

@metmajer Did you have something specific in mind how that central information about Aqua can be exposed to the pipelines?

metmajer: An `aqua` ConfigMap seems like a good approach to me. Ideally, if it's located centrally in the `OpenDevStack` namespace. We might be out of luck with our current clusters though: [Is there a way to share a configMap in kubernetes between namespaces?
](https://stackoverflow.com/questions/55515594/is-there-a-way-to-share-a-configmap-in-kubernetes-between-namespaces/55516643).

michaelsauter: Ah right, I somehow thought we can't do that - guess I was thinking about services. Duh.

Accessing the content of a central `ConfigMap` should actually work with `oc`, provided that the `jenkins` serviceaccount has `view` permissions in the central namespace. And given https://github.com/opendevstack/ods-core/blob/master/ods-setup/setup-ods-project.sh#L56, this should be possible. I checked in a Jenkins pod on both OpenShift 4 and 3.11, and it works in both.

So in theory we can have a central `ConfigMap` acting as a switch for Aqua.

renedupont: Currently, the `aqua` ConfigMap is meant to be in the project cd namespace because it also contains the `secretName` of the aqua user credentials that are different for each project/team. If there is a better way of using it we can change.

metmajer: I am not convinced that enabling/disabling Aqua at a project level will increase the trust in the EDP/ODS Platform, especially if this can happen at the will of every project lead. A power switch at the cluster level is still beneficial.

michaelsauter: @renedupont The `url` would not be project specific though. Is `registry` a project specific value? Not entirely clear to me. 

The `secretName` should probably be something that can be changed on project level. What do you think of a layered approach? As in, `url`, `registry` and `secretName` can be set in a global config map, and the local config map can override those? The global config map would also contain a switch if Aqua is on or off (and this would not be overridable as far as I understand?). Does this sound good or too complicated?

renedupont: Hello @metmajer, considering that
- you already have a pretty good and precise vision on how to further shape this
- I understood that it soon needs to be also tested in your environment
- I will not have much time to spent on this topic for at least the next 2-3 sprints (~4-6wks)
I would suggest that you (or your team) takes it on from here to not cause any delays on finishing it the way you envision.
I pushed it already to a branch on the opendevstack repo: https://github.com/opendevstack/ods-jenkins-shared-library/tree/feature/aqua-stage-master
Please let me know if this works for you?

metmajer: @renedupont I would love to, but my team is blocked with other duties needed for the upcoming release as well. We were informed that you want to own the integration with Aqua. In this case, please inform security about your situation.

metmajer: @renedupont forgot to add: Then, let's see how we can continue and how your sprints go. Maybe an opportunity opens up our end, but I don't see that in the coming weeks.

metmajer:
We depend on https://github.com/opendevstack/ods-core/pull/976 to be merged @segator

FYI @jafarre-viewnext

metmajer:
@renedupont I am unsure what to provide as a value for `registry`. I'd strive for a parameter-less solution that just works. Why should we request a project team to enter a registry value? Sorry, it's not clear to me.

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
metmajer: @michaelsauter @renedupont I added a question regarding the `registry` below. Agree that `secret` is right at the project level and needs to point to the x2 ods user. Regarding the additional `enabled` entry, I would initialise it to `false` at the cluster level and to `true` at the project level. Only if both (`&&`ed) evaluate to true, an Aqua scan will happen. Therefore, yes, the cluster-level means an override.

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -0,0 +1,138 @@
...
+                logger.info "The image scanned failed at least one of the Image Assurance Policies specified."
+                break
+            default:
+                steps.error "An unknown return code was returned: ${returnCode}"
+        }
+        logger.infoClocked(options.resourceName, "Aqua scan (via CLI)")
+        return returnCode
+    }
+
+    private createBitbucketCodeInsightReport(String aquaUrl, String registry, String imageRef, int returnCode) {
```
metmajer, line 105: @renedupont can you mention known report limitations (text length?) in a comment?

segator:
> We depend on [opendevstack/ods-core#976](https://github.com/opendevstack/ods-core/pull/976) to be merged into `master` by @segator
> The PR into `3.x` [opendevstack/ods-core#971](https://github.com/opendevstack/ods-core/pull/971) was already merged
> 
> FYI @jafarre-viewnext

Martin in the PR you mentioned this feature is only for 4.x so I closed the PR, shall I open back and apply change to 3.x?

metmajer:
@segator you are right, let me rephrase: 3.x was closed. the merge into master for 4.x is still pending. we need this PR so we can work on finishing the integration

docs/modules/jenkins-shared-library/partials/odsComponentStageScanWithAqua.adoc.tmpl
metmajer: @renedupont @michaelsauter I was also thinking of centralising the `url` property from the project-level ConfigMap inside the cluster-wide ConfigMap.

michaelsauter: > Regarding the additional enabled entry, I would initialise it to false at the cluster level and to true at the project level. Only if both (&&ed) evaluate to true, an Aqua scan will happen. Therefore, yes, the cluster-level means an override.

Agree!

> I was also thinking of centralising the url property from the project-level ConfigMap inside the cluster-wide ConfigMap

Would work for me.

metmajer: @michaelsauter @renedupont let's see if we can clarify the remaining question about the registry. I am not sure what this value indicates and if it should be at the project level or centralised. Is it something we can determine at ods configuration time? Ideally, we can make it a configuration parameter the users at the project-level don't have to think about.

michaelsauter: The docs say:

> `registry`: Refers to a name for the image registry given in the Aqua platform by an Aqua platform admin.

Not sure if this is a central registry? To me it sounds like one in which case central CM would work.

metmajer: @michaelsauter yes, I have read the docs, but assumed there was a specific reason for Rene to have the parameter defined at the project level. Not sure if Aqua has separate registries per project and if a cluster-wide central value would be sufficient? @maksymmaksym

renedupont: Hello! As there was no differentiation in the implemenation between project and global level yet before, every variable that is needed for the scannercli was in the (project-level) ConfigMap. As discussed above, in case there is only one aqua url and and one registry, it makes sense to already set it in the 'global' ConfigMap.

metmajer: Hi @renedupont ! Yes, we already clarified internally. Cheers!

metmajer:
Thanks @renedupont and @michaelsauter. Closing for being replaced by #661.

Issue #618, opened 2021-03-29
SonarQube edition name in ods Jenkins Shared library is diferent from ods core


braisvq1996:
**Describe the bug**
The environment value for the SonarQube edition in ods Jenkins Shared library is different from the one used for the instalation in ods core

**To Reproduce**
Check the code in ods core: https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/build/bc.yml#L110
Check the code in ods jenkins shared library: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/Context.groovy#L60

**Expected behavior**
Being able to get the Edition of SonarQube installed 

**Screenshots**
Jenkins master environment values:
![image](https://user-images.githubusercontent.com/26645694/112627647-2aa69000-8e32-11eb-9f5b-01fd399b1600.png)

Jenkinsfile:
![image](https://user-images.githubusercontent.com/26645694/112627011-537a5580-8e31-11eb-8261-707076f7a3c6.png)

Pipeline log:
![image](https://user-images.githubusercontent.com/26645694/112627126-77d63200-8e31-11eb-920e-4b215e83473c.png)

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack master

Refers to: https://github.com/opendevstack/ods-core/pull/967 & https://github.com/opendevstack/ods-core/issues/966

PR #619, opened 2021-03-29
Bugfix/sonar environment name

braisvq1996:
Refers to #618

@michaelsauter


michaelsauter (approved):
Thanks!

Should this be back ported to 3.x?

PR #620, opened 2021-04-13
Return blank with empty content in Document Chapter

hrcornejo:
When a Document Chapter has a predecessor and the new Document Chapter is empty the precesor was shown (and this is an error).


metmajer (approved):
Please provide the same fix into master, thanks!

PR #623, opened 2021-04-13
Return blank with empty content in Document Chapter

hrcornejo:
Same commit: https://github.com/opendevstack/ods-jenkins-shared-library/pull/620


metmajer (approved):
👍

PR #624, opened 2021-04-13
Fix description in tests

hrcornejo:
A "null" string is added in the Section 3.1 of the DTR when the Description of the Zephyr Unit Test doesn't has a Description


metmajer (commented):
@hrcornejo there seems to be a misunderstanding :( the old code is superior to the new code since most test cases only define a summary (title), but not a description text. With the new code, 99% of the test documents would show N/A. Please, cancel the PR and apply the old code to the DTP/DTR where we currently show N/A. Thanks!

hrcornejo:
@metmajer modified behaviour to DTR and DTP

s2oBCN (approved):
The LeVADocumentUseCaseSpec."create DTR" is not verifying anything about the document created. 
As we have a task to improve the document creation, I approve the PR

PR #640, opened 2021-05-04
component pipeline: repo name dynamic - master

clemensutschig:
@michaelsauter  merge?


michaelsauter (commented):
I think this looks right now. 

The branch has conflicts though so needs an update (could update the changelog at the same time then to have it tracked there - right now this "feature" would not work yet, right?)

clemensutschig:
@michaelsauter  it should, why not .. re changelog, this is "just inner workings" .. so I don't think this needs one, but ok either way - let me know

clemensutschig:
ping @michaelsauter

michaelsauter (approved):

PR #646, opened 2021-05-12
Add labels to generated OpenShift resources

jafarre-viewnext:
Closes https://github.com/opendevstack/ods-quickstarters/issues/512


src/org/ods/openshift/OpenShiftResourceMetadata.groovy
```diff
@@ -0,0 +1,462 @@
...
+ * Utility class to handle recommended and custom labels and annotations for OpenShift resources.
+ *
+ * @See <a href="https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/" >
+ *     Kubernetes: Recommended Labels</a>
+ * @See <a href="https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc" >
+ *     Guidelines for Labels and Annotations for OpenShift applications</a>
+ * @See <a href="https://helm.sh/docs/chart_best_practices/labels/" > Helm: Labels and Annotations</a>
+ *
+ */
+class OpenShiftResourceMetadata {
```
michaelsauter, line 18: Please add `@TypeChecked`, that is really helpful.

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
```diff
@@ -0,0 +1,462 @@
...
+        project:        'app.opendevstack.org/project',
+        projectVersion: 'app.opendevstack.org/project-version',
+        workInProgress: 'app.opendevstack.org/work-in-progress',
+    ]
+
+    /**
+     * <code>Map</code> with the mapping between label id's and metadata id's.
+     * Keys are label id's and values, metadata id's.
+     */
+    private static final mappings = [
```
michaelsauter, line 57: maybe name it `labelToMetadataMapping`?

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
```diff
@@ -0,0 +1,462 @@
...
+            def chartPath = "${config.chartDir}/Chart.yaml"
+            if (steps.fileExists(chartPath)) {
+                def chart = steps.readYaml(file: chartPath)
+                // The following replacement is as per the specification of this label
+                // and independent of the way we choose to sanitize labels otherwise.
+                return "${chart.name}-${chart.version.replace('+' as char, '_' as char)}"
+            }
+        }
+        return null
+    }
```
michaelsauter, line 429: Why do we read from chart here? why not let that be handled by Helm itself?

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -611,6 +611,61 @@ class OpenShiftService {
...
+        }
+        def labelStr = labels.collect { key, value ->
+            def label = key
+            label += value == null ? '-' : "='${value}'"
+            return label
+        }.join(' ')
+        if (logger.getDebugMode()) {
+            logger.debug("Setting labels ${labelStr} to resources ${resources} selected by ${selector}")
+        }
+        def script = "oc label --overwrite ${resources} "
```
michaelsauter, line 650: This might trigger a redeploy, right?

michaelsauter (commented):
Just glanced over it. Nice work.

I think in general the relationship between tools like Helm and this thing here is not entirely clear to me. Plus labelling causes a redeploy if I'm not mistaken which is not ideal I think?

michaelsauter:
In general, I am a bit uneasy with the perceived complexity this introduces. Here's where I'm coming from:

Mid-term, I think we should migrate to Helm templating. Together with this, I think we should / need to cut the automatic export functionality (there is no sane way to do it IMHO). As a consequence everyone would need to define their K8S resources in version control. I strongly believe this is the right way to go.
However with this on the horizon, I wonder how that would interact with the feature proposed here. Maybe there is a way to integrate labelling in a different way, e.g. through Helm? Or maybe only check for labels, instead of applying them in a separate process?

Just thinking out loud here ...

metmajer:
@michaelsauter I am not clear yet in which ways Helm would change the rules of the game compared to Quickstarters. Independent of this, components should always apply a set of labels we deem necessary, and the Release Manager would add some extra ones on top. As long as `app.kubernetes.io` labels are applied in a Helm-conform way, we should be future-proof, no?

michaelsauter:
@metmajer I think what needs to happen is that quickstarters ship with a Helm chart. Meaning we would keep the Helm templates under version control right from the start. We could by default set all the labels as desired by the platform. Then later on only a check is needed?

src/org/ods/services/OpenShiftService.groovy
jafarre-viewnext: In all tests I have done, labeling does not trigger redeploys. Otherwise, it is just a question of temporarily disabling automatic rollouts, but it doesn't seem to be needed at all.

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
jafarre-viewnext: I considered the possibility to integrate with the helm tool and even get several label values directly it, but I left that as out of scope (actually, setting this helm-related chart is an extra).
This was so easy to do, that it felt more straightforward just to read the chart file as a YAML. It can be improved in the future in quite a tranparent way, I would say.
If you don't feel confortable with my approach, I can disable this label for the moment or see how long it would take to integrate with the helm tool.

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
jafarre-viewnext: No problem

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
jafarre-viewnext: Will do.

jafarre-viewnext:
> In general, I am a bit uneasy with the perceived complexity this introduces. Here's where I'm coming from:
> 
> Mid-term, I think we should migrate to Helm templating. Together with this, I think we should / need to cut the automatic export functionality (there is no sane way to do it IMHO). As a consequence everyone would need to define their K8S resources in version control. I strongly believe this is the right way to go.
> However with this on the horizon, I wonder how that would interact with the feature proposed here. Maybe there is a way to integrate labelling in a different way, e.g. through Helm? Or maybe only check for labels, instead of applying them in a separate process?
> 
> Just thinking out loud here ...

Helm suggests how to set the recommended labels in the templates. It can be done that way in the future, or the shared library could use the helm tool to get the suggested template variable labels and apply them like it's being done in this PR.
Do you foresee any possible backwards-compatibility issues?

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
jafarre-viewnext: Done

src/org/ods/openshift/OpenShiftResourceMetadata.groovy
jafarre-viewnext: Done

jafarre-viewnext:
@michaelsauter ,@metmajer , when you have time, please, I need approval for the latest changes. Thank you.

metmajer (approved):

PR #664, opened 2021-06-09
Change the build status to Unstable when Jira Issues not DONE and deploying to D

jorge-romero:
Change to unstable the result of the build in case of documents not closed


vars/odsOrchestrationPipeline.groovy
```diff
@@ -82,7 +84,17 @@ def call(Map config) {
...
+                def cannotContinueAsHasOpenIssuesInClosingRelease = false
+                try {
+                    result = new InitStage(this, project, repos, startAgentStage).execute()
+                } catch (OpenIssuesException ex) {
+                    cannotContinueAsHasOpenIssuesInClosingRelease = true
+                }
+                if (cannotContinueAsHasOpenIssuesInClosingRelease) {
+                    logger.warn('Cannot continue as it has open issues in the release.')
+                    return
+                }
```
metmajer, line 97: @jorge-romero why rather complex? wouldn't the logger.warn + return inside the catch block be enough?

src/org/ods/orchestration/util/Project.groovy
```diff
@@ -362,20 +363,10 @@ class Project {
...
-                'since the following issues are work in progress: ':
-                "The pipeline failed since the following issues are work in progress (no documents were generated): "
-
-            this.getWipJiraIssues().each { type, keys ->
-                def values = keys instanceof Map ? keys.values().flatten() : keys
-                if (!values.isEmpty()) {
-                    message += '\n\n' + type.capitalize() + ': ' + values.join(', ')
-                }
-            }
+            String message = generateWIPIssuesMessage()
```
angelmp01, line 366: If you use only the message inside the if, create it inside

src/org/ods/orchestration/util/Project.groovy
```diff
@@ -406,6 +397,21 @@ class Project {
         return this
     }
 
+    def generateWIPIssuesMessage() {
```
angelmp01, line 400: Maybe we can move the responsability of generate this message to another class instead Project class. Think that project class is too big.

vars/odsOrchestrationPipeline.groovy
jorge-romero: That's was my first implementation, but it was complaining about trying to finalize the pipeline within a catch block, that's why I had to rewrite in this ugly way unfortunally

src/org/ods/orchestration/util/Project.groovy
jorge-romero: If you check the whole code as in the diff looks different it is inside the if block :)

src/org/ods/orchestration/util/Project.groovy
jorge-romero: Moved to another class, and moved the tests for that specific functionallity to a new Spec for that class

vars/odsOrchestrationPipeline.groovy
angelmp01: Another solution may be logger.warn + result=false inside the catch block. And remove cannotContinueAsHasOpenIssuesInClosingRelease var.

src/org/ods/orchestration/util/Project.groovy
```diff
@@ -406,6 +397,21 @@ class Project {
...
+            'since the following issues are work in progress: ' :
+            "The pipeline failed since the following issues are work in progress (no documents were generated): "
+
+        this.getWipJiraIssues().each { type, keys ->
+            def values = keys instanceof Map ? keys.values().flatten() : keys
+            if (!values.isEmpty()) {
+                message += '\n\n' + type.capitalize() + ': ' + values.join(', ')
+            }
+        }
+        message
```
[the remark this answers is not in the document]
jorge-romero: It is done in the new class afte refactoring that we have created to deal with the message  as @angelmp01 requested

angelmp01 (approved):

PR #669, opened 2021-06-16
Fix epic Issues not correctly ordered on the CSD

jafarre-viewnext:


metmajer (approved):
@jafarre-viewnext nice work! Please also provide a fix for master.

PR #680, opened 2021-06-23
Fix reference Document Version for the DTR and TIR are not correct

jafarre-viewnext:


src/org/ods/orchestration/util/Project.groovy
```diff
@@ -780,7 +780,7 @@ class Project {
 
     @NonCPS
     Map<String, DocumentHistory> getDocumentHistories() {
-        return this.data.documentHistories ?: [:]
+        return this.data.documentHistories
```
metmajer, line 783: Is this.data.documentHistories properly initialised?

src/org/ods/orchestration/util/Project.groovy
```diff
@@ -980,19 +980,22 @@ class Project {
...
     }
 
     @NonCPS
     DocumentHistory findHistoryForDocumentType(String documentType) {
-        // All docHistories for DTR and TIR should have the same version
-        def key = this.data.documentHistories.keySet().find { it.startsWith(documentType) }
-        return this.getHistoryForDocument(key)
+        def history = getHistoryForDocument(documentType)
+        if (!history) {
+            // All docHistories for DTR and TIR should have the same version
```
metmajer, line 990: I find the comment a bit confusing since the implementation of the method nowhere distinguishes between document types. Is this not a general principle?

src/org/ods/orchestration/util/Project.groovy
jafarre-viewnext: It is initialized to an empty map in the constructor and nobody touches it anywhere else.

src/org/ods/orchestration/util/Project.groovy
jafarre-viewnext: You are right, @metmajer , this method is itself misleading. There is more than one document history for each of these two document types, so it doesn't seem justified to just pick the first one found. The thing is that this method is actually used just to get the document version. As the comment says, the version should be the same in all document histories for the same document type, so it doesn't really matter.
I didn't feel like refactoring further in the scope of a bug fix, but it would be better to property turn this method into one that gets precisely the document version. Would you like me to do that?

metmajer: @jafarre-viewnext yes, please!

jafarre-viewnext: Done

metmajer (approved):

PR #681, opened 2021-06-24
Fix reference Document Version for the DTR and TIR are not correct (master)

jafarre-viewnext:


metmajer (approved):
@jafarre-viewnext  top!

PR #686, opened 2021-06-30
Add recommended labels to pods

jafarre-viewnext:
This pull request provides the functionality to update the templates in Deployment and DeploymentConfig resources, so that pods have the recommended labels automatically applied whenever they are created.


metmajer (approved):
Top job @jafarre-viewnext !

PR #689, opened 2021-07-13
mono-repo promote rollout verification - 3.x

### LINKED ISSUE #688: MRO fails during promotion to P/Q with mono-repo and 2 images
**Describe the bug**
Mono-repo with 2 images (and image triggers to :latest) fails during promote to Q/P with 

```
+ oc -n gihkw1-prod get dc/front-back -o 'jsonpath={.status.latestVersion}'
[Pipeline] sh (Rollout latest version of dc/front-back)
+ oc -n gihkw1-prod rollout latest dc/front-back
error: #11 is already in progress (Running).
[Pipeline] sh (Get latest version of dc/front-back)
+ oc -n gihkw1-prod get dc/front-back -o 'jsonpath={.status.latestVersion}'
[Pipeline] }
[Pipeline] // dir
[Pipeline] echo
WARN: script returned exit code 1
[Pipeline] }
Failed in branch front-back
[Pipeline] // parallel
[Pipeline] }
Failed in branch Deploy
[Pipeline] // parallel
[Pipeline] }
[Pipeline] // node
[Pipeline] echo
WARN: Error occured within the orchestration pipeline: script returned exit code 1
[Pipeline] echo
[Deploy] **** ENDED orchestration stage **** (took 42280 ms)
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] echo
[ods-mro-pipeline] **** ENDED orchestration pipeline **** (took 95127 ms)
``` 

 - OpenShift: [3.11]
 - OpenDevStack [3.x]

clemensutschig:
closes  #688 for 3.x 

@michaelsauter, can you try to think this thru as well?

### DIFF
--- a/src/org/ods/services/OpenShiftService.groovy
+++ b/src/org/ods/services/OpenShiftService.groovy
@@ -182,6 +182,7 @@ class OpenShiftService {
         replicationController
     }
 
+    @SuppressWarnings('LineLength')
     void startRollout(String name, int version) {
         try {
             steps.sh(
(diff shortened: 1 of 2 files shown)


src/org/ods/services/OpenShiftService.groovy
```diff
@@ -193,8 +194,23 @@ class OpenShiftService {
...
+                // we "assume" the running on to be the right one - and bubble
+                // that one up - as surrounding code will check anyway for matching of
+                // image sha's etc...
+                def replicationController = "${name}-${newVersion}"
+                def rolloutStatus = getRolloutStatus(replicationController)
+                if (rolloutStatus != 'running') {
+                    logger.debug("Deployment #${version} / latest #${newVersion} has errored out: ERR ${ex}, status: ${rolloutStatus}")
+                    throw ex
+                } else {
+                    logger.debug("New Deployment #${newVersion} already running")
```
michaelsauter, line 212: were you able to test this against the issue? I also think that if the rollout is running, it is OK to continue. However, I have two open questions in my head:

1. could there be a case where a rollout is running for a previous version, and we are not picking it up? that would suggest we should cancel, and then restart?
2. I don't really understand how the error happens yet so not sure if this actually solves the issue ...

clemensutschig: > 1. could there be a case where a rollout is running for a previous version, and we are not picking it up? that would suggest we should cancel, and then restart?

as we verify anyway the image sha's later - I think that is fine - but indeed - that would be option two (yet complex) because another trigger could fire in between, and we have the same issue... :;D

michaelsauter (approved):

PR #697, opened 2021-07-29
Feature/pipeline spec2

s2oBCN:
### Add functional test to LevaDoc

- The new test is in the file: **LevaDocumentUseCasePipelineSpec**
- Then some refactors in order to use ILogger instead of the Logger implementation

As I will go on vacation I ignored the tests, but I think it's a good idea to merge the code to master, as with these tests we can debug the code and generate the pdfs from the test. And this could be helpful to solve the tickets related to pdfs.

Later I can do a workshop in order to take all the advantages of the framework

### You will see in the LevaDocumentUseCasePipelineSpec header how to play with the test:
`
/**
 * IMPORTANT: this test use Wiremock files to mock all the external interactions.
 *
 * We have 2 flags to play with the test:
 *  - RECORD: When TRUE wiremock will record the interaction with the servers and compare the pdf results with the expected
 *  - GENERATE_EXPECTED_PDF_FILES: When TRUE it will remove the expected pdfs and create a new ones
 *
 *  ie:
 *  - RECORD=false & GENERATE_EXPECTED_PDF_FILES=false are the default values. So then it can be executed everywhere.
 *  - RECORD=true & GENERATE_EXPECTED_PDF_FILES=false will record and compare the generate pdfs with the 'old' expected files
 *      ==> with this combination, if there's an error,
 *          we can compare new pdf with the old one, and see the implications of our changes in the pdfs
 *  - RECORD=true & GENERATE_EXPECTED_PDF_FILES=true will record and generate new pdf expected files
 *
 */
`


jafarre-viewnext:
All those many autogenerated files, is it possible to integrate the autogeneration with the Gradle build instead of storing them in the code repository? Or do they need to be manually customised?
Otherwise, with the tests ignored by default, the merge seems harmless.

metmajer:
@s2oBCN I agree with @jafarre-viewnext

metmajer:
Since a review cannot and should not be done while @s2oBCN is on vacation, we'll table this PR and follow up after the release of ODS 4.

s2oBCN:
> All those many autogenerated files, is it possible to integrate the autogeneration with the Gradle build instead of storing them in the code repository? Or do they need to be manually customised?
> Otherwise, with the tests ignored by default, the merge seems harmless.

No, we need to store all these generated files, because we store there a know state. So next time you change the code it will compare the pdfs generated by the new code with the old know state.

metmajer:
@s2oBCN there needs to be an alternative. How does the state get into these files? Why not store the state somewhere and generate the files on the fly?

s2oBCN:
> @s2oBCN there needs to be an alternative. How does the state get into these files? Why not store the state somewhere and generate the files on the fly?

These files are the mocks of the external interactions, with these files we can generate the pdfs without any external interaction  (here we are mocking a complex Jira project). 
- Without these files we can not execute the test as a component test.
- We will generate the files another time only if we change our pdfs, otherwise we don't need to change the files.

I know that so many files can scare, but it's a standard way of working as are stored in test/resources/wiremock/nameOfTheTest/....

metmajer (dismissed):
@s2oBCN I now recall that there is a recording mode built into the framework that creates the data through WireMock. Overall, a very smart approach. Let's do it this way 👍

jorge-romero (approved):

metmajer:
@s2oBCN and @jorge-romero this requires a merge into the 4.x branch.

PR #701, opened 2021-08-03
Fix wrong entry in the document history when a new component is added.

jafarre-viewnext:


src/org/ods/orchestration/util/DocumentHistory.groovy
```diff
@@ -227,6 +212,21 @@ class DocumentHistory {
         sortDocHistories(dhs).reverse()
     }
 
+    // Find the previous environment where the document was generated, if it exists.
+    // Otherwise, use the target environment.
+    @NonCPS
+    private getSourceEnvironment(String targetEnvironment, String documentName) {
+        def documentType = LeVADocumentUtil.getTypeFromName(documentName) as String
+        def environment = null
+        def envs = ['D', 'Q', 'P']
```
metmajer, line 221: @jafarre-viewnext should `envs` be defined in-line? Don't we have a global reference already?

metmajer (changes requested):

metmajer (approved):

metmajer:
@jafarre-viewnext this change actually needs to go into 4.x, not 3.x, no?

src/org/ods/orchestration/util/DocumentHistory.groovy
jafarre-viewnext: I extracted the environments into an Enum

jafarre-viewnext:
Some refactoring to make the code more maintainable has been performed

metmajer (approved):

metmajer:
@jafarre-viewnext no need to add fixes into 3.x. Do we have a corresponding PR into master? Consider that what has a PR into master can be git merged directly into 4.x.

Issue #663, opened 2021-06-09
Sonarqube stage crashes in cnes report


jssnull:
If we've an ods quickstarter, and we create a PR for it.
If we trigger a build in the PR origin branch, the jenkins build will crash in sonarqube stage, concretely, when the cnes report
is called. It will retrieve an error like this:

```bash
java -jar /usr/local/cnes/cnesreport.jar -s https://mysonarinstance -t ****** -p myproject-mycomponent -a jenkins-mycomponent-cd-myproject-cd-mycomponent-develop-1 -b develop
SonarQube URL: https://mysonarinstance
SonarQube online: true
Detected SonarQube version: 8.2.0
[ERROR] Not found error sent by SonarQube server (code 404, URL https://mysonarinstance/api/navigation/component?component=myproject-mycomponent&branch=develop, Error {"errors":[{"msg":"Component 'myproject-mycomponent' on branch 'develop' not found"}]}), please check cnesreport compatibility with your SonarQube server version.
```

After having this error, If we delete the created PR and launch a build in the same branch again, it will work.
Then, seems that the cnes report is not able to retrieve data from PR in sonarqube.  This happens in sonarqube developer and above editions.

Seems that in the past, situations like this happened in cnes plugin: https://github.com/cnescatlab/sonar-cnes-report/issues/159

metmajer:
@jssnull when you create an issue that breaks relevant behavior of our Platform, please ensure there's some traction on the topic. Especially if you are tackling this from a support perspective. Did you already include any people that can help you on that matter?

s2oBCN:
The problem is because cnes-report (SonarQu tool) is not compatible with PR scan. https://github.com/cnescatlab/sonar-cnes-report/issues/159



This is because the scan is done against the PR and not against the Branch. To solve this we have to change the Jenkinsfile in order to scan the 'Branch'.  With this code:


odsComponentStageScanWithSonar(context,[analyzePullRequests: false])


PS. We think that this should be the default behaviour: Not to scan PRs.

michaelsauter:
I had to think a bit more about this and concluded that next to the merged solution, there is also another option. So just putting it out there as an FYI ...

[...]

PR #731, opened 2021-10-04
by default analyzePullRequests should be false

s2oBCN:
From this isssue https://github.com/opendevstack/ods-jenkins-shared-library/issues/663
Instead of change all the 'quickstarters' is better to change the default behaviour.

Detailed explanation:
- SonarQu can generate a scan for a branch or a scan for a PR. 
- We can only ask SonarQu a report of a branch. See https://github.com/cnescatlab/sonar-cnes-report/issues/159

As we are always asking for the report of the branch, not for a report of a PR. Then, by default, we should generate a report for the branch, not for the PR.

### DIFF
--- a/src/org/ods/component/ScanWithSonarStage.groovy
+++ b/src/org/ods/component/ScanWithSonarStage.groovy
@@ -50,7 +50,7 @@ class ScanWithSonarStage extends Stage {
                 .toList()
         }
         if (!config.containsKey('analyzePullRequests')) {
-            config.analyzePullRequests = true
+            config.analyzePullRequests = false
         }
         if (!config.containsKey('requireQualityGatePass')) {
             config.requireQualityGatePass = false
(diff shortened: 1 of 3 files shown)


src/org/ods/component/ScanWithSonarStage.groovy
```diff
@@ -50,7 +50,7 @@ class ScanWithSonarStage extends Stage {
                 .toList()
         }
         if (!config.containsKey('analyzePullRequests')) {
-            config.analyzePullRequests = true
+            config.analyzePullRequests = false
```
s2oBCN, line 53: If true, it will generate a report for the PR, not for the Branch in SonarQu

michaelsauter (approved):
Changes look fine to me.

In the future it would be nice to get the PR analysis feature back as I believe that is the only way to really benefit from SQ as a developer. if you don't have PR integration it is just a pain ...

metmajer (approved):

PR #737, opened 2021-10-08
When two coded tests are linked to the same test issue the…

angelmp01:
… pipeline doesn't fails

- Updated matchTestIssuesAgainstTestResults method to thrown exception when found multiple test results with same testissue.key and added test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Throw exception when two coded tests are linked to the same test issue (https://github.com/opendevstack/ods-jenkins-shared-library/pull/737)
 - Add functional test to LevaDoc  ([#697](https://github.com/opendevstack/ods-jenkins-shared-library/pull/697))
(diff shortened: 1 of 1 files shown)


src/org/ods/orchestration/usecase/JiraUseCase.groovy
```diff
@@ -256,13 +256,20 @@ class JiraUseCase {
     }
 
     void matchTestIssuesAgainstTestResults(List testIssues, Map testResults, Closure matchedHandler, Closure unmatchedHandler = null) {
+        def duplicateKeysErrorMessage = "Error: found duplicated Jira tests. Check tests with key: "
```
metmajer, line 259: @angelmp01 @s2oBCN @jorge-romero this error message is not at all clear to the end-user. Can you please be more specific?

metmajer: @angelmp01 What about... "Error: the following test cases are implemented multiple times each: JIRA-1, JIRA-2." We should as well omit the [ ] in the output. Thanks.

PR #741, opened 2021-10-19
[EDPC-993] The Developer Preview fails because "Duplicated Tests"

angelmp01:
- Added param checkDuplicateTestResults in computeTestDiscrepancies to check or not the duplicate tests results
- Added param checkDuplicateTestResults in matchTestIssuesAgainstTestResults to check or not the duplicate tests results
- In the creation of the documents TRC and CFTR the call to computeTestDiscrepancies method is with the param matchTestIssuesAgainstTestResults=false
  to avoid to check duplications in test results


clemensutschig:
@angelmp01 - great, yet build and Changelog check fail

Issue #745, opened 2021-10-25
java.util.ArrayList$Itr error when using the orchestration pipeline


braisvq1996:
**Describe the bug**
We have a project that is using a huge amount of resources for theis Jenkins-master instance (20 GB and JVM is -Xms10g -Xmx18g) in order to aboid this error, but it still appear from time to time when running developer previews.
I have only seen this if **document generation is enabled**.

**To Reproduce**
1. Make use of the orchestration pipeline to deploy some components and enable document generation
2. Keep running developer previews without increasing Jenkins resources until you see this error

**Expected behavior**
The execution of the orchestration pipeline with document generation should not fail

**Screenshots**
Resources usage when error happened:
![captura](https://user-images.githubusercontent.com/26645694/138669469-d855228a-3209-4813-aef0-0a26f47e5b64.jpg)

Log where the error took place):
![captura2](https://user-images.githubusercontent.com/26645694/138670259-add16548-2596-4cd6-867c-4f366f507444.jpg)

Another log where the error happened:
![image](https://user-images.githubusercontent.com/26645694/138672119-3b89589a-8a53-4d92-be4f-3fa41a1a4af2.png)

**Affected version (please complete the following information):**
 - OpenShift: 3.11 and 4.7
 - OpenDevStack 3.x and 4.x

clemensutschig:
@braisvq1996 - as discussed please test a couple of more runs of this combo - so we really see if there is not more to fix. 
also, I would add stories and tests etc ... so we get more traction now on the levadoc case - and see what we have missed there. 

@metmajer - fyi

clemensutschig:
we are stuck with the component pipeline (we get the $Itr - seriously at the script.sh!) @michaelsauter any ideas?

michaelsauter:
I looked at the source code right now but I do not understand how that line could produce an exception that `java.util.ArrayList$Itr` is not serialisable. Which list is used there? Are you sure the exception is from that line - and that there isn't an issue in reporting the exception?

clemensutschig:
so - with a while & sleep it seems to work ... :(

Tmrw we'll stick even more stories++ into jira to see if there is more we missed ...

braisvq1996:
with the changes on https://github.com/opendevstack/ods-jenkins-shared-library/pull/980 and https://github.com/opendevstack/ods-core/pull/1217 we do not have more memory issues with Jenkins so issue can be closed

PR #750, opened 2021-10-29
Fix ArrayList error in Orchestration pipeline

braisvq1996:
Fix for #745

### DIFF
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -528,7 +528,8 @@ class Context implements IContext {
             }
             logger.debug("Normalizing ENV")
             return normalizeEnvironment(rawEnv, keyPrefix)
-        }catch(err){
+        }catch(java.io.NotSerializableException err){
+            logger.debug(err)
             return [:]
         }
     }
--- a/src/org/ods/component/Context.groovy
+++ b/src/org/ods/component/Context.groovy
@@ -59,17 +59,17 @@ class Context implements IContext {
 
         config.sonarQubeEdition = script.env.SONAR_EDITION ?: 'community'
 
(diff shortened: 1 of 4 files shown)


metmajer:
@clemensutschig @braisvq1996 @s2oBCN @jorge-romero just a thought, guys... what if we create `Util.findAll` etc. methods with a "safe" implementation? This would improve code readability and would not require vast changes to existing codebase.

metmajer:
@braisvq1996 can you please fix the CodeNarc errors? Which are the results from your tests? Do the fixes show good effects?

braisvq1996:
@metmajer , with this changes the serialization issue does not seems to appear anymore, bue we do get ``Hit Jenkins serialization issue, attempt X`` in the Jenkins pipeline log.
I will fix the CodeNarc errors as soon as I can

clemensutschig:
@metmajer seems its working fine :) I also took a look at another project which shows similar errors - and the code we have put in place now should fix those (Even on 3.x) @s2oBCN

metmajer:
Very cool @braisvq1996 and @clemensutschig! I would fix the CodeNarc issues myself, but I don't have access to Brais' branch.

metmajer:
Colleagues, until Brais' branch is merged into ODS' fix branch, I want to document the approach here. The benefits of using utility methods is that...
a) these methods can be easily Unit tested. Unit tests would need to prove that, basically `java.util.Collection.findAll(...) == `Util.findAll(collection,...)`.
b) Changes to the codebase remain surgical, since methods from Java can be easily replaced with our utility methods. This helps in code reviews to assert that changes do not introduce failure.

clemensutschig:
super nice @metmajer  - the only thing we need to do is to remove the NonCPS annotations - because if you take a look, we use plenty of script..... in there

PR #751, opened 2021-11-02
Fix/remove from tip non intalled components

jorge-romero:


clemensutschig (approved):

metmajer:
@jorge-romero please get back to documenting what the fix intends to do with every PR.

PR #760, opened 2021-11-10
Fix/non breakable spaces

s2oBCN:
fix non breakable spaces in reports by removing special caracters


src/org/ods/orchestration/service/JiraService.groovy
```diff
@@ -260,7 +261,7 @@ class JiraService {
             throw new RuntimeException(message)
         }
 
-        return new JsonSlurperClassic().parseText(response.getBody())
+        return new JsonSlurperClassic().parseText(StringCleanup.removeCharacters(response.getBody()))
```
metmajer, line 264: @s2oBCN you should reconsider this interface. The characters to be removed depend on the scope. In a JiraService, we want to remove characters we don't want from Jira. But the StringCleanup class holds these characters. I suggest to refactor to `StringUtil.removeCharacters(text, characters)` and let the JiraService store the unwanted characters. Better coupling IMHO.

metmajer (changes requested):

src/org/ods/orchestration/service/JiraService.groovy
s2oBCN: The idea here is that it doesn't depends on JiraService, depends on what we don't want in our pdfs, so if tomorrow we have another interface we will use the same. Basilly,  decouple code.
And by the other hand don't add more code to JiraService

metmajer: @s2oBCN but the character '\u00A0' comes from Jira and is cleaned up in the JiraService. Hard-coding this into a generic StringCleanup class is not clean.

jorge-romero: @s2oBCN, @metmajer  The StringUtil.removeCharacter has been refactor to include the characters we want to remove and moved the characters Map (valueToRemove: NewValue f.e. nonbreakablespace: space) to the JiraService, so the class is reusable and the characters belongs to the calling instance, in this case to the JiraService.
Sorry for the number of commits as there was a misundertanding of the comments, with the squash they will dissapear

metmajer (approved):

PR #764, opened 2021-11-11
Fix/revert tip component filter (#759)

jorge-romero:
* Undo and comment changes for TIP removed components

* Fix CHANGELOG.md


jorge-romero:
Duplicated

PR #755, opened 2021-11-05
Bugfix/edpp to3x

s2oBCN:
Solver SSDS error and SerializationException

PR #769, opened 2021-11-12
Fix/null in ra doc - 3.x

s2oBCN:
Solve same problem to RA as done here ([#755](https://github.com/opendevstack/ods-jenkins-shared-library/pull/755))


clemensutschig:
this is from the wrong basis - or needs to be a cherry pick ...

s2oBCN:
> this is from the wrong basis - or needs to be a cherry pick ... I changed it now to 3.x - which is what it should be, or? @s2oBCN

We did this for the 4.x, but we can have the same problem here. So,  a cherrypick is not possible.

metmajer (approved):

PR #756, opened 2021-11-05
Fix ArrayList error in Orchestration pipeline - 4.x

metmajer:
- Fix SerializationException
- Fix NULL in SSDS generation
- Fix SaaS bug in FinalizeStage

PR #770, opened 2021-11-12
Fix ArrayList error in Orchestration pipeline - 4.x (#756)

s2oBCN:
Fix SerializationException
Fix NULL in SSDS generation
Fix SaaS bug in FinalizeStage


metmajer (approved):
@s2oBCN I did not check every detail, but if its the same that @clemensutschig approved for 4.x, please go ahead

PR #777, opened 2021-11-12
fix: reporting 3 bugs instead of 1 on deploy to Q in case there is 1 bug - master

petod108:
fix problem with test reporting and added more logs in order to understand better the process

Fixing bug:
When executing step "deploying on QA"
And there is one bug in test project-
Then after execution in Jira exist three same bugs
and from those one one is linked with the test Jira ticket.
which was produced by keeping same data structure for every type of test

added also better logging
and reducing size of function because the same code is repeated for every type of test (Acceptance, Installation, Integration)

### DIFF
--- a/src/org/ods/orchestration/TestStage.groovy
+++ b/src/org/ods/orchestration/TestStage.groovy
@@ -24,22 +24,8 @@ class TestStage extends Stage {
         def junit = ServiceRegistry.instance.get(JUnitTestReportsUseCase)
         def levaDocScheduler = ServiceRegistry.instance.get(LeVADocumentScheduler)
         def util = ServiceRegistry.instance.get(MROPipelineUtil)
-
         def phase = MROPipelineUtil.PipelinePhases.TEST
-
-        def globalStructure = [
-            testReportFiles: [],
-            testResults: [
-                testsuites: []
-            ],
-        ]
-        def globalData = [
-            tests: [
-                acceptance: globalStructure,
-                installation: globalStructure,
-                integration: globalStructure,
-            ]
-        ]
+        def globalData = getTestDataStructure()
 
         def preExecuteRepo = { steps_, repo ->
             levaDocScheduler.run(phase, MROPipelineUtil.PipelinePhaseLifecycleStage.PRE_EXECUTE_REPO, repo)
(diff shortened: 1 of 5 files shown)


src/org/ods/orchestration/TestStage.groovy
```diff
@@ -24,50 +24,32 @@ class TestStage extends Stage {
...
         def phase = MROPipelineUtil.PipelinePhases.TEST
-
-        def globalStructure = [
-            testReportFiles: [],
-            testResults: [
-                testsuites: []
-            ],
-        ]
-        def globalData = [
-            tests: [
```
petod108, line 37: This code was reason for bug behavior

metmajer:
@petod108 @s2oBCN @jorge-romero please consider that, currently, 4.x is not released. The Changelog shows a release date of 2021-05-11 which is inaccurate. Master should show the release in the changelog once released, until then additions are unreleased

src/org/ods/orchestration/TestStage.groovy
```diff
@@ -85,34 +67,32 @@ class TestStage extends Stage {
...
-        globalData.tests.acceptance.testResults = junit.parseTestReportFiles(
-            globalData.tests.acceptance.testReportFiles
-        )
-        globalData.tests.installation.testResults = junit.parseTestReportFiles(
-            globalData.tests.installation.testReportFiles
-        )
-        globalData.tests.integration.testResults = junit.parseTestReportFiles(
-            globalData.tests.integration.testReportFiles
-        )
+        globalData.tests.each {
```
clemensutschig, line 70: was this tested with marketplace etc ... this looks like making the $Itr error coming up again .. @s2oBCN @metmajer

s2oBCN: No, these changes were tested in a Box

metmajer: @s2oBCN please make sure the PR does not accidentally revert any NonSerializableException-related fixes. Once you checked, let's remove the Draft mode again. This is just to avoid last-minute errors creeping in again. Thanks!

clemensutschig: @s2oBCN - we fixed a bunch of $Itr issues in the teststage - and all this code is not in here right now ... ´largely via: 
https://github.com/opendevstack/ods-jenkins-shared-library/pull/770/files 

All that needs to go into master as well and merged with this one here

petod108: @clemensutschig @metmajer 
I tested this with E2E tests, code has no serialization error.
For this kind of iteration there is Jenkins support https://issues.jenkins.io/browse/JENKINS-50997
Only what is important is to know that iterator is **Immutable**

PR #783, opened 2021-11-18
Increase Socket Timeout on Pipeline Orchestrator 4x

segator:
When we are generating very big Reports With Docgen sometimes we get socket timeout


clemensutschig:
@segator  - changelog

clemensutschig (approved):

PR #789, opened 2021-11-25
Fix image not shown in document chapters

jorge-romero:
We introduce the jira image conversion from for some specific case where the image document chapters were retrieved in a different way or from the .yaml that exists in the folder docs of the Release Manager


s2oBCN (approved):

clemensutschig:
@jorge-romero this must be forced ported to master! @metmajer

PR #791, opened 2021-11-29
fix: reporting 3 bugs instead of 1 on deploy to Q in case there is 1 …

petod108:
…bug (#778)

Co-authored-by: Peter Durisin <durisin.peter.ext@boehringer-ingelheim.com>


petod108:
threre is big PR of all changes in 4.0

PR #792, opened 2021-11-30
4x to master

s2oBCN:
After the deployment in PROD merge changes to master


src/org/ods/services/OpenShiftService.groovy
```diff
@@ -804,7 +804,7 @@ class OpenShiftService {
      */
     String patch(String resource, Map<String, ?> patch, String path = null, String project = null) {
         if (!resource) {
-            throw new IllegalArgumentException('You must specify the resource to patch')
+            throw new IllegalArgumentException('You must specify the resource to path')
```
metmajer, line 807: @s2oBCN  this seems wrong, should be patch, not path

s2oBCN: It was a draft... now is ready ;)

jafarre-viewnext (approved):

PR #827, opened 2022-01-19
Better handling of Unit tests in D and DevPreview

hrcornejo:
- The execution status of all Unit tests, for all components, are reported into Jira.
- The DTR document is generated and contains information that is consistent with Jira.
- When doing a Developer Preview, the Release Manager pipeline fails only at the end of the Finalize Stage.
- When deploying to DEV, the Release Manager pipeline fails at the end of the Build Stage.
- The Release Manager reports the the failure status into its Release Status issue.


metmajer:
It would be nice, but I am not sure if this implementation will make the requirements. Consider that a component might decide to fail the build after the first failing unit tests. Currently, components are executed in groups with a `failFast` property. Therefore, after one failing component, the release manager pipeline would fail.

hrcornejo:
> I see many many files in which a generic value is changed by sergio.sacristan%40domain.com
> 
> Is it on purpose?
@victorpablosceruelo there are only 3 files changed between this branch and master.

hrcornejo:
> It would be nice, but I am not sure if this implementation will make the requirements. Consider that a component might decide to fail the build after the first failing unit tests. Currently, components are executed in groups with a `failFast` property. Therefore, after one failing component, the release manager pipeline would fail.

@metmajer added failFast dependent of WIP.
With 2 projects in parallel with this configuration:
image
Launching a WIP all the stages are executed, doc generated and issues updated in Jenkins (with failing tests in one component):
image
Launching a 1.1 version to DEV, the pipeline fails after Build:
image

metmajer:
@hrcornejo fantastic! Interestingly, in the second example, the pipeline fails, but we don't see that the component failed. I assume this is because the component itself does not throw an exception/error on test failure?

test/groovy/org/ods/orchestration/BuildStageSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+    def setup() {
+        script = new PipelineScript()
+        steps = Mock(PipelineSteps)
+        levaDocScheduler = Mock(LeVADocumentScheduler)
+        project = Spy(createProject())
+        util = Mock(MROPipelineUtil)
+        jira = Mock(JiraUseCase)
+        junit = Mock(JUnitTestReportsUseCase)
+        jenkins = Mock(JenkinsService)
+        logger = new Logger(script, !!script.env.DEBUG)
```
s2oBCN, line 42: This is a test, why initialize with !!script.env.DEBUG?

s2oBCN (dismissed):

test/groovy/org/ods/orchestration/BuildStageSpec.groovy
hrcornejo: C&P from other test. Updated in new commit

s2oBCN (approved):

PR #843, opened 2022-02-14
Fix the SCR table in SSDS (chapter 2.3) appears cut off

jafarre-viewnext:
Process pull-request data directly in memory instead of generating an intermediate CSV file


src/org/ods/orchestration/usecase/BitbucketTraceabilityUseCase.groovy
```diff
@@ -37,109 +32,41 @@ class BitbucketTraceabilityUseCase {
...
-            file.text,
-            Record.CSV,
-            ['commitDate', 'author', 'reviewers', 'pullRequestUrl', 'commitSHA', 'component']
-        ).findAll { it != null }
-
-        def dataSize = data.size()
-        for (def i = 0; i < dataSize; i++) {
-            def info = data[i]
-            result << getCommitInfo(info)
+        def mergeInfo = processRepositories().collect { record ->
```
s2oBCN, line 35: Please can you change it by a "for" with index. We saw it's better for the nonSerialException

jafarre-viewnext: Actually, I forgot checking the @NonCPS annotations and there's a mistake. I will fix it right away.

jafarre-viewnext: Now the token is correctly obtained from CPS code and the rest of the processing has been made NonCPS, so there will be no NonSerializableException issues.

jafarre-viewnext: As per the conversation maintained, I will change it to an indexed loop anyway.

Issue #851, opened 2022-02-21
NonCPS function calls logger which causes CPS mismatch


albertpuente:
**Describe the bug**
checkIfAllIssuesHaveVersions in DocumentHistory.groovy calls logger.warn, which is a function not annotated with NonCPS, causing the function to return a string instead of a map.

**To Reproduce**
1. Run any release manager pipeline that contains issues where the version is not specified.

**Expected behavior**
The warning should not compromise the execution flow of the function.

**Affected version (please complete the following information):**
 - OpenDevStack 4.x

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
[](https://...)DEBUG: Retrieved latest CFTR version 1 from saved history
[Pipeline] echo
DEBUG: Parsing jira data to document history
expected to call org.ods.orchestration.util.DocumentHistory.computeEntryData but wound up catching org.ods.util.Logger.warn; see: https://jenkins.io/redirect/pipeline-cps-method-mismatches/[](...)
[Pipeline] echo
WARN: Document history not valid. We don't have a version for  the following elements'TESTBUG-252'. If you are not using versioning and its automated document history you can ignore this warning. Otherwise, make sure all the issues have a version attached to it.
[Pipeline] echo
DEBUG: parseJiraDataToDocumentHistoryEntry: versionMap = WARN: Document history not valid. We don't have a version for  the following elements'TESTBUG-252'. If you are not using versioning and its automated document history you can ignore this warning. Otherwise, make sure all the issues have a version attached to it.
[Pipeline] echo
WARN: Error: Creating document of type 'CFTR' for project 'testbug' in phase 'Test' and stage 'PRE_END' has failed: Could not find matching constructor for: org.ods.orchestration.util.DocumentHistoryEntry(java.lang.String, java.lang.Long, java.lang.String, java.lang.String, java.lang.String).
[Pipeline] echo
WARN: Error occured within the orchestration pipeline: Error: Creating document of type 'CFTR' for project 'testbug' in phase 'Test' and stage 'PRE_END' has failed: Could not find matching constructor for: org.ods.orchestration.util.DocumentHistoryEntry(java.lang.String, java.lang.Long, java.lang.String, java.lang.String, java.lang.String).
[Pipeline] echo
DEBUG: [jira-update-release-TESTBUG-210]  (took 1222 ms)
[Pipeline] echo
[Test] **** ENDED orchestration stage **** (took 292671 ms)
[Pipeline] }
[...]

PR #852, opened 2022-02-21
Remove call to a cps method in a non_cps method

jorge-romero:


albertpuente:
Successfully tested in dev env.

clemensutschig:
closes #851 for 4.x

PR #853, opened 2022-02-21
Bump antora page version from 4.x to 5.x in master

tbugfinder:


gerardcl:
I was thinking if we have agreed on already having a 5.x preview, or is this basically master? if not then I would just document up to 4.x (and no previews for now)

michaelsauter:
@gerardcl In the past we always assigned the next major version to `master`, and rendered that as a preview so one can already read up on it.

Not sure if this is worthwhile, or if it would be enough to have the `master` content just in GitHub ... I don't really have an opinion on that.

gerardcl (approved):

tbugfinder:
Well,  ods-documentation builds the pages, so the branches here should be setup properly.

PR #872, opened 2022-03-16
Feautre/move upload jenkins job log

angelmp01:


src/org/ods/services/NexusService.groovy
```diff
@@ -208,6 +212,28 @@ class NexusService {
...
 
+    @NonCPS
+    String uploadJenkinsJobLog(String projectKey, String buildNumber, InputStream jenkinsJobLog, PipelineUtil util) {
+        String nexusPath = "${projectKey.toLowerCase()}/${buildNumber}"
+
+        WeakPair<String, InputStream> file = new WeakPair<String, InputStream>(JENKINS_LOG + ".txt", jenkinsJobLog)
+        WeakPair<String, InputStream> [] files = [ file ]
+
+        String logFileZipped = "${JENKINS_LOG}.zip"
+        byte[] zipArtifact = util.createZipArtifact(logFileZipped, files, true)
```
s2oBCN, line 223: Is it possible to not have in memory the byte[]?

angelmp01: IMO, this improvment is not in the behaviour of this PR.

s2oBCN (approved):

PR #873, opened 2022-03-17
Removal of tests unused fields

victorpablosceruelo:


angelmp01 (approved):

src/org/ods/orchestration/TestStage.groovy
```diff
@@ -73,10 +69,10 @@ class TestStage extends Stage {
                 [it.key.capitalize()],
                 it.value.testResults as Map
             )
-            // Parse again all test report files into a single data structure for one type of test
-            it.value.testResults = junit.parseTestReportFiles(it.value.testReportFiles as List<File>)
         }
 
+        globalData.tests = [ : ]
+
         levaDocScheduler.run(phase, MROPipelineUtil.PipelinePhaseLifecycleStage.PRE_END, [:], globalData)
```
s2oBCN, line 76: why we need to send globalData? as is an empty map and we are not using it for anything (as far as I know)

s2oBCN (approved):

PR #878, opened 2022-03-26
WIP - memory leak fix shared lib Fix/extract enums 4.x

### LINKED ISSUE #857: releasemanager (with EDP, JIRA) . memory used/consumed during run is not reclaimed  / freed up after run
We run jenkins with EDP - for document generation - and can see that memory used during the run is not reclaimed / freed up after

image

the run started at 5:44pm - and was done at 11pm CET - you can see - no GC // reclaim

@braisvq1996 @albertpuente

clemensutschig:
closes #857


.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
jafarre-viewnext, line 1: Was this file committed on purpose?

s2oBCN:
Solver here https://github.com/opendevstack/ods-jenkins-shared-library/pull/902#pullrequestreview-965317488

PR #795, opened 2021-12-10
proposal to fix file uploads when repositoryTipe is 'pipy'

jchaves:
We finally managed to put together a working example of minimal changes needed to correct the situation we detected.

Hopefully this should work without interfering with anything else.

PR #893, opened 2022-04-14
implements pypi bugfix for master

### LINKED ISSUE #785: Method Conflict on repositoryType='pypi' ?
https://github.com/opendevstack/ods-jenkins-shared-library/blob/0d7713e38e9b864d9f2cb638723fd0c57ca521a3/src/org/ods/services/NexusService.groovy#L100

Without knowing too much about the UniRest lib, it seems that the way we are constructing our rest call on the line above is causing some conflict that causes 'pypi' artifacts to end up stored in Nexus with the file name "application/octet-stream".

It would seem that there is a conflict with the method we are trying to call ( 1 ) and the one that is effectively called ( 2 )
[1]: https://github.com/Kong/unirest-java/blob/838cd1d6bb9c80d2e790d8d36fbcf511a248b016/unirest/src/main/java/kong/unirest/HttpRequestMultiPart.java#L69
[2] : https://github.com/Kong/unirest-java/blob/838cd1d6bb9c80d2e790d8d36fbcf511a248b016/unirest/src/main/java/kong/unirest/HttpRequestMultiPart.java#L94

I'm no expert and am in fact having some difficulties following the flow in NexusService, but I guess the fact that we never define a filename in 'nexusParams' for the 'pypi' repositoryType might have something to do with this?

Maybe for type pypi we should use the more explicit [3]: https://github.com/Kong/unirest-java/blob/838cd1d6bb9c80d2e790d8d36fbcf511a248b016/unirest/src/main/java/kong/unirest/HttpRequestMultiPart.java#L87
?

Any feedback will be appreciated!

jchaves:
port of #795 to master
closes #785

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@
 - JiraService.createIssueType(JiraService.groovy:194) fails when description is missing ([#850](https://github.com/opendevstack/ods-jenkins-shared-library/pull/850))
 - Fix resolveJiraDataItemReferences create list with null  ([#865](https://github.com/opendevstack/ods-jenkins-shared-library/pull/865))
 - Add new Infrastructure Service ([#740](https://github.com/opendevstack/ods-jenkins-shared-library/pull/740))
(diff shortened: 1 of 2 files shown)


braisvq1996 (approved):
👍

PR #900, opened 2022-05-02
Fix/extract enums2

s2oBCN:
From https://github.com/opendevstack/ods-jenkins-shared-library/pull/878
Closes https://github.com/opendevstack/ods-jenkins-shared-library/issues/857

Here is how the log looks like:
image


jafarre-viewnext:
As I already commented in the ticket, we are using private or undocumented API here. This is likely to cause problems in the future, whenever we need to change the versions of the JDK or the libraries.
However, the problem needs to be fixed urgently and we haven't found a better solution, so I give my approval.
Amazing job everybody!

jafarre-viewnext (approved):

PR #913, opened 2022-05-27
Feature/retrieve debug information

victorpablosceruelo:
The code developed here generates and saves information that can be used by tests in a future stage of developments/refactors of docGen, so we could test that from the same data in a jenkins pipeline execution we obtain exactly the same pdfs.

It is rather interesting to include it in master before refactors like the ones proposed in:

https://github.com/opendevstack/ods-jenkins-shared-library/tree/feature/afterMoveLevaDoc

https://github.com/opendevstack/ods-document-generation-svc/tree/feature/moveLevaDoc

Thanks.

### DIFF
--- a/src/org/ods/orchestration/FinalizeStage.groovy
+++ b/src/org/ods/orchestration/FinalizeStage.groovy
@@ -8,6 +8,7 @@ import org.ods.orchestration.util.Project
 import org.ods.services.BitbucketService
 import org.ods.services.OpenShiftService
 import org.ods.services.GitService
+import org.ods.util.PipelineDebugInfo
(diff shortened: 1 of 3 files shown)


victorpablosceruelo:
I would like to have a discussion on the contents of this PR. 

It would be an improvement to have it into master and being able to export information from pipelines using previous versions and compare results to new ones. 

Maybe the idea of keeping the files in Jenkins is not good enough (we might be able to find another place to store them) and the idea of running it always is not good enough too, but I think that the idea of being able to ensure that after a change the resulting documents are similar enough is something we need.

@jorge-romero @jafarre-viewnext, what do you think about this?

PR #916, opened 2022-07-01
Feature/support helm in release manager

clemensutschig:
first cut of HELM support, which also once and for all eliviates the DeploymentConfig limitations

This is still dirty! - helm variables from component pipeline are NOT in the RM


michaelsauter:
Alright ... I have silently watched this for a while ... first of, kudos to the nice work!

I thought about the Helm integration on a more general note, and I'm going to throw a bit of a curve ball to you ;)

My basic question is: do you even want to make the Helm integration work like the Tailor integration?

The Tailor integration is very "Deployment(Config) centric", because:

* that's what the first version of the shared lib did
* operators were not a thing
* it allowed easy implementation of some sanity checks and extraction of pod data to render the GxP docs

However, I think long-term you want to be in a place where you can basically deploy any Helm chart. This means that you do not know which resources are your deployment units (think CRDs from operators, like, for example, a "Cluster" resource for a Postgres cluster). So I am wondering if it wouldn't be nice to have the Helm "branch" somewhat separated from the Tailor part. In the Helm "branch", I would question which checks are really needed (e.g. the checks that Deployments are controlled by the release manager are probably not needed at all because the Helm chart cannot be exported from the DEV namespace, so by definition all things are version controlled). Or, is it really needed to collect pod data like IP addresses to put them into the GxP templates when IP addresses may change anyway? Maybe we can use other evidence instead which is better suited and easier to get?

So - just some food for thought. I am aware this would probably blow up the scope of this PR - but it may also make a few things easier that were difficult because of the current constraints?

docs/modules/jenkins-shared-library/partials/odsComponentStageCopyImage.adoc
```diff
@@ -0,0 +1,66 @@
...
+The primary intention is for helm charts so that external images can be imported.
+
+== Options
+
+[cols="1,2"]
+|===
+| Option | Description
+
+
+| *branch* +
```
clemensutschig, line 16: not needed

michaelsauter: Yes I think this is because src/org/ods/component/CopyImageOptions.groovy extends `Options`, which has the branch param. The docs are auto-generated from that. If you do not extend from `Options`, you should be able to get rid of it.

clemensutschig (commented):
Martin - one big thing is the tests... we should write a bunch of tests all around this?

michaelsauter:
From our conversation yesterday @clemensutschig:

* I really really like that this solution doesn't use the `latest` tag any longer, but instead the "short Git Sha tag". Awesome! Do you still want to maintain a `latest` tag at all? BTW - this will mean the spec is updated, which automatically triggers a rollout for deployments. There is some code that does a manual "rollout" if not automatically triggered if I am not mistaken - I think this is not necessary for the `Deployment` path ... you may want to check that.
* I really really like that it is not pausing / unpausing any longer, but I think that collides with #686. I do not really know the consequences of this. May need to document the differences in behaviour?
* From the current implementation, I do not see how one could supply different values files for different environments. So for example, one might maintain something like a `values.dev.yaml`, `values.test.yaml` and `values.prod.yaml` to configure the application depending on the environment. I'm linking https://github.com/opendevstack/ods-pipeline/blob/master/docs/tasks/ods-deploy-helm.adoc, which describes how this works for ODS Pipeline. The approach we took is quite similar to how Tailor can be configured differently based on environments / namespaces. Keep in mind that you also need the same logic for "secrets files" (see next bullet point), which are just encrypted values files.
* It looks like you still need to handle Helm "secrets files" in the "RM flow". In the component pipeline, the Helm upgrade is surrounded by code that exposes a secret key: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy#L192. As a side note, we switched from GPG keys to AGE keys in ODS Pipeline because the handling / CLI tool is so much easier. But that's probably a topic for another day.
* I observed that the `version` / `appVersion` in the Helm chart is not connected to the "version" set in the release manager (typically from Jira). This may need documentation to avoid confusion? In ODS Pipeline we are overwriting the version of the Helm package with the one defined by the user in `ods.yaml`, but tbh I am not sure if that is worth it.
* You should add `--three-way-merge` to the `diff` invocation (see https://github.com/opendevstack/ods-pipeline/blob/master/deploy/ods-pipeline/charts/tasks/templates/task-ods-deploy-helm.yaml#L78-L81). While Helm 3 uses a 3-way merge during `upgrade`, helm-diff only compares previous/new by default. Adding the flag teaches helm-diff to use a 3-way merge as well, which is a good thing for consistency. Otherwise the diff may exit with 0, while the upgrade would have made a change.
* You also may want to set `HELM_DIFF_IGNORE_UNKNOWN_FLAGS=true` for the diff command (https://github.com/opendevstack/ods-pipeline/blob/master/cmd/deploy-with-helm/helm.go#L160) as that ensures the diff command also works when someone adds more flags to `helm upgrade` that helm-diff does not understand. If you add this env var, I believe you do not need https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/services/OpenShiftService.groovy#L140 anymore.

I hope this is helpful, let me know if there are other questions that I may help with!

clemensutschig:
@michaelsauter - thanks for for the hard challenge . this helps a lot ..

`From the current implementation, I do not see how one could supply different values files for different environments. So for example, one might maintain something like a values.dev.yaml, values.test.yaml and values.prod.yaml to configure the application depending on the environment. I'm linking https://github.com/opendevstack/ods-pipeline/blob/master/docs/tasks/ods-deploy-helm.adoc, which describes how this works for ODS Pipeline. The approach we took is quite similar to how Tailor can be configured differently based on environments / namespaces` - is fixed now - thru a new param, which allows you to pass those env dependent files :)

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -137,8 +137,11 @@ class OpenShiftService {
         valuesFiles.collect { upgradeFlags << "-f ${it}".toString() }
         values.collect { k, v -> upgradeFlags << "--set ${k}=${v}".toString() }
         if (withDiff) {
-            def diffFlags = upgradeFlags.findAll { it != '--atomic' }
+            def diffFlags = upgradeFlags.findAll { it != '--atomic' || it != '--three-way-merge'}
```
michaelsauter: I do not understand this line. You wouldn't need to check since you handle the case that `--three-way-merge` is already given down below, no? Apart from that, `--three-way-merge` isn't a known flag for `helm upgrade`.

`--atomic` is filtered out because it is a known flag for `helm upgrade`, but it is not handled by `helm diff`. But as I said I think it's better to set the env var to ignore unknown flags in diff, otherwise you prevent most upgrade flags to be used as it would break helm diffing.

test/groovy/vars/OdsComponentStageRolloutOpenShiftDeploymentSpec.groovy
```diff
@@ -187,18 +191,17 @@ class OdsComponentStageRolloutOpenShiftDeploymentSpec extends PipelineSpockTestB
     }
     def deploymentInfo = script.call(context)
 
+    // We really only care about the call to `helmUpgrade'
     then:
     printCallStack()
     assertJobStatusSuccess()
-    deploymentInfo['DeploymentConfig/bar'][0].deploymentId == "bar-124"
```
clemensutschig, line 193: I am surprised this was removed ... I would expect that we get a deployment back ... ?

vars/odsComponentStageCopyImage.groovy
```diff
@@ -0,0 +1,59 @@
+import org.ods.component.CopyImageStage
+import org.ods.component.IContext
+
+import org.ods.services.OpenShiftService
+import org.ods.services.ServiceRegistry
+import org.ods.util.Logger
+import org.ods.util.ILogger
+
+def call(IContext context, Map config = [:]) {
```
clemensutschig, line 9: test missing ... and you can take a lot from the rolloutstage test above .. (here we should find imported images in the .builds)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Remove orchestration pipeline restriction ([#867](https://github.com/opendevstack/ods-jenkins-shared-library/pull/867))
```
clemensutschig, line 4: remove the orchestration .. restriction on WHAT? ..

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+        // to make this work, we need a target image stream first
+        openShift.findOrCreateImageStream("${context.cdProject}", "${this.options.image.split(':').first()}")
+
+        // read the target registry auth (for this project's -cd project ..)
+        def targetInternalRegistryToken = steps.readFile '/run/secrets/kubernetes.io/serviceaccount/token'
+
+        def sourcetoken = options.sourceCredential ? "--src-creds ${options.sourceCredential}" : ''
+
+        try {
```
clemensutschig, line 49: we should change this to returnStatus in the sh task ...(and add a nice label) -rather than the quick way as done here

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,90 @@
...
+            openShift.importImageTagFromProject(
+                context.targetProject, imageName, context.cdProject,
+                imageTag, imageTag
+            )
+            logger.info("!!! Image ${this.options.image} tagged into ${context.targetProject}")
+        }
+    }
+
+    protected String stageLabel() {
+        return "${STAGE_NAME} (${context.componentId})"
```
clemensutschig, line 88: we should change this to contain the image name - and not just the component id ...

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+        // read the target registry auth (for this project's -cd project ..)
+        def targetInternalRegistryToken = steps.readFile '/run/secrets/kubernetes.io/serviceaccount/token'
+
+        def sourcetoken = options.sourceCredential ? "--src-creds ${options.sourceCredential}" : ''
+
+        try {
+            steps.sh """
+            set +x
+            skopeo copy --src-tls-verify=false ${sourcetoken} \
```
michaelsauter, line 52: `--src-tls-verify=false`?

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,90 @@
...
+            ${this.options.registry}/${this.options.repo}/${this.options.image} \
+            --dest-creds openshift:${targetInternalRegistryToken} \
+            ${STR_DOCKER_PROTOCOL}${context.clusterRegistryAddress}/${context.cdProject}/${this.options.image} \
+            --dest-tls-verify=false
```
michaelsauter, line 56: `--dest-tls-verify=false`?

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,90 @@
...
+            """
+        } catch (err) {
+            throw err
+        }
```
michaelsauter, line 60: This try/catch seems useless ;)

src/org/ods/component/RolloutOpenShiftDeploymentOptions.groovy
```diff
@@ -47,6 +47,13 @@ class RolloutOpenShiftDeploymentOptions extends Options {
      * directory referenced by `chartDir` exists. */
     List<String> helmValuesFiles
 
+    /**
+     * List of paths to values files (empty by default). Only relevant if the
+     * directory referenced by `chartDir` exists.
+     * These must contain a suffix called '.env.yml' - which will be replaced
```
michaelsauter, line 53: Is this meant to say that `env` gets replaced? If so, maybe change the notation to e.g. `.<env>.yml' - that would make it clearer for me.

michaelsauter: Forget my comment, understood how it works now. Maybe we can make an example here that tells users they need to create values.dev.yaml, values.test.yaml and values.prod.yaml - and then if they pass values.env.yaml here, those files will be picked up.

src/org/ods/component/CopyImageStage.groovy
michaelsauter: Please see https://github.com/opendevstack/ods-pipeline/blob/master/cmd/deploy-with-helm/main.go#L189-L218, this should be possible.

clemensutschig:
> > I think this iready to be merged .. @metmajer @michaelsauter - for a very big first cut - it's feature wise at what tailor does / did
> 
> **Thanks for all the help I received here.** Should I open a new issue to craft a quickstarter that uses helm so people see how it would work?

This warrants a bigger discussion- how we want to handle this ... :/ (e.g. do it for all quickstarters in one short, .. or a flag, .. or)

docs/modules/jenkins-shared-library/partials/odsComponentStageCopyImage.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+ This needs to be in the following format: [REGISTRY/]REPO/IMAGE[:TAG]
+
+
+| *tagIntoTargetEnv* +
+_Boolean_
+|true will tag the image from the -cd namespace into the targetEnvironment that the pipeline is running for
+
+
+| *verifyTLS* +
```
metmajer, line 67: Sure about this as an option?

docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
```diff
@@ -158,6 +171,10 @@ _boolean_
 
 |===
 
-All the created or updated OpenShift resources are automatically labeled to ease their management.
+== Differences between tailor and helm deployments
+
+When tailor does the rollout, all the created or updated OpenShift resources are automatically labeled to ease their management.
+This is in contrast to helm rollouts which rely on the chart providing the desired labels.
```
metmajer, line 177: This might be problematic, since we rely on certain labels to able to correlate resources with their respective systems, e.g. with Aqua. FYI @maksymmaksym

src/org/ods/component/Context.groovy
```diff
@@ -74,6 +74,7 @@ class Context implements IContext {
         config.buildUrl = script.env.BUILD_URL
         config.buildTag = script.env.BUILD_TAG
         config.buildTime = new Date()
+        config.dockerRegistry = script.env.DOCKER_REGISTRY
```
metmajer, line 77: Would prefer to remove Docker since OpenShift serves a registry for OCI-compliant container images.

src/org/ods/component/CopyImageOptions.groovy
```diff
@@ -0,0 +1,45 @@
...
+    String sourceCredential
+
+    /**
+     * verifyTLS allows the stage to ignore certificate validation errors.
+     *
+     * The default is to verify certificate paths
+     */
+    Boolean verifyTLS
+
+    // INTERNAL!!
```
metmajer, line 32: Don't you want to use the `private` access modifier to make the internal nature transparent?

src/org/ods/component/CopyImageStage.groovy
metmajer: @serverhorror if conversations have been resolved, please visibly "resolve" them.

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,105 @@
...
+        def sourcetoken = options.sourceCredential ? "--src-creds ${options.sourceCredential}" : ''
+
+        int status = copyImage(sourcetoken, targetInternalRegistryToken, STR_DOCKER_PROTOCOL)
+        if (status != 0) {
+            script.error("Could not copy `${this.options.sourceImageUrlIncludingRegistry}', status ${status}")
+        }
+
+        def imageName = this.options.image.split(':').first()
+        def imageTag = this.options.image.split(':').last()
+
```
metmajer, line 62: You have provided some inline comments above about the "algorithm". Why not continue here to aid readability?

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -108,52 +111,62 @@ class RolloutOpenShiftDeploymentStage extends Stage {
...
-                    steps.error "Helm cannot be used in the orchestration pipeline yet."
-                    return
-                }
+            if (isHelmDeployment) {
+                logger.info "Rolling out ${context.componentId} with HELM, selector: ${options.selector}"
                 refreshResources = true
                 helmUpgrade(context.targetProject)
-            } else if (steps.fileExists(options.openshiftDir)) {
+            }
+            if (steps.fileExists(options.openshiftDir)) {
```
metmajer, line 139: What happens if both Chart.yaml and the openshiftDir exists?

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -108,52 +111,62 @@ class RolloutOpenShiftDeploymentStage extends Stage {
...
+                paused = false
+                return rolloutData
+            } else {
+                rolloutData = rollout(deploymentResources, originalDeploymentVersions, true)
+                paused = false
+                return rolloutData
+            }
         } finally {
-            if (paused) {
+            if (paused && !isHelmDeployment) {
```
metmajer, line 169: Please put Tailor and Helm-related logic into a Strategy class and apply the one or the other strategy vs. having a single logic that tries to handle both. This is otherwise hard to read and mock/test.

Example: class Tailor {
    prepare(context,...)
    apply()
    isApplicable()
    ...
}

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -193,12 +208,31 @@ class RolloutOpenShiftDeploymentStage extends Stage {
                 if (pkeyFile) {
                     steps.sh(script: "gpg --import ${pkeyFile}", label: 'Import private key into keyring')
                 }
-                options.helmValues.imageTag = options.imageTag
+
+                // we add two things persistent - as these NEVER change (and are env independent)
+                options.helmValues['registry'] = context.clusterRegistryAddress
```
metmajer, line 213: Does `tailorApply` read `options.helmValues` here?

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -240,8 +277,33 @@ class RolloutOpenShiftDeploymentStage extends Stage {
                     originalVersion = originalVersions[resourceKind][resourceName] ?: 0
                 }
 
-                def podData = rolloutDeployment(resourceKind, resourceName, originalVersion)
-
+                def podData = [:]
+                if (!isHelm) {
```
metmajer, line 281: See above, please provide a clean separation of concerns. Support and maintenance will be improved.

src/org/ods/orchestration/util/Project.groovy
```diff
@@ -1524,6 +1528,11 @@ class Project {
         "${getKey()}-cd-${secretName}"
     }
 
+    String getHelmPrivateKeyCredentialsId() {
```
metmajer, line 1531: Why is this not part of a HelmService?

metmajer (changes requested):
Hi @serverhorror @clemensutschig! I have some concerns about the quality of the code. There is little encapsulation and tests do not cover the many changes which were made. I am not confident approving this yet, code should show better cohesion and testability. Also, labels are not applied when using Helm, which will cause issues with Aqua. We had done extensive work previously and should seek to provide the same labels with all deployment methods.

metmajer:
We had a discussion about dedicated Helm Quickstarters and decided that Helm/Tailor will be based on conventions of the existing Quickstarters. Helm is not a feature of a Quickstarter, it's just the deployment mechanism. What about the `fe-be-monorepo` quickstarter? Has this been tested?

metmajer:
Sure, @serverhorror, let's have a chat on this 👍

metmajer:
@maksymmaksym can you please post the data you currently monitor in Aqua to enable Incident Management?

metmajer:
@serverhorror there are certain labels we need to apply to the deployed resources, answer from @maksymmaksym pending.

You asked for details: I suggest to get familiar with the *Strategy Pattern* to implement a deployment strategy. An abstract `DeploymentStrategy` class would implement logic which is generic and therefore independent from Tailor or Helm and provides *Template Methods* which need to be implemented by its concrete subclasses `TailorDeploymentStrategy` and `HelmDeploymentStrategy`. Template methods could focus on the initialisation, application, or post-processing (e.g. apply labels) of a particular deployment method.

The implementation of the strategy could start with taking the code I reviewed and extracting it as-is into the two concrete strategy classes, then remove what's not needed for this particular strategy. This way you could come up with a single `apply(context)` method per strategy, which does Helm or Tailor, but not both. Depending on whether Helm or Tailor would be used, the one of the other strategy would be applied. In the case of Helm, the `HelmDeploymentStrategy` might be served via `HelmService::getDeploymentStrategy()` method. A `HelmService` is useful for encapsulating additional service methods, e.g. `getHelmPrivateKeyCredentialsId()`. Once this is implemented, the contents of the concrete strategy classes could be refactored into additional template methods or methods in the abstract `DeploymentStrategy` class if the logic would otherwise be duplicated across the Helm and Tailor classes.

Does that help?

src/org/ods/component/CopyImageOptions.groovy
metmajer: @serverhorror You can remove the "X is an INTERNAL value" comments, since this is already indicated by the `private` access modifier. `protected` would have been an alternative if you need to make the same field available in a subclass, but that's not the case.

maksymmaksym:
@metmajer for every deployed resource we would like to track the associated CI name and a person who deployed the workload.

metmajer:
Thanks @maksymmaksym. Hmm.. but the name of the person who deployed is not contained in the current implementation. The end-responsibility of a deployed system is in the hands of the system lead who is documented elsewhere, correct? Can I assume that the CI name (System Name) would then be enough, albeit a bit less convenient?

maksymmaksym:
@metmajer if there is a CIs in all the cases its should be fine.

metmajer:
@serverhorror following the discussion with @maksymmaksym, the system name is enough. We do not want to add the user who deployed. Currently, we apply the following labels when using component pipelines or the Release Manager pipeline. `app.opendevstack.org/system-name` is the minimum requirement from security, but I urge you to apply the others from the Release Manager as well.

Screenshot 2022-08-18 at 22 22 08

clemensutschig:
To which resource types does this apply?

Von meinem iPhone gesendet

jafarre-viewnext:
> To which resource types does this apply? I guess we could try to find all resources and label them in one shot - rather than patching https://www.krenger.ch/blog/openshift-add-or-remove-label/#:~:text=You%20can%20find%20more%20information%20on%20labeling%20nodes,the%20example%20%E2%80%9Cmylabel%E2%80%9D%29%20from%20a%20node%20or%20pod%3A @serverhorror
> […](#)

The labels are currently applied to all resources in the following way:
- The quickstarter pipeline labels all created resources with default labels.
- The component pipeline updates the labels for every deployment, as some of the values may (and will) change.
- When the component pipeline updates the labels, it takes into account whether it is being run through the release manager and applies additional labels: system name, change id (project-version) and whether we are performing a developer preview or a release.
- It is also taken into account whether there is a helm chart, in order to apply helm-specific labels. This may need to be reviewed.
- The deployment configs are also updated to add/updated the labels to the templates. This allows new pods to be properly labeled whenever they are created.


`OpenShiftResourceMetadata` should do all the job without (or almost without) modification. @serverhorror if you see the need to modify it or just need more information about how it works, don't hesitate to contact me.

Note: it was a bit tricky to avoid unnecessary automatic rollouts while performing all these actions. It would be a good idea to make sure that, after the modifications, we are still performing the minimum possible number of rollouts (hopefully one).

src/org/ods/component/Context.groovy
clemensutschig: if we change this we need to change it in ods-core etx.. i suggest to leave for now

src/org/ods/orchestration/util/Project.groovy
clemensutschig: agree!

metmajer:
@serverhorror besides what you‘re thinking, why specifically is the use of our existing functions a problem in conjunction with Helm and where specifically do you see a race condition?

metmajer:
@serverhorror I thought that unwanted redeployments were previously addressed by @jafarre-bi. Juan, can you comment?

metmajer:
Let me add a bit more context to your thinking @serverhorror.

1. We previously decided to attach metadata as OpenShift labels since downstream (required) security tools did not support consuming annotations.
2. Concerning the ownership of these metadata, we see ODS as an opinionated framework and therefore decided to attach specific metadata automatically and consistently for identification and housekeeping purposes. Therefore, it's also OK to override any user-provided input. We consider this viable since we use our own namespaces. Users are free to apply other labels or annotations as they see fit.

We had a discussion about the labelling strategy in https://github.com/opendevstack/ods-quickstarters/issues/512 and documented the final outcome in https://github.com/opendevstack/ods-jenkins-shared-library/pull/715. Hope this helps!

metmajer:
@serverhorror @jafarre-bi given Martin's input that Pod-labels cannot be reliably applied using Kubernetes’ Deployment configurations (a standard with Helm charts), as opposed to OpenShift-specific DeploymentConfig resources, we're discussing internally how precisely we can enable security-related incident monitoring for the former without those labels in place with @maksymmaksym

tbugfinder:
@serverhorror - Your awesome summary has two pro/con sections named "Mutating Admission Controller". Could you edit the post?

metmajer:
@serverhorror thanks for the write-up! Currently travelling, but will comment shortly.

metmajer:
@serverhorror will you consider @michaelsauter's PR for this one?

metmajer:
Hello @jafarre-bi and @jorge-romero. After @maksymmaksym's approval, please give this contribution from @serverhorror a review. Martin has worked out a solution most welcome by our users and applied some decent refactorings. Please ensure refactorings are properly tested, after all deployment strategies are at the heart of delivery. Thanks much!

metmajer:
@braisvq1996 once reviewed, I request that we include this feature in an ODS release soon after our current EDP release. We have users eagerly waiting for this.

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -0,0 +1,260 @@
...
+
+        logger.info "Rolling out ${context.componentId} with HELM, selector: ${options.selector}"
+        refreshResources = true
+        helmUpgrade(context.targetProject)
+
+        if (refreshResources) {
+            deploymentResources = openShift.getResourcesForComponent(
+                context.targetProject, DEPLOYMENT_KINDS, options.selector
+            )
+        }
```
jorge-romero, line 109: The way the variable refreshResources is used make the if condition in line 105 always be true, it does not depends on configuration or previous execution results

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -0,0 +1,255 @@
...
+    private void applyMandatoryLabels(Map<String, String> labels, Map<String, List<String>> deploymentResources) {
+        logger.debug("${this.class.name} -- MANDATORY LABELS")
+        logger.debug(
+            JsonOutput.prettyPrint(
+                JsonOutput.toJson(labels)))
+        logger.debug("${this.class.name} -- MANDATORY LABELS (cleaned)")
+        logger.debug(
+            JsonOutput.prettyPrint(
+                JsonOutput.toJson(labels)))
+        labels.remove { it -> it.value == null }
```
jafarre-bi, line 132: Null values are used to mark labels for deletion by OpenShiftResourceMetadata.updateMetadata(). In some cases, one or more optional labels may not apply anymore and need to be deleted. If you just ignore them, we may end up with wrong labels set.
Actually OpenShiftResourceMetadata.updateMetadata should do its job correctly, regardless of using Tailor or Helm. Is there any difference that makes it unsuitable for the Helm strategy? It should be possible to adapt it.

jafarre-bi: Sure, I will schedule a meeting

src/org/ods/component/CopyImageStage.groovy
```diff
@@ -0,0 +1,106 @@
...
+            [image: internalImageRef, source: options.sourceImageUrlIncludingRegistry])
+
+        // If this option is set will additionally tag the image in the target namespace, not just the cd namespace
+        if (options.tagIntoTargetEnv) {
+            openShift.findOrCreateImageStream(context.targetProject, imageName)
+            openShift.importImageTagFromProject(
+                context.targetProject, imageName, context.cdProject,
+                imageTag, imageTag
+            )
+            logger.info("!!! Image ${this.options.image} tagged into ${context.targetProject}")
```
metmajer, line 82: Note to self and @jafarre-bi @jorge-romero for the planned activity of improving log messages, and error messages in particular, we may want to use class name prefixes instead of `!!!`. Also, any errors which are known and can be handled, we need to introduce error numbers which resolve to help documentation.

src/org/ods/component/HelmDeploymentStrategy.groovy
metmajer: @jafarre-bi were you able to resolve your questions in the meeting?

src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy
```diff
@@ -5,20 +5,23 @@ import groovy.transform.TypeCheckingMode
...
 @TypeChecked
 class RolloutOpenShiftDeploymentStage extends Stage {
 
     public final String STAGE_NAME = 'Deploy to OpenShift'
-    private final List<String> DEPLOYMENT_KINDS = [
-        OpenShiftService.DEPLOYMENT_KIND, OpenShiftService.DEPLOYMENTCONFIG_KIND,
-    ]
+//    private final List<String> DEPLOYMENT_KINDS = [
+//        OpenShiftService.DEPLOYMENT_KIND, OpenShiftService.DEPLOYMENTCONFIG_KIND,
+//    ]
```
metmajer, line 18: If this code and the imports on top are not used, can we remove it?

metmajer (commented):
Hello @serverhorror, a few quick thoughts:

- Some Specs show commented-out code, is this code supposed to stay?
- There is no Spec available for the TailorDeploymentStrategy, but it would really beneficial to have one. We're operating at the core of delivery and I would appreciate your support here.

src/org/ods/component/HelmDeploymentStrategy.groovy
metmajer: @serverhorror ok, we'll discuss internally!

metmajer:
> Is this not good enough?

@serverhorror Yes, it is for me. I missed that one!

src/org/ods/component/ScanWithAquaStage.groovy
```diff
@@ -234,7 +234,7 @@ class ScanWithAquaStage extends Stage {
         try {
             URI report = nexus.storeArtifact(
                 "${nexusRepository}",
-                "${context.projectId}/${context.componentId}/" +
+                "${context.projectId}/${this.options.resourceName}/" +
```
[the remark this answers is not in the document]
metmajer: @serverhorror this concerns Aqua, and shouldn't be mixed into a Helm-related PR. Please undo and open a new PR with this change.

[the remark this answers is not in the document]
clemensutschig: This is a must fix - we saw this breakibg when Helm is used..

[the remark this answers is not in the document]
metmajer: @clemensutschig I clarified already with Martin. we will take a look to ensure that we're not inconsistent with other integrations that follow the same pattern.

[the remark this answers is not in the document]
hrcornejo: @metmajer @serverhorror we have now two integrations of this type: Aqua and SonarQube. The code afected for SonarQube is in: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/ScanWithSonarStage.groovy#L282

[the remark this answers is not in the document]
jafarre-bi: The latest comment by @hrcornejo has already been solved. I consider this conversation resolved.

jafarre-bi:
IMO this is ready to merge. I see there have been build errors that should be solved first.
I give my approval now, but please, do not merge it before the tests that are under development are complete and pass.
Thank you and congratulations, @serverhorror, for the terrific work!

jafarre-bi (approved):

metmajer (approved):
Well done, @serverhorror, kudos for the hard work and contribution :) Please wait with the merge until our test cases indicate success. We are working on it and will let you know soon. Then, let's please ensure we squash-merge this thing :)

metmajer:
To add more context to this conversation and hopefully clarify: the PR by @serverhorror is most welcome and will be released early next year. Therefore, we are not under pressure of merging this code into `master`. We see benefits in making contributions, including our own ones, _Done_ before having them merged, which enables us to forget about them for now and expect successful test executions during our formal release process vs. being unnecessarily blocked later on when the deep knowledge has already started to vanish. Quality becomes of extra importance when building a Platform, especially when changing something at its core. Thanks for your understanding.

jafarre-bi (approved):

metmajer:
Over the last weeks, @serverhorror @jafarre-bi and others worked on triaging random errors during pipeline executions using Helm. We didn't see any clear evidence that this was related to Helm directly but needed a better understanding of the overall stability to decide on the next steps. When the team was confident about the final result, we decided to consider the implementation Done and to give the merge a go. Thanks for everyone's patience. Some contributions need a close look to ensure the quality of the resulting feature. This was certainly the case here. Well done, everyone!

PR #920, opened 2022-07-18
Feature/edpc 1323 dev preview from release branch if exists

victorpablosceruelo:
IMPLEMENTATION GENERAL IDEAS
- [x] After deploying to D, if we come back to development phase and request for "developer preview" this one must use as branch the one created when deploying to D. In plain text, the "developer preview" functionality needs to use the branch created for "deploying to D" (we call it "release branch" from now on) if that one exists. When the release branch does not exist, we'll use master for the release manager. When the release branch does not exist and it is not the release manager, we'll use the branch reported in Release Manager's metadata.yml cfg (or master if it is not reported).
- [x] The idea behind this behaviour is that developers continue working in the stabilization branch until they promote.

DRAWBACKS
- [x] Developers need to pull the changes to "master" branch (or equivalent) manually, when until now they worked in the other way: they implemented in master and used "developer preview" to see the results before merging into the stabilization branch.  

NOTES:
- [x] Needs to checkout the release branch if it exists, when still in WIP.
- [x] Needs to commit to that branch, not to master as done until now.
- [x] Needs to implement this behaviour for all the components, not only for the release manager.
- [x] Needs to keep previous behaviour when not in WIP.

ENSURE:
- [x] Behaviour is only changed when in WIP. 
- [x] When release branch is found, every operation against the git repo (pull, push, ...) must be done against that branch.

MANDATORY
- [x] Implement tests that ensure code behaves as expected.

### DIFF
--- a/src/org/ods/orchestration/FinalizeStage.groovy
+++ b/src/org/ods/orchestration/FinalizeStage.groovy
@@ -98,6 +98,9 @@ class FinalizeStage extends Stage {
 
         levaDocScheduler.run(phase, PipelinePhaseLifecycleStage.PRE_END)
 
+        logger.debug("Project has failing tests? ${project.hasFailingTests()}")
+        logger.debug("Project has unexecuted jira tests? ${project.hasUnexecutedJiraTests()}")
+
         // Fail the build in case of failing tests.
         if (project.hasFailingTests() || project.hasUnexecutedJiraTests()) {
             def message = 'Error: '
@@ -124,6 +127,7 @@ class FinalizeStage extends Stage {
             util.failBuild(message)
             throw new IllegalStateException(message)
         } else {
+            logger.debug("Reporting pipeline status to Jenkins...")
             project.reportPipelineStatus()
             if (!project.isWorkInProgress) {
                 bitbucket.setBuildStatus (steps.env.BUILD_URL, project.gitData.commit,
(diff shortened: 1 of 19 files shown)


victorpablosceruelo:
@fbeba-bi, @petod108, @hrcornejo, @jafarre-viewnext: this is ready for review. 
Thanks !!!

hrcornejo:
@victorpablosceruelo seems good to me, but I couldn't try in a box (connections problems) so I can't verify.

hrcornejo (approved):

src/org/ods/orchestration/FinalizeStage.groovy
```diff
@@ -145,7 +145,8 @@ class FinalizeStage extends Stage {
             repoPushTasks << [ (repo.id): {
                 steps.dir("${steps.env.WORKSPACE}/${MROPipelineUtil.REPOS_BASE_DIR}/${repo.id}") {
                     if (project.isWorkInProgress) {
-                        git.pushRef(repo.branch)
```
victorpablosceruelo, line 148: This change needs to be reverted.
If problematic, we'll need to evaluate why.

victorpablosceruelo: Doing change in PR https://github.com/opendevstack/ods-jenkins-shared-library/pull/925/

Issue #938, opened 2022-09-22
ReleaseManger failed job is not keeping status from Running to Failed


petod108:
**Describe the bug**
Failing Relasemanager job is not setting status 'Failed' and instead is keeping status 'Running'

**To Reproduce**
http://ec2-54-155-126-35.eu-west-1.compute.amazonaws.com/job/edp-e2e-tests/job/e2e/job/master/job/ApproveFailedDeployReleaseSpec/66/
1. first QuickStarter for E2E project Geb-Spock created project with missing metadata.yml
2. set E2E project to has failing test 
3. Developers preview failed because of missing metadata.yml
4. changed status manually to Success
5. run "Deploy to D" failed because of missing metadata.yml
6. changed status manually to Success
7. move to 'D SLC DOCS APPROVED'
8. run "Deploy to Q" failed because of Error: "unable to find latest tag for version 1.0/1.0."
9. status of release manager didn't changed never from Running to Failed

**Expected behavior**
Even Release manager failed is should change status from Running to Failed

**Screenshots**
[Jenkins Log](http://ec2-54-155-126-35.eu-west-1.compute.amazonaws.com/job/edp-e2e-tests/job/e2e/job/master/job/ApproveFailedDeployReleaseSpec/66/)

**Affected version (please complete the following information):**
 - ods-jenkins-shared-library (master: 59d3d7be5452cc77c9387fc0f32c875b4fc95365)

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
12:16:45      java.lang.RuntimeException: Error: unable to find latest tag for version 1.0/1.0.
12:16:45      	at org.ods.orchestration.InitStage.checkOutRepoInPromotionMode(InitStage.groovy:498)
12:16:45      	at org.ods.orchestration.InitStage.checkOutReleaseManagerRepository(InitStage.groovy:440)
12:16:45      	at org.ods.orchestration.InitStage.run(InitStage.groovy:56)
12:16:45      	at org.ods.orchestration.Stage.execute(Stage.groovy:40)
12:16:45      	at ___cps.transform___(Native Method)
12:16:45      	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
12:16:45      	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[...]

PR #939, opened 2022-09-22
#938 updated release manager status when an error raise in the checkout

angelmp01:
When the git checkout raise an Exception, the RM is not informed because the Registry is not initialized.

### DIFF
--- a/src/org/ods/orchestration/InitStage.groovy
+++ b/src/org/ods/orchestration/InitStage.groovy
@@ -53,7 +53,12 @@ class InitStage extends Stage {
         // to concurrent releases).
         def envState = loadEnvState(logger, buildParams.targetEnvironment)
 
-        Exception checkOutException = checkOutReleaseManagerRepository(buildParams, git, logger)
+        def hasErrorsDuringCheckout = false
+        try {
+            checkOutReleaseManagerRepository(buildParams, git, logger)
+        } catch (Exception) {
+            hasErrorsDuringCheckout = true
+        }
 
         logger.debug 'Load build params and metadata file information'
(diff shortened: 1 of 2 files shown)


src/org/ods/orchestration/InitStage.groovy
```diff
@@ -479,7 +487,7 @@ class InitStage extends Stage {
         }
     }
 
-    private void checkOutRepoInPromotionMode(GitService git,
+    private Exception checkOutRepoInPromotionMode(GitService git,
```
victorpablosceruelo, line 490: I think this change is not necessary...

src/org/ods/orchestration/InitStage.groovy
```diff
@@ -430,26 +434,30 @@ class InitStage extends Stage {
         }
     }
 
-    private void checkOutReleaseManagerRepository(def buildParams, def git,
+    private Exception checkOutReleaseManagerRepository(def buildParams, def git,
```
victorpablosceruelo, line 437: I would have written the try/catch block in the method that calls this one.
This one's complexity is far away than what a normal reader can understand and introducing the try/catch blocks complicates it even more.
But this is just an opinion. You can leave it there if you consider it is better in this way.

victorpablosceruelo (changes requested):

src/org/ods/orchestration/InitStage.groovy
angelmp01: done

braisvq1996 (approved):

victorpablosceruelo (approved):

PR #942, opened 2022-09-29
Remove reflection in docs creation and serializable problem in arraylist iterator (inside collectentries)

victorpablosceruelo:
We found 2 issues in documents creation:
- [x] Calling a method using java reflection causes the following error (1), which hides another errors in called methods.
- [x] Method collectentries uses java iterator, which is not serializable. Needs to be removed from code.

(1)
```
[docgen-edp4tb-CFTP] Creating document of type 'CFTP' for project 'edp4tb' in phase 'Build' and stage 'POST_START'
expected to call org.ods.orchestration.usecase.LeVADocumentUseCase.invokeMethod but wound up catching 
org.ods.orchestration.usecase.LeVADocumentUseCase.createCFTP; 
see: https://jenkins.io/redirect/pipeline-cps-method-mismatches/
```


metmajer:
@victorpablosceruelo the reflection-based approach is working though, the message is misleading. Can you describe the problem despite the message?

hrcornejo:
@victorpablosceruelo @metmajer IMO the `expected to call...` is not a problem. When a Serialization problem occurs (or other problem) this doesn't hide the error.

jafarre-bi:
> @victorpablosceruelo @hrcornejo the reflection-based approach is working though, the message is misleading because Jenkins's CPS context does not expect this. Can you describe the problem despite the message? Overall, I think that the implementation just got a lot more complex.

metmajer:
@jafarre-bi @hrcornejo if I understand Hector correctly, then this PR does not solve the problem that any information is masked by the CPS warning message. Is my understanding correct?

jafarre-bi:
Accidentally closed the PR and re-opened it.

I agree with @metmajer and @hrcornejo 
@victorpablosceruelo Concretely about unrolling that collect, this can better be achieved by extracting it into a nonCPS method. The only thing is that the inner log.debug should be extracted to the CPS calling method.
However, I am not sure that change cannot wait. Did you often find exceptions regarding serialization in that concrete collect?

jafarre-bi:
> @jafarre-bi @hrcornejo if I understand Hector correctly, then this PR does not solve the problem that any information is masked by the CPS warning message. Is my understanding correct?

About this I am not totally sure. Calling a CPS method from a NonCPS one raises an exception. This exception could indeed mask other exceptions. @victorpablosceruelo Do you have some concrete example you could show to us about this message hiding other errors?

victorpablosceruelo:
> @victorpablosceruelo @hrcornejo the reflection-based approach is working though, the message is misleading because Jenkins's CPS context does not expect this. Can you describe the problem despite the message? Overall, I think that the implementation just got a lot more complex.

Hi Martin.

It is true that the implementation got a little bit complex by removing reflection, but the message Jenkins gave us tells us we cannot assure the implementation will behave as expected after this message shows up.

Besides, it hides errors in the called methods and we waste a lot of time debugging to see where the problem is coming from.

Although I neither like the new implementation and I clearly think that running reflection in groovy code should work well and not like it does (calling java code to call non-java code later), at the point we are right now with the memory problem in

https://github.com/opendevstack/ods-jenkins-shared-library/issues/857

I really think this is necessary. We need to remove from the equation variables we cannot control easily. The new code is harder to maintain, ugly to see, more complex because the old one is just one line of code, etc. But once you read it the idea is really simple. You now I love the KISS philosophy: Keep It Simple Stupid. 

I hope you understand and approve the PR (adding you right now). This time I will prefer to have as approvals all reviewers. Hope you all understand.

victorpablosceruelo:
> I agree with @metmajer and @hrcornejo @victorpablosceruelo Concretely about unrolling that collect, this can better be achieved by extracting it into a nonCPS method. The only thing is that the inner log.debug should be extracted to the CPS calling method. However, I am not sure that change cannot wait. Did you often find exceptions regarding serialization in that concrete collect?

Sure, we could do it better, but then we would need much more time for testing it. This is so simple that I am sure it will do its work.
And the answer to "did you often find" is: we were trying to run the pipeline in the DEV environment and we could not pass this part of the code. After we did this change it ended as expected. 

I do not know guys. My feeling is that with this PR we simplify and remove a serialization problem, but I will not merge this into master if you consider this is not necessary.

victorpablosceruelo:
> > @jafarre-bi @hrcornejo if I understand Hector correctly, then this PR does not solve the problem that any information is masked by the CPS warning message. Is my understanding correct?
> 
> About this I am not totally sure. Calling a CPS method from a NonCPS one raises an exception. This exception could indeed mask other exceptions. @victorpablosceruelo Do you have some concrete example you could show to us about this message hiding other errors?

I am sorry I do not have the example, but I remember seeing strange behaviours.
Nevertheless, Internet is full of unexpected behaviours when you call cps code from noncps code. Examples:

https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/
https://stackoverflow.com/questions/69484244/jenkins-groovy-why-does-withcredentials-introduce-a-notserializableexception

What I cannot assure is that it will work always if we leave the old code...

But as I said before, I will not merge without all your approvals guys.

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -1729,12 +1730,23 @@ class LeVADocumentUseCase extends DocGenUseCase {
...
+        result.put(DocumentType.CSD as String, getVersionString(DocumentType.CSD))
+        result.put(DocumentType.SSDS as String, getVersionString(DocumentType.SSDS))
+        result.put(DocumentType.RA as String, getVersionString(DocumentType.RA))
+        result.put(DocumentType.TRC as String, getVersionString(DocumentType.TRC))
+        result.put(DocumentType.DTP as String, getVersionString(DocumentType.DTP))
+        result.put(DocumentType.DTR as String, getVersionString(DocumentType.DTR))
+        result.put(DocumentType.CFTP as String, getVersionString(DocumentType.CFTP))
+        result.put(DocumentType.CFTR as String, getVersionString(DocumentType.CFTR))
+        result.put(DocumentType.TIR as String, getVersionString(DocumentType.TIR))
+        result.put(DocumentType.TIP as String, getVersionString(DocumentType.TIP))
```
jafarre-bi, line 1743: Instead of unrolling the loop, it would be better to use a for-each loop over DocumentType.values()

jafarre-bi: By the way, a for-each loop can also be used in getVersion for Environment.values()

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -1776,4 +1789,75 @@ class LeVADocumentUseCase extends DocGenUseCase {
...
+        Map data = null
+        if (args.length > 0) {
+            repo = args[0]
+            if (args.length > 1) {
+                data = args[1]
+            }
+        }
+
+        switch(methodName) {
+            case "createCFTP":
```
jafarre-bi, line 1804: Without a clear proof that calling the createXXX methods from NonCPS code is causing problems, I don't see this change justified. The warning has been known to be there for a long time, but we never found problems in this concrete case, yet. My recommendation is to remove this change from the PR.

jafarre-bi (changes requested):

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
metmajer: @jafarre-bi do you see anything that we need to take from this PR? If not, we might as well close.

metmajer:
@jafarre-bi @hrcornejo if `collectEntries` needs to be replaced, the following example could help. Apparently, the approach didn't make it into the PR last time https://github.com/opendevstack/ods-jenkins-shared-library/pull/750#issuecomment-961984666.

PR #948, opened 2022-10-11
Fix when the git checkout raise an Exception, the RM is not informed …

angelmp01:
…because the Registry is not initialized. ([#938]https://github.com/opendevstack/ods-jenkins-shared-library/issues/938)


braisvq1996 (approved):
Cherry-pick in 4.x branch of #939

Issue #955, opened 2022-11-07
Risk Priority Number is rendered as N/A when is 0


albertpuente:
**Describe the bug**
The risk priority number in the documentation is rendered as "N/A" when its source value is 0, which is a valid value (and a number).

**To Reproduce**
1. Define a RA with RPN = 0.
2. Generate RA document.

**Expected behavior**
RPN should appear as 0.

**Affected version (please complete the following information):**
 - OpenDevStack 4.x

metmajer:
Closed via #956

PR #956, opened 2022-11-07
RPN to use number even if 0 and RP to preserve value unless null

albertpuente:
Solves #955


src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -590,8 +590,8 @@ class LeVADocumentUseCase extends DocGenUseCase {
                     probabilityOfOccurrence: probabilityOfOccurrence ? probabilityOfOccurrence."short" : "None",
                     severityOfImpact: severityOfImpact ? severityOfImpact."short" : "None",
                     probabilityOfDetection: probabilityOfDetection ? probabilityOfDetection."short" : "None",
-                    riskPriority: riskPriority ? riskPriority.value : "N/A",
-                    riskPriorityNumber: r.riskPriorityNumber ?: "N/A",
+                    riskPriority: (riskPriority != null) ? riskPriority.value : "N/A",
```
jafarre-bi, line 593: If you want to get a text value for the riskPriority, you need to use either .text (long) or .short (short). It is similar to the probabilityOfDetection, for example. Please, note that one of the enum values for the riskPriority, concretely riskPriority.value == 0, has short "N" and text "N/A". This may create confusion with the "N/A" that would be shown if riskPriority is null.
Not very important, but the (riskPriority != null) ? ... can perfectly be left as before riskPriority ? ...
For consistence with the other enums that are shown here, I suggest:
riskPriority = riskPriority."short" : "N/A"

jafarre-bi (changes requested):

jafarre-bi (approved):

metmajer:
@albertpuente @jafarre-bi the same PR needs to go into master so we don't lose it for future releases.

PR #961, opened 2022-11-10
Merge master branch over 4.x

hrcornejo:


jafarre-bi:
This will have to be force-pushed into 4.x, so that the histories of master and 4.x become perfectly aligned. So, closing the PR.

PR #968, opened 2023-01-20
Deploy into Q fails

angelmp01:
Updated InitStage.groovy to show a correct exception when fails the checkout of the release repository


braisvq1996 (approved):
:rocket:

PR #982, opened 2023-03-13
Bug/981 component fails to deploy more than one deploymentconfig

### LINKED ISSUE #981: RM (to DEV) fails with one component containing multiple deployment(s)means
**Describe the bug**
Due to the changes introduced in a previous release, the mono repository configuration cannot deploy more than one DeploymentConfig.

https://github.com/opendevstack/ods-jenkins-shared-library/blob/63d3fa0435721cfc0bb9f21f24d15227784f27d7/src/org/ods/orchestration/phases/FinalizeOdsComponent.groovy#L145

**To Reproduce**
1. Try to release a repository with more than one deploymentConfig.
2. See error: `java.lang.RuntimeException: Found more than 1 Deploymentmeans 2 for repo database. This is not supported``

**Expected behavior**

 - OpenDevStack 4.x

**Additional context**
Having multiple dcs in one component used to work in the datascience qs.

serverhorror:
Fixes #981

### DIFF
--- a/src/org/ods/orchestration/phases/FinalizeOdsComponent.groovy
+++ b/src/org/ods/orchestration/phases/FinalizeOdsComponent.groovy
@@ -137,31 +137,28 @@ class FinalizeOdsComponent {
             throw new RuntimeException ("Deploymentmeans for repo ${repo.id}" +
                 " are not found. These means are automatically generated!\n" +
                 " Likely you upgraded and tried to deploy to Q instead of rebuilding on dev")
-        } else if (deploymentMeans.size() > 1) {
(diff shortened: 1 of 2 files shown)


src/org/ods/orchestration/phases/FinalizeOdsComponent.groovy
```diff
@@ -137,41 +137,39 @@ class FinalizeOdsComponent {
             throw new RuntimeException ("Deploymentmeans for repo ${repo.id}" +
```
clemensutschig, line 137: see internal thread - maybe we should remove this check as well?! .. can you try to create a component W/O deployment . and transport it thru?!

PR #1046, opened 2023-10-22
Bug/989 bitbucketservice uses deprecated api call

### LINKED ISSUE #989: Bitbucket API call deprecation notice
**Describe the bug**
An API call we use in the shared library have been market as deprecated in Bitbucket version 7.14, released on June 2021. Currently we are using version 7.21.

This api call is used in [BitbucketService](https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/services/BitbucketService.groovy#L286-L314)

We should update the api call to the new one porvided by Atlassian
 https://docs.atlassian.com/bitbucket-server/rest/7.21.0/bitbucket-rest.html#idp230

**Screenshots**
https://developer.atlassian.com/server/bitbucket/rest/v808/api-group-deprecated/#api-build-status-latest-commits-commitid-get

image

**Affected version:**
 - OpenShift: [4.11]
 - OpenDevStack [4.2]
 - Bitbucket [7.21]

@metmajer

serverhorror:
fix #989


braisvq1996:
> There's still one call left. I can't really find a good way to get rid of it but we're mostly there now. @braisvq1996 -- opinions? It gets us at least along the way...

Hi Martin, we are preparing a Quickstarter release at the moment. Once this is done I should have time to check this PR

PR #1078, opened 2024-03-11
Add Openshift appDomain in Quickstarter Pipeline context

braisvq1996:
Related to https://github.com/opendevstack/ods-quickstarters/issues/997

In this PR we add to the context a new value, appDomain in order to be to configure Nexus URL dynamically in Quickstarter pipeline

Simplified the process to get openshift domain, intead of creating a dummy route and deleting it afterwards, we take the domain from cluster console url

### DIFF
--- a/src/org/ods/quickstarter/Context.groovy
+++ b/src/org/ods/quickstarter/Context.groovy
@@ -1,13 +1,29 @@
 package org.ods.quickstarter
 
-import com.cloudbees.groovy.cps.NonCPS
+import org.ods.util.Logger
+import org.ods.util.IPipelineSteps
+import org.ods.util.PipelineSteps
+
+import org.ods.services.OpenShiftService
+import org.ods.services.ServiceRegistry
 
 class Context implements IContext {
 
-    private final Map config
+    private final Logger logger
+    private IPipelineSteps steps
+    private final def script
+    private OpenShiftService openShiftService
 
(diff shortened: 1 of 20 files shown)


src/org/ods/quickstarter/Context.groovy
```diff
@@ -145,4 +157,22 @@ class Context implements IContext {
...
+    @NonCPS
+    String getAppDomain() {
+        config.appDomain
+    }
+
+    void setAppDomain() {
+        if (config.projectId) {
+            logger.startClocked("${config.componentId}-get-oc-app-domain")
+
+            this.openShiftService = new OpenShiftService(steps, logger)
```
jafarre-bi, line 169: Context is just a regular bean, providing getters to access the information inside the config map.
It may be acceptable to add a setter, but not to have the setter perform complex business logic to retrieve the app domain.
The setter should get a String as parameter and just assign it to the config. The logic to provide the value should be somewhere else. Please, contact me to discuss this in detail.

jafarre-bi (changes requested):

src/org/ods/quickstarter/Context.groovy
braisvq1996: cleaned

jafarre-bi (dismissed):

jafarre-bi (approved):

PR #1083, opened 2024-03-25
Reduce Jira calls

### LINKED ISSUE #654: rm: performance / excessive external calls in LevaDocumentHistory:getReferencedDocumentsVersion
**Describe the bug**
with jenkins transforming all calls into CPS - we need to be extra carefull on large open (functional) loops - that do extensive external work .. leading to NonSerializableExceptions ..

The method in 
https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy#L1729-L1750

is a great (=bad) example ... we go out for each docType, in each document generation, to find the histories against jira (in the most excessive case) - also w/o using any caching

**To Reproduce**
follow up from opendevstack/ods-jenkins-shared-library/#645

**Expected behavior**
performant code :D

 - OpenShift: [3.11/4]
 - OpenDevStack [3.x/master]

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
see below
```

**Additional context**
...

hrcornejo:
Closes #654 and also reduces the number of Jira calls to update the labels of a Test issue. Based in old draft PR #829

### DIFF
--- a/src/org/ods/orchestration/util/ConcurrentCache.groovy
+++ b/src/org/ods/orchestration/util/ConcurrentCache.groovy
@@ -11,7 +11,8 @@ class ConcurrentCache<K, V> {
     private final ConcurrentMap<K, V> cache
     private final Function<? super K, ? extends V> computeIfAbsent
 
-    ConcurrentCache(Function<? super K, ? extends V> computeIfAbsent, int initialCapacity = 16, float loadFactor = 0.75f, int concurrencyLevel = 1) {
+    ConcurrentCache(Function<? super K, ? extends V> computeIfAbsent,
+                    int initialCapacity = 16, float loadFactor = 0.75f, int concurrencyLevel = 1) {
         this.cache = new ConcurrentHashMap(initialCapacity, loadFactor, concurrencyLevel)
         this.computeIfAbsent = computeIfAbsent
     }
@@ -26,4 +27,5 @@ class ConcurrentCache<K, V> {
     String toString() {
         return cache.toString()
     }
+
 }
--- a/src/org/ods/orchestration/util/ConcurrentCache.groovy
+++ b/src/org/ods/orchestration/util/ConcurrentCache.groovy
@@ -2,28 +2,24 @@ package org.ods.orchestration.util
 
 import com.cloudbees.groovy.cps.NonCPS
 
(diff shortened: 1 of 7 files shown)


jorge-romero (approved):

src/org/ods/orchestration/util/ConcurrentCache.groovy
```diff
@@ -0,0 +1,33 @@
...
+import com.cloudbees.groovy.cps.NonCPS
+
+import java.util.concurrent.ConcurrentMap
+import java.util.function.Function
+
+class ConcurrentCache<K, V> {
+
+    private final ConcurrentMap<K, V> cache
+
+    ConcurrentCache(def cache) {
```
jafarre-bi, line 12: Let's use a ConcurrentHashMap and don't let the user provide it to avoid possible concurrency issues. We could pass the constructor parameters for the map: initial size, load factor and/or concurrency level for writes.

src/org/ods/orchestration/util/ConcurrentCache.groovy
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @NonCPS
+    def get(K key,
+            Function<? super K, ? extends V> computeIfAbsent) {
+        cache.computeIfAbsent(key, computeIfAbsent)
+    }
+
+    @NonCPS
+    def put(K key, V value) {
```
jafarre-bi, line 23: For the same reason, to avoid concurrency problems, let's not provider a put method. Elements will be added using the computeIfAbsent only.

jafarre-bi (changes requested):

src/org/ods/orchestration/util/ConcurrentCache.groovy
hrcornejo: Done

src/org/ods/orchestration/util/ConcurrentCache.groovy
hrcornejo: Done

jorge-romero (approved):

jafarre-bi (approved):

PR #1094, opened 2024-04-19
feature/EDPC-2185 risk assessment table

rpanduga95:
Included  the complete Requirement description to Risk Assessment Table.


jafarre-bi (approved):

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -626,6 +625,23 @@ class LeVADocumentUseCase extends DocGenUseCase {
...
     }
 
+    /**
+     * If the risk is associated to technical specification task which is itself associated with a User story,
+     * return the technical specification task
+     * otherwise, return the system requirements
+     * @param risk
+     * @return
+     */
+    private List<Project.JiraDataItem> getRequirements(Project.JiraDataItem risk) {
```
jafarre-bi, line 635: There will never be more than one requirement, so you don't need to return a List. Just return JiraDataItem.

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -626,6 +625,23 @@ class LeVADocumentUseCase extends DocGenUseCase {
...
+        def requirements
+        if (risk.getResolvedTechnicalSpecifications().size() > 0) {
```
jafarre-bi, line 637: getResolvedTechnicalSpecifications() is a slow method. You don't want to call it twice.
I suggest you initialize the variable requirements with getResolvedTechnicalSpecification and then use the variable.

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -585,8 +585,7 @@ class LeVADocumentUseCase extends DocGenUseCase {
                     name: r.name,
                     description: convertImages(r.description),
                     proposedMeasures: "Mitigations: ${ mitigationsText }<br/>Tests: ${ testsText }",
-                    requirements: requirements.findAll { it != null }.collect { it.name }.join("<br/>"),
-                    requirementsKey: requirements.findAll { it != null }.collect { it.key }.join("<br/>"),
+                    requirements: requirements.findAll { it != null },
```
jafarre-bi, line 588: As there will only be one requirement, It would be better to call the entry and the variable requirement instead of requirements. This will affect the template. The method can also be called getRequirement.

jafarre-bi (changes requested):

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
rpanduga95: We can´t send `JiraDataItem`, Because, each risk can have multiple Tech Specifications or System Requirements. So, It should be List<>

jafarre-bi (approved):

PR #1108, opened 2024-05-08
Enable StatefulSet

### LINKED ISSUE #1107: Support StatefulSet in Helm charts

Currently only `Deployment` and `DeploymentConfig` are supported resource types in release manager. That makes it hard to deploy (some) well-known, commonly used charts (e.g. bitnami charts)

Support StatefulSet in release manager

none

**Additional context**

no additional context

serverhorror:
fix #1107 

output resources fetched from getRolloutData for debugging

/CC @clemensutschig, @dvvanessastoiber

### DIFF
--- a/src/org/ods/component/HelmDeploymentStrategy.groovy
+++ b/src/org/ods/component/HelmDeploymentStrategy.groovy
@@ -116,7 +116,7 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
         // // cases since they don't have triggers.
         // metadataSvc.updateMetadata(false, deploymentResources)
         def rolloutData = getRolloutData(deploymentResources) ?: [:]
-        logger.info(groovy.json.JsonOutput.prettyPrint(groovy.json.JsonOutput.toJson(rolloutData)))
+        logger.info(JsonOutput.prettyPrint(JsonOutput.toJson(rolloutData)))
         return rolloutData
     }
 
--- a/.github/workflows/changelog-enforcer.yml
+++ b/.github/workflows/changelog-enforcer.yml
@@ -8,7 +8,7 @@ jobs:
   changelog:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4.1.4
+      - uses: actions/checkout@v4.1.6
       - uses: dangoslen/changelog-enforcer@v3
         with:
           changeLogPath: 'CHANGELOG.md'
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -7,7 +7,7 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout
-              uses: actions/checkout@v4.1.4
+              uses: actions/checkout@v4.1.6
 
             - name: check the value of github.workspace and runner.temp
               run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,17 @@
 
 ## Unreleased
 * Support StatefulSet in Helm charts ([#1108](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1108))
+* Support deployment with only selected components ([#1081](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1081))
 * The Document History indicates clearly when an SLC document has not changed ([#1099](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1099))
+* Fix using repo variable before declaration ([#1117](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1117))
 
 ### Fixed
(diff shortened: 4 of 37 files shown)


src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -95,8 +95,10 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
         logger.info "Rolling out ${context.componentId} with HELM, selector: ${options.selector}"
         helmUpgrade(context.targetProject)
 
+        def deploymentKinds = DEPLOYMENT_KINDS << 'StatefulSet'
```
metmajer, line 98: @serverhorror, is this temporary, or do you plan to add 'StatefulSet' into the global DEPLOYMENT_KINDS?

metmajer: I suggest changing the _protected final_ field AbstractDeploymentStrategy.DEPLOYMENT_KINDS to _protected_ only in the concrete subclass and use a constant in class OpenShiftService instead of a plain String for 'StatefulSet'.

metmajer (changes requested):

metmajer (changes requested):
@serverhorror Thanks, Martin! Please change the _protected final_ field AbstractDeploymentStrategy.DEPLOYMENT_KINDS to _protected_ only and amend it in the _HelmDeploymentStrategy_ class. I also suggest you introduce a new field _OpenShiftService.STATEFULSET_KIND_ vs. using an in-line String.

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -58,15 +58,24 @@ class DeployOdsComponent {
                     applyTemplates(openShiftDir, deploymentMean)
 
                     def retries = project.environmentConfig?.openshiftRolloutTimeoutRetries ?: 10
+                    def podData = null
                     for (def i = 0; i < retries; i++) {
-                        def podData = os.checkForPodData(project.targetProject, deploymentMean.selector)
+                        podData = os.checkForPodData(project.targetProject, deploymentMean.selector)
                         if (podData) {
                             return podData
```
jafarre-bi, line 65: I think this should be a break, not a return.

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -58,15 +58,24 @@ class DeployOdsComponent {
...
                         }
                         steps.echo("Could not find 'running' pod(s) with label '${deploymentMean.selector}' - waiting")
                         steps.sleep(12)
                     }
 
+                    if (podData == null) {
```
jafarre-bi, line 71: I think this should be just if (podData), because emtpy also means no running pods were found and, otherwise, line 81 would throw anyway.

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -58,15 +58,24 @@ class DeployOdsComponent {
...
+                        throw new RuntimeException(
+                        "Error: Could not find 'running' pod(s) with label"
+                        + "'${deploymentMean.selector}'"
+                        )
+                    }
+
+                    // TODO: What if podData is empty?
                     // TODO: Once the orchestration pipeline can deal with multiple replicas,
```
jafarre-bi, line 79: TODO's annotate unfinished work and should not appear in final code.
The case podData is empty should be handled above and the TODO can be removed.
With respect to the second TODO, I know it was there before, but please remove the TODO. Mentioning that currently we only support one instance is enough.

src/org/ods/component/HelmDeploymentStrategy.groovy
jafarre-bi: Why not actually creating OpenShiftService.STATEFUL_SET_KIND and adding it to the DEPLOYMENT_KINDS list? Perhaps this list could be renamed to WORKLOAD_KINDS or something similar.

jafarre-bi (changes requested):

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -58,15 +58,24 @@ class DeployOdsComponent {
...
+                        podData = os.checkForPodData(project.targetProject, deploymentMean.selector)
                         if (podData) {
                             return podData
                         }
                         steps.echo("Could not find 'running' pod(s) with label '${deploymentMean.selector}' - waiting")
                         steps.sleep(12)
                     }
 
+                    if (podData == null) {
+                        throw new RuntimeException(
```
jafarre-bi, line 72: Please, create a specific exception for this, so that it can be caught, if necessary.

serverhorror:
This is dead code

PR #1111, opened 2024-05-14
Workaround for bitbucket trouble with git-notes

ElfenB:
Context: Bitbucket is generating ghost commits that have the commit message "Notes added by 'git notes add'" if git-notes is used (e.g. with semantic-release). To prevent Jenkins starting builds that will result in failure because there is nothing to build, the build should be stopped when those commits are coming in.
Reference: https://github.com/semantic-release/semantic-release/discussions/2017#discussioncomment-995308


braisvq1996:
Yes, the only way would be on the webhook proxy in ods-core repository, but I do not think the feature to filter out certain commits is available.

PR #1116, opened 2024-05-22
Simplify successor management since now issue links are no longer inherited

jordivx:
Update project loadJiraDataForCurrentVersion since now issue links are no longer inherited and appear in latest release deltadocgen report.

### DIFF
--- a/src/org/ods/orchestration/util/Project.groovy
+++ b/src/org/ods/orchestration/util/Project.groovy
@@ -1331,6 +1331,19 @@ class Project {
         return mergedData
     }
 
+    /**
+     * The method checks each issue in the 'relatedIssues' list against the related issues in the 'deltaDocgenData'.
+     * If the issueToUpdate key does not appear in the deltaDocgenData as a related issue but the deltaDocGen has some
+     * issues related for that issue type, the issue is added to the list of issues to be removed.
+     *
+     * @param relatedIssues A list of related issues to be examined.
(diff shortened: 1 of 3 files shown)


src/org/ods/orchestration/util/Project.groovy
```diff
@@ -1266,6 +1268,62 @@ class Project {
...
+                        def relatedIssuesToRemove = findRelatedIssuesToRemove(relatedIssues, deltaDocgenData, relatedIssueType, issueType, issueToUpdate)
+                        mergedData[issueType][issueToUpdate.key][relatedIssueType].removeAll { relatedIssuesToRemove.contains(it) }
+                    }
+                }
+            }
+        }
+        return mergedData
+    }
+
+    protected static List findRelatedIssuesToRemove(List<String> relatedIssues, Map deltaDocgenData, String relatedIssueType, String issueType, Map issueToUpdate) {
```
hrcornejo, line 1299: Add Javadoc to this method. Why this is static?

hrcornejo (changes requested):

src/org/ods/orchestration/util/Project.groovy
jordivx: Added Javadoc. The method is static since it doesn't depend on any instance variables or methods of the class it's defined in.

hrcornejo (approved):

PR #1118, opened 2024-05-28
Substitute env username and password variables for nexus by using instead Jenkins variable "cd-user-with-password"

jeronimolina:
- Modified all calls to NexusService by inserting the Jenkins secret "cd-user-with-password" instead of retrieving the NEXUS_USERNAME and NEXUS_PASSWORD env variables. On vars\odsComponentStageScanWithSonar.groovy, vars\odsComponentStageScanWithAqua.groovy, vars\odsComponentStageScanWithTrivy.groovy and src\org\ods\component\Pipeline.groovy
- NexusService class: removed the retrieving of the env variables in the function readConfigFromEnv
- NexusService class: added jenkins credentials in function newFromEnv


src/org/ods/component/Pipeline.groovy
```diff
@@ -151,8 +151,17 @@ class Pipeline implements Serializable {
...
-                                context.nexusUrl, context.nexusUsername, context.nexusPassword))
+                            script.withCredentials([
+                                script.usernamePassword(
+                                    credentialsId: context.credentialsId,
+                                    usernameVariable: 'USERNAME',
+                                    passwordVariable: 'PASSWORD'
+                                )
+                            ]) {
+                                NexusService nexusService = new NexusService(context.nexusUrl,
+                                    script.env.USERNAME as String, script.env.PASSWORD as String)
```
jafarre-bi, line 162: It is not a good practice to store clear-text credentials in long-living attributes. It would be better to use withCredentials inside the NexusService whenever needed. Is there anything that prevents us from doing this?

jafarre-bi (approved):

src/org/ods/component/Pipeline.groovy
jafarre-bi, line 162: It is not a good practice to store clear-text credentials in long-living attributes. It would be better to use withCredentials inside the NexusService whenever needed. Is there anything that prevents us from doing this?

jafarre-bi (changes requested):

jafarre-bi:
@jeronimolina we can have a call to comment on this.

src/org/ods/component/Pipeline.groovy
jeronimolina: This was my first approach when I started coding it, but I found the blocker of CPS methods defined here https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/. Some of the methods in the NexusService class that needed credentials were already labeled as @NonCPS and when I introduced the credentials there the methods needed to be CPS per definition of Jenkins. When I removed the label of the methods, this created a cascade of errors which I could not solve as it impacted methods way above that were already labeled as @NonCPS and needed to be changed too. I lost a lot of time trying to fix it but the task complicatedtoo much and needed more of a general refactor of the shared library. This is the reason why i decided the other approach which was simpler and cleaner.

jeronimolina:
Hello @jafarre-bi I added encryption on the NexusService class, now we are not storing directly the plain password into the class. If you can take a look and give me your opinion!

src/org/ods/services/NexusService.groovy
```diff
@@ -10,24 +10,16 @@ import org.apache.http.client.utils.URIBuilder
...
+    final def script
+    final String credentialsId
 
-    NexusService(String baseURL, String username, String password) {
+    NexusService(String baseURL, def script, String credentialsId) {
         if (!baseURL?.trim()) {
             throw new IllegalArgumentException("Error: unable to connect to Nexus. 'baseURL' is undefined.")
         }
 
-        if (!username?.trim()) {
```
jafarre-bi, line 23: We should validate that script is not null and credentialsId is not blank.

src/org/ods/services/NexusService.groovy
```diff
@@ -88,12 +69,23 @@ class NexusService {
...
+        ]) {
+            def restCall = Unirest.post("${this.baseURL}/service/rest/v1/components?repository={repository}")
+                .routeParam('repository', repository)
+                .basicAuth(script.env.USERNAME, script.env.PASSWORD)
+            return processsStoreArtifact(restCall, repository, artifact, contentType, repositoryType, nexusParams)
+        }
+    }
 
+    @NonCPS
+    private URI processsStoreArtifact(def restCall, String repository, byte[] artifact, String contentType, String repositoryType, Map nexusParams = [ : ]) {
```
jafarre-bi, line 88: There's a typo, one more s than needed. I would also call it processStoreArtifactRes(posnse). Otherwise it's not very clear.

src/org/ods/services/NexusService.groovy
```diff
@@ -133,19 +125,30 @@ class NexusService {
...
+            script.usernamePassword(
+                credentialsId: credentialsId,
+                usernameVariable: 'USERNAME',
+                passwordVariable: 'PASSWORD'
+            )
+        ]) {
+            String urlToDownload = "${this.baseURL}/repository/${nexusRepository}/${nexusDirectory}/${name}"
+            def restCall = Unirest.get("${urlToDownload}")
+                .basicAuth(script.env.USERNAME, script.env.PASSWORD)
+            return (processRetrieveArtifact(restCall, urlToDownload, nexusRepository, nexusDirectory, name, extractionPath))
```
jafarre-bi, line 146: I would call it processRetrieveArtifactRes (or Response).

src/org/ods/services/NexusService.groovy
```diff
@@ -170,20 +173,27 @@ class NexusService {
...
-        response.ifFailure {
-            throw new RuntimeException ("Could not retrieve data from '${urlToDownload}'")
+        script.withCredentials([
+            script.usernamePassword(
+                credentialsId: credentialsId,
+                usernameVariable: 'USERNAME',
+                passwordVariable: 'PASSWORD'
+            )
+        ]) {
+            String urlToDownload =
```
jafarre-bi, line 187: withCredentials should contain only lines 189-191.

src/org/ods/services/NexusService.groovy
```diff
@@ -88,12 +69,23 @@ class NexusService {
...
+            script.usernamePassword(
+                credentialsId: credentialsId,
+                usernameVariable: 'USERNAME',
+                passwordVariable: 'PASSWORD'
+            )
+        ]) {
+            def restCall = Unirest.post("${this.baseURL}/service/rest/v1/components?repository={repository}")
+                .routeParam('repository', repository)
+                .basicAuth(script.env.USERNAME, script.env.PASSWORD)
+            return processsStoreArtifact(restCall, repository, artifact, contentType, repositoryType, nexusParams)
```
jafarre-bi, line 83: This should be outside the withCredentials block

src/org/ods/services/NexusService.groovy
```diff
@@ -133,19 +125,30 @@ class NexusService {
...
-        String urlToDownload = "${this.baseURL}/repository/${nexusRepository}/${nexusDirectory}/${name}"
-        def restCall = Unirest.get("${urlToDownload}")
-            .basicAuth(this.username, this.password)
+        script.withCredentials([
+            script.usernamePassword(
+                credentialsId: credentialsId,
+                usernameVariable: 'USERNAME',
+                passwordVariable: 'PASSWORD'
+            )
+        ]) {
```
jafarre-bi, line 142: The withCredentials block should contain only lines 144-145.

test/groovy/org/ods/PipelineMock.groovy
```diff
@@ -0,0 +1,28 @@
+package org.ods
+
+class PipelineMock {
```
jafarre-bi, line 3: Let's do it in a simpler way. You can double-check that the steps parameter you are passing implements IPipelineSteps. Add a withCredentials Method to IPipelineSteps and its implementations. Then just create a Stub(IPipelineSteps) for the test methods.

jafarre-bi (dismissed):

jafarre-bi (approved):

PR #1121, opened 2024-06-04
Fix Nexus service

braisvq1996:
BugFix for nexus service of the issues added in https://github.com/opendevstack/ods-jenkins-shared-library/pull/1118

### DIFF
--- a/src/org/ods/services/NexusService.groovy
+++ b/src/org/ods/services/NexusService.groovy
@@ -79,8 +79,7 @@ class NexusService {
 
     @SuppressWarnings('LineLength')
     URI storeComplextArtifact(String repository, byte[] artifact, String contentType, String repositoryType, Map nexusParams = [ : ]) {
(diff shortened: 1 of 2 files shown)


src/org/ods/services/NexusService.groovy
```diff
@@ -190,15 +190,15 @@ class NexusService {
...
+        def response
         steps.withCredentials([
             steps.usernamePassword(
                 credentialsId: credentialsId,
                 usernameVariable: 'USERNAME',
                 passwordVariable: 'PASSWORD'
             )
         ]) {
-            response.basicAuth(steps.env.USERNAME, steps.env.PASSWORD).asString()
+            response = Unirest.get("${urlToDownload}").basicAuth(steps.env.USERNAME, steps.env.PASSWORD)
```
hrcornejo, line 201: Without .asString no call is done

hrcornejo (changes requested):

hrcornejo (approved):

PR #1124, opened 2024-06-05
COntext variables nexusUsername and nexusPassword are set again by nexusService

jeronimolina:
This is because some quickstarters need that variables, and they are going to use the developer user with only read only

### DIFF
--- a/src/org/ods/services/NexusService.groovy
+++ b/src/org/ods/services/NexusService.groovy
@@ -56,6 +56,16 @@ class NexusService {
         } else {
             throw new IllegalArgumentException("Environment variable 'NEXUS_URL' is required")
         }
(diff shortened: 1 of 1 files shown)


hrcornejo:
@jeronimolina, @jafarre-bi and @braisvq1996 with this change the Nexus service will use **cd-user-with-password** as credentials to manage document in Nexus, and the quickstarters will use the env set in Jenkins (NEXUS_USERNAME, NEXUS_PASSWORD) as credentials to access repo. is right?

braisvq1996:
@hrcornejo it is correct, additionally env user will get only read permission on nexus on the public repositories for retro compatibility

hrcornejo (approved):

PR #1132, opened 2024-06-24
Fix helm deployment information not provided

henrjk:
This information is missing when aiming to generate an installation report.
(See EDPC-2361)
Note that this code run into an image verification issue, so this is a draft only.


henrjk:
This has been superseded by #1133

PR #1135, opened 2024-07-01
TIR - remove dynamic pod data and surface helm status

henrjk:
This change is meant to be released with the change in https://github.com/opendevstack/ods-document-generation-templates/pull/145 although there is no technical coupling at the moment.

If helm is used for deployment, the components deployed by helm are no longer listed as toplevel elements in the TIR. Instead the helm release is represented by its status so that users will see tables under:

Deployment Resource: <helm-release-name>-deploymentMean
Deployment Resource: <helm-release-name>-deploymentStatus

The DeploymentStatus has rows surfacing the releaseName, releaseRevision, namespace, deployStatus, deployDescription, lastDeployed and resources.

Here an example of the resources:
"Deployment: backend-helm-monorepo-chart-component-a, backend- helm-monorepo-chart-component-b, Service: backend-helm-monorepo-chart"

Non-Helm deployments are surfaced as before.

Further change notes:
- To avoid CPS issues in Jenkins all types were using their own file and the Immutable annotation was avoided.
- Type HelmStatusData captures helm status with high fidelity whereas HelmStatusSimpleData captures shallower information in particular omitting the details of resources except their names and kinds. This approach simplifies the json Helm status json parsing code while keeping the consuming code relatively lean.
In the deployment artifact the simplified helmStatus is kept under key 'helmStatus' in the deploymentMean information. Which is also written to `ods-deployments.json` by DeploymentDescriptor.
- PodData fields capturing dynamic data where removed, except for the podName which has a prefix useful for correlating dynamic data with pods. At the moment this is still needed for both tailor and helm deployments. The code which finds this named pods did require retry logic even for helm installs using --atomic.


src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -58,21 +60,29 @@ class DeployOdsComponent {
                     applyTemplates(openShiftDir, deploymentMean)
 
                     def retries = project.environmentConfig?.openshiftRolloutTimeoutRetries ?: 10
-                    def podData = null
+                    def podDataContext = [
```
jafarre-bi, line 63: If this data is only to generate an error message, please, generate only when you are about to echo the message.

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
```diff
@@ -227,6 +237,15 @@ class DeployOdsComponent {
...
                         deploymentMean.helmAdditionalFlags,
                         true)
+
+                    def helmStatus = os. helmStatus(project.targetProject, deploymentMean.helmReleaseName)
+                    def helmStatusMap = helmStatus.toMap()
+                    deploymentMean.helmStatus = helmStatusMap
+                    logger.info("${this.class.name} -- HELM STATUS")
+                    logger.info(
+                        JsonOutput.prettyPrint(
+                            JsonOutput.toJson(helmStatusMap)))
```
jafarre-bi, line 247: Same as above

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
```diff
@@ -1130,6 +1131,119 @@ class LeVADocumentUseCase extends DocGenUseCase {
         return this.createDocument(documentType, repo, data_, [:], modifier, getDocumentTemplateName(documentType, repo), watermarkText)
     }
 
+    /**
+     * Helm releases become top level elements, tailor deployments are left alone.
+     */
+    @SuppressWarnings('CyclomaticComplexity')
```
jafarre-bi, line 1137: It is not a good practice to suppress quality warnings. Let's talk and see how we could simplify this code.

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -155,6 +157,33 @@ class OpenShiftService {
...
+        try {
+            def helmStdout = steps.sh(
+                script: "helm -n ${project} status ${release} --show-resources  -o json",
+                label: "Gather Helm status for release ${release} in ${project}",
+                returnStdout: true
+            ).toString().trim()
+            def object = new JsonSlurperClassic().parseText(helmStdout)
+            def helmStatusData = HelmStatusData.fromJsonObject(object)
+            helmStatusData
+        } catch (Exception e) {
```
jafarre-bi, line 181: In groovy you can just catch(e), but it's better to catch concrete exceptions, if possible.

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -429,7 +458,12 @@ class OpenShiftService {
         )
     }
 
-    // Returns data about the pods (replicas) of the deployment.
+    boolean isDeploymentKind(String kind) {
```
jafarre-bi, line 461: This feels like a private method

src/org/ods/util/HelmStatusData.groovy
```diff
@@ -0,0 +1,181 @@
...
+    // release name
+    String name
+    String namespace
+
+    HelmStatusInfoData info
+
+    // version is an integer but we map it to a string.
+    String version
+
+    @SuppressWarnings(['IfStatementBraces'])
```
jafarre-bi, line 20: Better to comply with the rules instead of suppressing warnings

src/org/ods/util/HelmStatusData.groovy
```diff
@@ -0,0 +1,181 @@
...
+                    if (resourceData != null) {
+                        resources << resourceData
+                    }
+                }
+                resourcesMap.put(key, resources)
+            }
+            info.resources = resourcesMap
+            status.info = new HelmStatusInfoData(info)
+            new HelmStatusData(status)
+        } catch (Exception e) {
```
jafarre-bi, line 63: Same as above

jafarre-bi (changes requested):

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -94,10 +95,16 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
...
-        )
+        logger.info("${this.class.name} -- HELM STATUS")
+        logger.info(
+            JsonOutput.prettyPrint(
+                JsonOutput.toJson(helmStatusMap)))
+
+        // not sure if we need both HELM STATUS and DEPLOYMENT RESOURCES"
         logger.info("${this.class.name} -- DEPLOYMENT RESOURCES")
         logger.info(
             JsonOutput.prettyPrint(
```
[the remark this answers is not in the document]
henrjk: The logging of DEPLOYMENT RESOURCES comes from master. Should it be changed?

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -124,7 +132,7 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
                 options.helmValues['componentId'] = context.componentId
 
                 // we persist the original ones set from outside - here we just add ours
-                Map mergedHelmValues = [:]
+                Map mergedHelmValues = [:] as Map<String, String>
```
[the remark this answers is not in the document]
henrjk: Thanks for the suggestion - I will make that change and also look out for similar cases.

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
henrjk: May rationale here was that the case where no pods are not found is not typically happening. If we move the message generation in a conditional it could be that the code itself is flawed and potentially cause an exception. 
Given that I already have tested the code as is I am now confident that the code would work and we could still move it.

src/org/ods/orchestration/usecase/LeVADocumentUseCase.groovy
henrjk: The codeNarc report says its cyclomatic complexity of 21 where the threshold is 20. Now in https://codenarc.org/codenarc-rules-size.html the rules are stated. Adding another statement if (logger.debugMode) to not do json encoding when not logging will raise complexity by one.
Let's discuss whether there are good ideas to make this more clear.

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -161,46 +169,66 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
...
-                // TODO: Once the orchestration pipeline can deal with multiple replicas,
-                // update this to store multiple pod artifacts.
-                // TODO: Potential conflict if resourceName is duplicated between
-                // Deployment and DeploymentConfig resource.
-                context.addDeploymentToArtifactURIs(resourceName, podData[0]?.toMap())
+                steps.echo("${msgPodsNotFound} - waiting")
+                steps.sleep(12)
+            }
+            if (!podData) {
+                throw new RuntimeException(msgPodsNotFound)
```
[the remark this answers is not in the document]
henrjk: My expectation is that it should never happen. Would there be a reason to nonetheless catch it somewhere?

src/org/ods/orchestration/phases/DeployOdsComponent.groovy
henrjk: Thanks will change!

src/org/ods/services/OpenShiftService.groovy
henrjk: Surrounding code in OpenShiftService has simply `catch (ex) {`. Not sure this is better, but at least it would be one style.

src/org/ods/services/OpenShiftService.groovy
henrjk: It used when generating the TIR data in assembleDeployments called by org.ods.orchestration.usecase.LeVADocumentUseCase#createTIR
My rationale was that the createTIR method should not need to know what deployment kinds we support. Let's discuss more.

src/org/ods/util/HelmStatusData.groovy
henrjk: To my eyes the code looks cleaner in this particular case.

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -94,10 +95,16 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
 
         logger.info "Rolling out ${context.componentId} with HELM, selector: ${options.selector}"
         helmUpgrade(context.targetProject)
+        HelmStatusSimpleData helmStatus = openShift.helmStatus(context.targetProject, options.helmReleaseName)
+        def helmStatusMap = helmStatus.toMap()
```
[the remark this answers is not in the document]
henrjk: Thank you - will adjust the code!

henrjk:
This PR has been replaced by PR #1143.

PR #1149, opened 2024-08-26
Release 4.5.5

braisvq1996:


jafarre-bi (approved):
The hasGivenTypes method admits several improvements, but should work correctly. We should review it later.

PR #1170, opened 2024-10-24
Pass inherited branches config to Aqua Stage

mobilutz:
While debugging an issue where the Aqua stage was ALWAYS run even when the build-stage was not, I stumbled over the issues that the inherited branches config is not being passed to the new Aqua stage that is created.

I tested it on a running jenkins setup, and can say that with this change, I see the following log output:
```
Skipping stage 'Build OpenShift Image (app-backend-base)' for branch 'prettier' as it is not covered by: 'main', 'review'.

Skipping stage 'Aqua Security Scan' for branch 'prettier' as it is not covered by: 'main', 'review'.
```

I know that a test for this case is missing, but I was not able to install JDK 11 locally, and with any newer JDK version I could not get groovy to work.

Any support on the testing front is greatly appreciated.

PS: Without this change, the truncated logs looks like this:
```
Skipping stage 'Build OpenShift Image (app-backend-base)' for branch 'prettier' as it is not covered by: 'main', 'review'.

oc -n ods get ConfigMap/aqua -o json

...

Skipping as imageRef could not be retrieved. Possible reasons are:
-> The aqua stage runs before the image build stage and hence no new image was created yet.
-> The image build stage was not executed because the image was imported.
-> The aqua stage and the image build stage have different values for 'resourceName' set.
```

The problem is, that the OpenShift image is not being build because of the branches setup, but then the AquaScan is started here: 
https://github.com/opendevstack/ods-jenkins-shared-library/blob/f5f9d14ca9887c721d9afe17c09b7f5c19a5bde8/vars/odsComponentStageBuildOpenShiftImage.groovy#L27

Another idea would be not to run the Aqua stage at all when the build stage is not executed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Fix Tailor deployment drifts for D, Q envs ([#1055](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1055))
 * Fix branch mapping logic to ensure correct environment assignment ([#1263](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1263))
 * Fix Nexus not accepting npm artifacts when uploaded through odsComponentStageUploadToNexus ([#1268](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1268))
+* Aqua Stage is being skipped when branches are not eligable ([#1170](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1170))
 
 ## [4.12.1] - 2026-04-10
 ### Fixed
--- a/test/groovy/vars/OdsComponentStageScanWithAquaSpec.groovy
+++ b/test/groovy/vars/OdsComponentStageScanWithAquaSpec.groovy
@@ -310,4 +310,57 @@ class OdsComponentStageScanWithAquaSpec extends PipelineSpockTestBase {
         assertJobStatusSuccess()
     }
 
+    def "forwards branches config into ScanWithAquaStage"() {
+        given:
+        def c = config + [environment: 'dev']
+        IContext context = new Context(null, c, logger)
+        def aquaConfigMapName = ScanWithAquaStage.AQUA_CONFIG_MAP_NAME
+
+        OpenShiftService openShiftService = Stub(OpenShiftService.class)
+        openShiftService.getConfigMapData('ods', aquaConfigMapName) >> [
+            enabled: true, alertEmails: "mail1@mail.com", url: "http://aqua", registry: "internal"
+        ]
+        openShiftService.getConfigMapData("foo", aquaConfigMapName) >> [enabled: true]
+        ServiceRegistry.instance.add(OpenShiftService, openShiftService)
+
+        NexusService nexusService = Stub(NexusService.class)
+        ServiceRegistry.instance.add(NexusService, nexusService)
+
+        and: "mock the stage constructor and validate inherited config"
+        GroovyMock(ScanWithAquaStage, global: true)
+        ScanWithAquaStage.getAQUA_CONFIG_MAP_NAME() >> aquaConfigMapName
+
+        when:
+        def script = loadScript('vars/odsComponentStageScanWithAqua.groovy')
+        script.call(
+            context,
+            [
+                imageLabels: [JENKINS_MASTER_OPENSHIFT_BUILD_NAMESPACE: 'ods'],
+                resourceName: 'my-image',
+                branches: ['master', 'release/']
+            ]
+        )
+
+        then:
+        1 * new ScanWithAquaStage(
+            _,
+            _,
+            {
+                it.resourceName == 'my-image' &&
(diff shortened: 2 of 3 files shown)


gerardcl:
Hi @mobilutz , you could try using [sdkman](https://sdkman.io/) for the local java env management.

tbugfinder:
Regarding the develpoment setup I have the same challenges. It's tedious to figure out what's expected and needed. An option might to adding devcontainer setup.

Starting point: https://github.com/opendevstack/ods-jenkins-shared-library/pull/1173/files

gerardcl:
> > Hi @mobilutz , you could try using [sdkman](https://sdkman.io/) for the local java env management.
> 
> I am not using java in any other project, and I currently only use [`mise`](https://mise.jdx.dev/) as my overall version manager.
> 
> When I try to install JDK 11 with mise, I get the following:
> 
> ```shell
> ❯ mise use java@11
> mise ERROR no metadata found for version 11
> mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
> ```
> 
> It would be good, to have a possibility like devcontainer os something else that does not rely on a JDK install locally.
> 
> @gerardcl I know that everyone is very busy, but would it be possible to have you are someone else support with writing a test setup for this. I tried to copy things from here, but without running it locally I don't think I can finish the test setup.
> 
> https://github.com/opendevstack/ods-jenkins-shared-library/blob/f5f9d14ca9887c721d9afe17c09b7f5c19a5bde8/test/groovy/vars/OdsComponentStageBuildOpenShiftImageSpec.groovy#L218-L251

hi @mobilutz feel free to connect internally, thanks

metmajer:
@mobilutz I suggest to follow the devcontainer approach which is already used by some other Quickstarters (inf-terraform) and ODS itself. @BraisVQ can you share the example we have been exchanging with a user in the business?

metmajer:
By the way, this was the approach @BraisVQ and I wanted to test with regards to devcontainer using Jenkins agent images:

```
{
    "image": "default-route-openshift-image-registry.apps.eu-dev.ocp.aws.boehringer.com/ods/jenkins-agent-python:4.x",
    "postStartCommand": "sleep infinity"
}
```
 
To be able to access this image from your local machine, you will need to login to our OpenShift container image registry in EU first:

```
docker login -u openshift -p $(oc whoami -t) default-route-openshift-image-registry.apps.eu-dev.ocp.aws.boehringer.com
```

It would require some testing though.

mobilutz:
> @mobilutz could you describe again what your problem with Aqua is?

@metmajer The problem I have is, that the Aqua stage is running even when the build stage was skipped. That means that the image that Aqua should scan is not there and of course it is failing.

Regarding `devcontainers`: I will try to use it, the only problem I have is that I do not have access to OpenShift on my local machine as I am an external-employee and do not use Citrix for my Development Environment :|
I will try it out nevertheless.

metmajer:
@mobilutz not sure I understand the scenario. Instead of building, what do you do instead? Can you elaborate on the concept of "inherited branches"?

mobilutz:
> @mobilutz not sure I understand the scenario. Instead of building, what do you do instead? Can you elaborate on the concept of "inherited branches"?

@metmajer I think it is easier with sharing our (redacted) Jenkinsfile that help to show the situatuion:
```
...
odsComponentPipeline(
  ...
  branchToEnvironmentMapping: [
    'master': 'review',
  ],
  ...
) { context ->
  ...
  container('node') {
    ...
  }
  ...
  container('node') {
    stage('Quality Check') {
      parallel (
        ...
        "Tests": {
          sh(
            label: 'Test',
            script: 'npm run test',
          )
        },
        ...
        failFast: true
      )
    }
  }
 ...
}
...
```

As you can see, we are only map `master` to an environment `review`. Only the `master` branch is therefore build with the `BuildOpenShiftImageStage`. That Build-Stage only runs on the `master` branch, and on all other branches it skips because it is not needed.
But in that Build-Stage, the Aqua-Stage is started, and that then ALWAYS run, even when it is not needed.

To prevent the Aqua-Stage to run in this situation, it is needed to pass the branches to the `inheritedConfig`.

I hope that brings more light in the situation that I guess more than we have.

mobilutz:
This change is still missing specs, I do not see myself writing specs in the near future ☹️ 

But I can say that we are running this change with the help of custom forked-repositories of the `ods-jenkins-shared-library` for a couple of weeks now without any issues.

Maybe someone can support here with writing specs, this change will help developers a lot as the CI pipeline run can be shortened by quite a bit depending on the setup the repository has.

mobilutz:
Thanks @hrcornejo, I almost forgot about this change.

I rebased the changes. You can take a look at it again.

hrcornejo (approved):

hrcornejo (changes requested):
Please @mobilutz fix the failing test:
image

hrcornejo (approved):

PR #1180, opened 2024-11-27
add helmStringValues support

### LINKED ISSUE #1179: Helm chart validation error if a commit hash starts with 8 or more numbers
**Describe the bug**
[ Credit goes to @renedupont ] A Helm chart can contain a values.schema.json where data types are validated. The `imageTag` is supposed to be a string, because it is basically taken from the short commit hash and usually has a letter in it. Sometimes the leading 8 characters of the commit hash are all numbers. With the current implementation, such a special commit hash will cause the validation to fail, because helm interprets it as a number, rather than a string.

**To Reproduce**
1. Work on a component branch with auto deployment enabled
2. Generate a commit with the leading 8 characters of the commit hash all numbers
3. Push the code
4. See error in Jenkins

**Expected behavior**
Commit hashes with leading numbers should not abort Jenkins builds.

**Screenshots**
N/A

 - OpenShift: 4.x
 - OpenDevStack 4.x

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
random-project:
- global.imageTag: Invalid type. Expected: string, given: integer
```

**Additional context**
N/A

faust2199:
Fixes #1179

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixed
 * Fix Tailor deployment drifts for D, Q envs ([#1055](https://github.com/opendevstack/ods-jenkins-shared-library/pull/1055))
+* Helm chart validation error if a commit hash starts with 8 or more numbers ([#1179](https://github.com/opendevstack/ods-jenkins-shared-library/issues/1179))
 
 ## [4.6.0] - 2024-10-23
 
--- a/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
+++ b/docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
@@ -144,11 +144,21 @@ _String_
  referenced by `chartDir` exists.
 
 
-| *helmValues* +
+| *helmStringValues* +
 _Map<String,&nbsp;String>_
 |Key/value pairs to pass as values (by default, the key `imageTag` is set
  to the config option `imageTag`). Only relevant if the directory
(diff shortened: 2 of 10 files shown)


gerardcl (approved):
LGTM! 🚀

jafarre-bi (commented):
Do we need both values and string values? We are using strings in both cases. Not sure about its usefulness.
Apart from that, this is a new parameter that we need to reflect in the TIR. The changes to the TIR are about to be merged. When that is ready, merge those changes into this branch and add the support for this new parameter.

jafarre-bi (changes requested):
Please wait for the changes to the TIR which are about to be merged and then update the TIR with this new parameter.

faust2199:
> Do we need both values and string values? We are using strings in both cases. Not sure about its usefulness. Apart from that, this is a new parameter that we need to reflect in the TIR. The changes to the TIR are about to be merged. When that is ready, merge those changes into this branch and add the support for this new parameter.

It is possible that someone has been using existing automatic type conversion. A hypothetical example:
```
  odsComponentStageRolloutOpenShiftDeployment(context, [
    'selector': "app.kubernetes.io/name=${context.componentId}",
    'helmEnvBasedValuesFiles': ["values.env.yaml"],
    'helmValues': [
      'replicaCount': '3',
    ],
  ])
```
The above works, because it becomes `--set replicaCount=3`, then 3 becomes integer. If we make it always `--set-string`, then it will not work. Of course, it is also possible that no one has been using it this way and we don't have to maintain backwards compatibility.

PR #1189, opened 2025-01-22
Bugfix for TRC document references wrong Reference Documents

valituguran:
In the Traceability Document, in section 2 Reference Documents, we need to remove the following references:

- Combined Functional and Requirements Testing Plan

And add the following references instead:

- System and Software Design Specification incl. Source Code Review Plan
- Test Cases Plan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


valituguran:
Here is how the document looks after the change:
image

hrcornejo (approved):

PR #1211, opened 2025-04-25
Do not retrieve components from Jira when using an alternate Jira project and not generating documentation

jafarre-bi:
When using an alternate Jira project, component retrieval may fail. However, we don't need to retrieve the components if we aren't generating documentation.


jafarre-bi:
Added a change to ignore WIP doc chapters when not generating documentation

hrcornejo (approved):

PR #1262, opened 2026-02-25
#1260 Improves/fixes Trivy Stage

renedupont:
Closes #1260
 
 **Notable Changes:**
- Enables to use trivy's severity threshold with exit-code option while showing correct bitbucket code insight symbol and keeping report as requested in issue #1260
- Some refactoring including moving code that should be in service layer to TrivyService for better separation of concern
- Fixes bug that exceptions that happen while archiving report to nexus does just return null instead of exception
- Adds a couple of tests to the existing ones
- Renames return code int variable `TRIVY_OPERATIONAL_ERROR` into `TRIVY_FAIL` because the scan can now fail due to exceeding the severity threshold as well.


BraisVQ (commented):
I do not see any problem in the code, let me test it tomorow and if I do not find any issue I approve it

BraisVQ (approved):
🙆

PR #1267, opened 2026-03-05
support ECR-only deployments without EKS rollout and with EKS rollout

s2oBCN:
Adds support for
- pushing container images to AWS ECR without requiring a full EKS cluster deployment. This enables scenarios where images need to be synced to ECR but the actual deployment is handled externally (e.g. by an AWS Lambda).
- pushing images  to AWS ECR and execute EKS cluster deployment


jafarre-bi:
@s2oBCN I don't have any previous information about this PR. Is there a shape for it? Let's talk about it offline, please.

src/org/ods/component/HelmDeploymentStrategy.groovy
```diff
@@ -86,13 +49,19 @@ class HelmDeploymentStrategy extends AbstractDeploymentStrategy {
             logger.warn 'Skipping because of empty (target) environment ...'
             return [:]
         }
-
-        // Tag images which have been built in this pipeline from cd project into target project
-        retagImages(context.targetProject, getBuiltImages())
+        // Maybe we need to deploy to another namespace
+        // (ie we want to deploy a monitoring stack into a specific namespace)
+        def targetProject = context.targetProject
+        if (options.helmValues['namespaceOverride']) {
```
jafarre-bi, line 55: While this may be a useful feature in some scenarios, offering this option opens the door to situations that aren't acceptable in the context of the EDP, whether by mistake or intentionally. Moreover, the EDP deploys only EDP components, and they go to the standard namespaces. Is there anything that prevents the current use case from fulfilling this requirement?

docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
```diff
@@ -165,6 +178,12 @@ _List<String>_
  directory referenced by `chartDir` exists.
 
 
+| *helmWithOnlyECR* +
+_Boolean_
+|Whether to use Helm for deployment or only sync with ECR (defaults to `false`)
```
jafarre-bi, line 183: Can you improve the explanation? It can be a bit confusing what the effect of the true/false values is.

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -1087,9 +1087,18 @@ class OpenShiftService {
     @TypeChecked(TypeCheckingMode.SKIP)
     List<PodData> parsePodJson(podJson, String resourceName = null) {
         List<PodData> pods = []
-        if (podJson && podJson.items.collect { it.status?.phase?.toLowerCase() }.every { it == 'running' }) {
+        if (podJson &&
+            podJson.items.collect { it.status?.phase?.toLowerCase() }.every { it == 'running' || it == 'succeeded' }) {
```
jafarre-bi, line 1091: Why gather data for "succeeded" pods?

src/org/ods/services/OpenShiftService.groovy
```diff
@@ -1087,9 +1087,18 @@ class OpenShiftService {
...
             // If we got passed a resourceName we need to collect all the pod data from each pod
             pods = extractPodData(podJson)
+        } else {
+            logger.debug("Not all pods are in 'running' state or succeeded")
+            podJson.items.collect { pod ->
+                [name: pod.metadata?.name, phase: pod.status?.phase?.toLowerCase()]
+                if (pod.status?.phase?.toLowerCase() != 'running') {
```
jafarre-bi, line 1098: This will also show successful pods. Is it intended?

jafarre-bi (changes requested):

src/org/ods/component/HelmDeploymentStrategy.groovy
s2oBCN: Yes, this is a real use case of a project, they have the standard namespce to deploy the code of the app, and then another one to deploy the monitoring stack

src/org/ods/services/OpenShiftService.groovy
s2oBCN: Because some pods were already deployed successfully, they completed their tasks and then disappeared.

src/org/ods/services/OpenShiftService.groovy
s2oBCN: Because some pods were already deployed successfully, they completed their tasks and then disappeared.

docs/modules/jenkins-shared-library/partials/odsComponentStageRolloutOpenShiftDeployment.adoc
s2oBCN: done

src/org/ods/services/OpenShiftService.groovy
jafarre-bi: Ah, they weren't Deployments, then, understood.

src/org/ods/component/HelmDeploymentStrategy.groovy
jafarre-bi: The release flow D -> Q -> P is essential to ensure traceability, with only one namespace per environment. If they need to deploy into different P environments, they may create independent releases, as is being done for other use cases. Allowing free specification of the namespace is a no-go. Let's take it offline to see how this use case could fulfil its requirements without this feature. Thank you.

s2oBCN: This is ODS infra

jafarre-bi: After the conversation and clarification of the exact needs, we agreed to proceed with a solution that has limited flexibility in the choice of namespaces. This solution will be considered incubating and will only be available for specific ods-infra use cases. Extending its support to more use cases will be evaluated in the future if needed.

jafarre-bi (approved):

PR #1268, opened 2026-03-06
Fix Nexus not accepting npm artifacts when uploaded through odsComponentStageUploadToNexus

EamonnODonnell:
Issue caused by missing filename in multipart request.  
Added logic to capture the distribution filename for NPM and pass it in the call to Nexus

### DIFF
--- a/src/org/ods/services/NexusService.groovy
+++ b/src/org/ods/services/NexusService.groovy
@@ -146,8 +146,12 @@ class NexusService {
                 new ByteArrayInputStream(artifact),
                 ContentType.create(contentType),
                 nexusParams['pypi.asset'].substring( nexusParams['pypi.asset'].lastIndexOf("/") + 1 ))
-        }
-        else {
+        } else if (repositoryType == 'npm') {
+            restCall = restCall.field("npm.asset",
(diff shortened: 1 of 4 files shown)


src/org/ods/services/NexusService.groovy
```diff
@@ -146,8 +146,12 @@ class NexusService {
...
                 ContentType.create(contentType),
                 nexusParams['pypi.asset'].substring( nexusParams['pypi.asset'].lastIndexOf("/") + 1 ))
-        }
-        else {
+        } else if (repositoryType == 'npm') {
+            restCall = restCall.field("npm.asset",
+                new ByteArrayInputStream(artifact),
+                ContentType.create(contentType),
+                nexusParams['npm.asset'])
+        } else {
```
hrcornejo, line 154: Please update also https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/test/groovy/org/ods/services/NexusServiceSpec.groovy adding/updating unit tests to check this new type

hrcornejo (changes requested):

hrcornejo (approved):

PR #1273, opened 2026-05-18
Bugfix/tir document history

angelmp01:


src/org/ods/orchestration/util/DocumentHistory.groovy
```diff
@@ -315,7 +315,8 @@ class DocumentHistory {
     @NonCPS
     private String rationaleIfConcurrentVersionsAreFound(DocumentHistoryEntry currentEntry) {
         def oldVersionsSimplified = this.data.collect {
-            [id: it.getEntryId(), version: it.getProjectVersion(), previousVersion: it.getPreviousProjectVersion()]
+            [id: it.getEntryId(), version: it.getProjectVersion(),
+             previousVersion: it.getPreviousProjectVersion(), docVersion: it.getDocVersion()]
```
jafarre-bi: I think we aren't using the docVersion, so it can be removed.

jafarre-bi (changes requested):

jafarre-bi (approved):

PR #1276, opened 2026-06-10
Tune SonarQube scannner

BraisVQ:
Done in this PR:
- Added a configurable timeout (default 10 min), it does not make the pipeline fail
- Added a configurable file limit size to scan (default 2 MiB)


valituguran (dismissed):
Update the scanTimeout field so that the comment matches its actual default value, other than that looks nice, great work!

valituguran (approved):

PR #1290, opened 2026-08-17
Add unit tests of ods-infra and ods-library components to DTR and Jira

matzehecht:


matzehecht:
@jafarre-bi I saw you and @jorge-romero added a feature to report unit tests for ods-infra components to Jira. I would be nice to report those in this case to the DTR as well to keep it consistent. 
Additionally, it would be nice to do the same for ods-library components, as those may be used to develop traditional code used by other components of type `ods` (ODS_CODE) or similar.

Joselee2908 (approved):

